PR #9, opened 2019-07-14
add encryption and tests

snjax:


barryWhiteHat:
I made a PR to remove the json from version control https://github.com/snjax/maci/pull/1

If you accept this you will need to make some code to do the circom compiltation and trusted setup in the tests. Which i think its a good idea because we want to test the actual code there. Not what we have checked in to version control.

weijiekoh:
Shall we close this PR?

barryWhiteHat:
Sure as this feature has been added in another PR.

PR #10, opened 2019-09-14
Added Encryption, flow, repository + code cleanup

kendricktan:
Added encryption from PR https://github.com/barryWhiteHat/maci/pull/9 and tidied it up.

Also restructured the project into an express web-api for preparation of the coordinator backend.

Added `dockerfile`, and fixed library versions in `package.json` for deterministic builds as well as some scripts for quick testing/debugging:

```bash
yarn circuit:compile
yarn circuit:setup
yarn circuit:generatewitness
yarn circuit:test # executes ./app/utils/crypto.js
```

Added `flow` type-checker into `./app/utils/crypto.js` (who doesn't like types?) 😄


app/contracts/Migrations.json
```diff
@@ -0,0 +1,1385 @@
+{
```
barryWhiteHat: Probably good to remove these jsons as they are automatically created.

circuits/circuit.circom
```diff
@@ -0,0 +1,122 @@
...
+    // i.e. check existing hash exists in the tree
+    component senderExistence = LeafExistence(k, 3);
+
+    senderExistence.preimage[0] <== decrypted_data[0]; // publickey_x
+    senderExistence.preimage[1] <== decrypted_data[1]; // publickey_y
+    senderExistence.preimage[2] <== decrypted_data[2]; // action
+    senderExistence.root <== tree_root;
+
+    for (var i=0; i<k; i++) {
+        senderExistence.paths2_root_pos[i] <== sender_proof_pos[i];
```
barryWhiteHat: Do we need to binary constrin `senderExistence.paths2_root_pos`

circuits/verify_eddsamimc.circom
```diff
@@ -1,5 +1,5 @@
-include "../circomlib/circuits/eddsamimc.circom";
-include "../circomlib/circuits/mimc.circom";
+include "../node_modules/circomlib/circuits/eddsamimc.circom";
```
barryWhiteHat: Is this hot handled directly in "../node_modules/circomlib/circuits/eddsamimc.circom";

kendricktan: Not for messages of arbitrary length

app/contracts/Migrations.json
kendricktan: I think we should keep them as gitignoring these JSON files will cause more trouble in the future when after deploying them

circuits/circuit.circom
kendricktan: This is from https://github.com/rollupnc/RollupNC_tutorial I'll run some tests

PR #11, opened 2019-09-28
Merkle Tree and Smart Contract(s)

kendricktan:
Additions:

1. Added an optimized merkletree for the EVM (based off semaphore's https://github.com/kobigurk/semaphore/tree/master/semaphorejs/contracts)
2. Implemented barebones smart contract to mimic the registry functionality


contracts/Whitelist.sol
```diff
@@ -0,0 +1,20 @@
+pragma solidity 0.5.11;
```
barryWhiteHat: Not sure we need to have whitelist

package.json
```diff
@@ -8,29 +8,38 @@
...
-    "circuit:setup": "cd app/circuits && snarkjs setup",
-    "circuit:generateverifier": "mkdir -p contracts && cd app/circuits && snarkjs generateverifier && mv verifier.sol ../../contracts/Verifier.sol",
+    "circuit:setup": "cd app/circuits && snarkjs setup --protocol groth",
+    "circuit:generateverifier": "mkdir -p contracts && cd app/circuits && snarkjs generateverifier --protocol groth && mv verifier.sol ../../contracts/Verifier.sol",
     "circuit:test": "flow-node --experimental-worker app/utils/crypto.js"
   },
   "dependencies": {
     "body-parser": "^1.19.0",
     "circom": "0.0.31",
     "circomlib": "0.0.6",
```
barryWhiteHat: We should update this later when we get a chacne. This version has security bugs.

app/utils/crypto.js
```diff
@@ -124,7 +126,10 @@ const bobHash = mimc7.multiHash(bobPosition)
 
 // Calculate tree root
 // NOTE: Tree root is basically what has "happended"
-const treeRoot = mimc7.multiHash([aliceHash, bobHash])
+const merkletree = createMerkleTree(1, BigInt(0))
+merkletree.insert(aliceHash)
+merkletree.insert(bobHash)
```
barryWhiteHat: possibly refactor this into tests and lib.

contracts/Whitelist.sol
kendricktan: Sorry could you add your reviews to https://github.com/barryWhiteHat/maci/pull/12 I accidentally pushed to the wrong origin/master 😓 

I wanted a whitelist because I wanted the authorize the maci contract as the central authority to interact with the merkletrees - this is to isolate the logic of handling the events and the logic of updating/inserting leaves into the merkletree and will be easier to maintain in the long run

barryWhiteHat: I reviewed both

PR #12, opened 2019-09-28
Added Smart Contracts + JS Merkle Tree with Circom

kendricktan:
Additions:

1. Added an optimized merkletree for the EVM (based off semaphore's https://github.com/kobigurk/semaphore/tree/master/semaphorejs/contracts)
2. Ensured merkle tree implementation works with circom circuits
3. Implemented barebones smart contract to mimic the registry functionality


app/utils/contract.js
```diff
@@ -0,0 +1,63 @@
...
+
+// const registryMTContract = new ethers.Contract(
+//   contractAddresses['registryMTAddress'],
+//   merkleTreeContractDef.abi,
+//   wallet
+// )
+
+const { createMerkleTree } = require('./merkletree')
+
+const main = async () => {
```
barryWhiteHat: possibly move this to a test.

kendricktan: Yes that is a good point, definitely would be good to have some tests

barryWhiteHat: Im okay to merge as is. But just note we should refactor this in the future.

PR #14, opened 2019-10-26
Coordinator Logic, Tests, and Docker/Infra setup

kendricktan:
Whats included:

1. MVP Coordinator logic (with tests and everything)
    - Added redis for caching
    - Added postgres for persistent storage
2. Added coordinator logic testing
    - Insert, Insert, Insert invalid, Update
3. Added merkletree database serialization/deserialization
    - Added merkletree tests
4. Added docker-compose file for convenient local development


contracts/MACI.sol
```diff
@@ -0,0 +1,114 @@
...
+            publisherPublicKey,
+            hashedEncryptedMessaghe
+        );
+    }
+
+    // Updates stateTree should the decryption of
+    // the message is successful
+    function insertMessage(
+        uint256 hashedEncryptedMessage
+    ) public onlyOwner {
```
barryWhiteHat: Ideally this should be open so that anyone can post a message on chain. 

if `owner` is able to censor votes then they can change the result.

contracts/MACI.sol
```diff
@@ -0,0 +1,114 @@
...
+        stateTree.insert(hashedEncryptedMessage);
+        emit MessageInserted(hashedEncryptedMessage);
+    }
+
+    // Inserts a new user
+    function insertUser(
+        uint256 hashedEncryptedMessage
+    ) public onlyOwner {
```
barryWhiteHat: It would be nice to have a smart contract handle this that is not owner.

contracts/MACI.sol
kendricktan: https://github.com/barryWhiteHat/maci/pull/14/files/dd54075c88c6832a34cef1800491fc25ce11b113#diff-22ed51d802f0e4dbf7914ecd35f50992R69

So anyone can publish a message via `pubishMessage`, but only messages that have been successfully decrypted will be inserted

barryWhiteHat: wait i thought that the encryption happens inside the snark? Does this call insertMessage after it has passed ?

kendricktan: The user submits an encrypted message to the coordinator (encrypted with the edch of the user and coordinators keys)

kendricktan: And yes it calls insert message after it has passed

PR #16, opened 2019-11-01
Circom fixes

kendricktan:
Going with the spirit of trying to keep PRs as concise as possible, this PR fixes the logic behind the circom circuit for the state tree update according to the maci architecture spec https://hackmd.io/koAV1hpVSzuB1HZZCo844g


app/circuits-interactions.js
```diff
@@ -0,0 +1,136 @@
...
+const coordinatorPublicKey = privateToPublicKey(coordinatorSecretKey)
+
+// Create 3 users (for registration)
+const user1SecretKey = randomPrivateKey()
+const user1PublicKey = privateToPublicKey(user1SecretKey)
+const user1Message = [...user1PublicKey, 0n]
+const user1EncryptedMsg = signAndEncrypt(
+  user1Message,
+  user1SecretKey,
+  user1SecretKey,
```
barryWhiteHat: Why do you pass user1SecretKey twice?

kendricktan: That's because signing and encrypting are using two different keys. We need to somehow be able to generate the ecdh key, so one public key will always be exposed

PR #22, opened 2019-12-01
Improved private key generation and made tests work in Linux

weijiekoh:
This PR changes the code which generates the EdDSA private key to use a cryptographically secure 32-byte value.

The previous code is not ideal because it 1) uses `Math.random`, which is not secure, and 2) generates a 64-digit base10 number and modulos that by the BabyJub subOrder value, which is a 75-digit base10 number.

It also makes `contracts.js` import `SignupToken.json` not `signupToken.json`, as Linux systems are case sensitive.


weijiekoh:
Hi @kendricktan, one small issue with the merge. This line should read:

https://github.com/barryWhiteHat/maci/blob/master/app/utils/crypto.js#L17

`return BigInt('0x' + crypto.randomBytes(32).toString('hex'))`

and not modulo the babyjub suborder.

weijiekoh:
Hey @kendricktan, @barryWhiteHat just pointed out that the private key needs to be modulo the field size, not the subgroup, as it has to go into the snark. Let's all discuss :)

barryWhiteHat:
`snarkScalarField = 21888242871839275222246405745257275088548364400416034343698204186575808495617n`

PR #25, opened 2019-12-19
Use MiMC instead of blake512 for key derivation

kendricktan:
> barry: Seems like blake2s is overkill for key derivation
> kobi: Cool, lets move to mimc then

This PR changes the key derivation from __blake512__ to __mimc__, rational is to re-user the same hash function (mimc) in the circom circuits and smart contracts for ecdh key derivation. (This is used to prove that the coordinator did decrypt the data with the correct secret key).

As usual, some tests are added, you can run the tests via:
```bash
# In terminal 1
yarn ganache
# In terminal 2
yarn test
```


kobigurk:
Mimc hash should use the zero key :)

kendricktan:
@kobigurk Shouldn't we follow the circom standard? https://github.com/iden3/circomlib/blob/58f758d5ada9672a5ca5282c88f051e25a4f0b09/src/mimc7.js#L7

kobigurk:
@kendricktan Yeah, though the argument that’s passed now as 91 is the key, not NROUNDS, which is a constant

PR #26, opened 2019-12-23
Migrated all hash functions to Mimcsponge 

kendricktan:
Migrated hash functions to use mimc sponge instead of mimc to fit the specification.

This includes the key derivation. 

Also included some more tests


test/testCircuit.js
```diff
@@ -0,0 +1,102 @@
...
+        'message': stringifyBigInts(encryptedMsg),
+        'private_key': stringifyBigInts(sharedKey)
+      }
+
+      const witness = circuit.calculateWitness(circuitInputs)
+
+      // Get the outputs
+      for (let i = 0; i < msg.length; i++) {
+        const idx = circuit.outputIdx(i)
+        assert.equal(msg[i].toString(), witness[idx].toString())
```
weijiekoh: I suggest that we also check if the circuit generates ciphertexts which match the values in `encryptedMsg`.

test/testCircuit.js
```diff
@@ -0,0 +1,102 @@
...
+      const circuitInputs = {
+        'from_x': stringifyBigInts(pk1[0]),
+        'from_y': stringifyBigInts(pk1[1]),
+        'R8x': stringifyBigInts(signature.R8[0]),
+        'R8y': stringifyBigInts(signature.R8[1]),
+        'S': stringifyBigInts(signature.S),
+        'preimage': stringifyBigInts(msg)
+      }
+
+      // If circuit can calculate witness
```
weijiekoh: I suggest adding failing tests which detect exceptions when given invalid signatures.

test/testCircuit.js
```diff
@@ -0,0 +1,102 @@
+const path = require('path')
+const { assert } = require('chai')
+const compiler = require('circom')
+const { stringifyBigInts, unstringifyBigInts } = require('../_build/utils/helpers')
+const { Circuit } = require('snarkjs')
+const { hashLeftRight } = require('../_build/utils/crypto')
```
weijiekoh: I'm not familiar with Flow but I'm not sure that we should import from`_build` rather than from the original source file.

kendricktan: Unfortuntately we mocha doesn't read flow syntax and will throw an error if you import it from 'app' :|

test/testCircuit.js
kendricktan: The circuit isn't responsible for encrypting the values, its used to decrypt the encrypted values

PR #28, opened 2019-12-30
Quadratic vote tally circuit

weijiekoh:
The main contributions in this PR are:

CheckRoot: a circuit that takes a list of leaves and outputs its Merkle root.

QuadVoteTally: the quadratic vote tallying circuit. Operates on batches of leaves instead of the full state tree. It should match the spec. Outputs a hash of the salt + the cumulative vote tally.

CalculateTotal: a circuit which takes a list of numbers and returns their sum.

It also includes unit tests for the above components.


weijiekoh:
Please do't merge yet - I'd like to update the test file to work on more than one batch of state leaves.

weijiekoh:
Ready for review!

PR #30, opened 2020-01-01
Update state tree

kendricktan:
Added circom circuit, solidity files and related tests for `update_state_tree` according to the specifications: https://hackmd.io/koAV1hpVSzuB1HZZCo844g#Check-1-That-the-message-has-been-encrypted-with-the-correct-key

To run the tests:
```
yarn circuit:compile
yarn circuit:setup
yarn circuit:generateverifier

yarn test
```


circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
+
+  signal valid_update;
+  valid_update = valid_signature + sufficient_vote_credits + correct_nonce + valid_state_leaf_index + valid_vote_options_leaf_index;
+
+
+  // If any of the above checks return false
+  // We update the random leaf instead
+
+  // Note: valid_update 0 is true, 1 is false
+  signal selected_state_tree_hash;
```
weijiekoh:
```suggestion
  signal selected_state_leaf_hash;
```

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
+  selected_state_tree_hash <-- valid_update == 0 ? new_state_tree_leaf.hash : random_leaf;
+
+  signal selected_state_tree_path_elements[depth];
+  for (var i = 0; i < depth; i++) {
+    selected_state_tree_path_elements[i] <-- valid_update == 0 ? state_tree_path_elements[i] : random_leaf_path_elements[i];
```
weijiekoh:
```suggestion
    selected_state_tree_path_elements[i] <== valid_update == 0 ? state_tree_path_elements[i] : random_leaf_path_elements[i];
```

Is it possible to use `<==` just in case?

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
+  selected_state_tree_hash === no_op_valid_update;
+
+  component new_state_tree = MerkleTreeUpdate(depth);
+  new_state_tree.leaf <== selected_state_tree_hash;
+  for (var i = 0; i < depth; i++) {
+    new_state_tree.path_elements[i] <== selected_state_tree_path_elements[i];
+    new_state_tree.path_index[i] <== selected_state_tree_path_index[i];
   }
 
   new_state_tree_root <== new_state_tree.root;
```
weijiekoh: I'd suggest moving this line to the end of the circuit, since it's an output and it feels more intuitive that it would be located there there, like a return statement in a function would be.

test/testMACILogic.js
```diff
@@ -223,111 +237,164 @@ describe('MACI', () => {
...
-        existing_public_key: stringifyBigInts(user2PublicKey),
-        existing_state_tree_leaf: stringifyBigInts(user2Leaf),
-        ecdh_private_key: stringifyBigInts(ecdhPrivateKey)
-      }
+      const wasmBn128 = await buildBn128()
+      const zkSnark = groth
 
-      const witness = circuit.calculateWitness(circuitInput)
+      const circuit = new Circuit(updateStateTreeCircuitDef)
+      const witness = circuit.calculateWitness(circuitInputs)
```
weijiekoh: I think we should also get the output (the new state root) and test whether it matches our off-chain Merkle root.

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
 
-  // **** 1. Make sure the leaf exists in the cmd tree **** //
-  component cmd_tree_value_exists = LeafExists(levels);
-  cmd_tree_value_exists.root <== cmd_tree_root;
-  cmd_tree_value_exists.leaf <== encrypted_data_hash.hash;
-  for (var i = 0; i < levels; i++) {
-    cmd_tree_value_exists.path_elements[i] <== cmd_tree_path_elements[i];
-    cmd_tree_value_exists.path_index[i] <== cmd_tree_path_index[i];
+  for (var i = 0; i < message_length - 1; i++) {
+    decrypted_command.out[i] === command[i];
```
kobigurk: Style question - do we need the extra variable `command` and the extra constraints it entails? Can't we use decrypted command directly?

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
-    decrypted_data.message[i] <== encrypted_data[i];
+  // Check 5. Make sure the hash of the data corresponds to the 
+  //          existing leaf in the state tree
+  component existing_state_tree_leaf_hash = Hasher(state_tree_data_length);
+  existing_state_tree_leaf_hash.key <== 0;
+  for (var i = 0; i < state_tree_data_length; i++) {
+    existing_state_tree_leaf_hash.in[i] <== state_tree_data[i];
+  }
+  existing_state_tree_leaf_hash.hash === state_tree_leaf;
+
```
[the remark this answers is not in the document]
kobigurk: :+1:

circuits/verify_signature.circom
```diff
@@ -1,27 +1,123 @@
...
+    5299619240641551281634865583518297030282874472190772894086521144482721001553,
+    16950150798460657717958625567821834550301663161624707787222815936182638968203
+  ];
+  component mulFix = EscalarMulFix(253, BASE8);
+  for (i=0; i<253; i++) {
+    mulFix.e[i] <== snum2bits.out[i];
+  }
+
+  // Valid should equal to 0 if its valid
+  valid <== (mulFix.out[0] - addRight.xout) + (mulFix.out[1] - addRight.yout);
```
kobigurk: I think this might not be safe - I might be able to craft inputs that make this 0 while the actual constraint is not satisfied.
If you want both addends to be zero, it might be better to use `IsZero`.

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
+  for (var i = 0; i < state_tree_data_length; i++) {
+    new_state_tree_leaf.in[i] <== new_state_tree_data[i];
   }
 
-  // **** 3.2 Validate signature against existing public key **** //
-  component signature_verifier = VerifySignature(message_length);
+  // Checks to see if we should update the
+  // random leaf (Note: 0 is true, anything non-zero is false)
+  signal valid_signature;
+  valid_signature <-- signature_verifier.valid == 0 ? 0 : 1;
```
kobigurk: This doesn't generate a constraint, `IsZero` should be used here.

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
 
-  signature_verifier.from_x <== existing_public_key[0]; // public key x
-  signature_verifier.from_y <== existing_public_key[1]; // public key y
+  signal sufficient_vote_credits;
+  sufficient_vote_credits <-- new_vote_credits > 0 ? 0 : 1;
```
kobigurk: This doesn't generate a constraint, `IsZero` should be used here. In this case, you also want `> 0`, which probably means you want to limit the range of `new_vote_credits` (e.g., 32 bits).

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
 
-  signature_verifier.R8x <== decrypted_data.out[message_length]; // sig R8x
-  signature_verifier.R8y <== decrypted_data.out[message_length + 1]; // sig R8x
-  signature_verifier.S <== decrypted_data.out[message_length + 2]; // sig S
+  signal correct_nonce;
+  correct_nonce <-- decrypted_command.out[5] == (state_tree_data[4] + 1) ? 0 : 1;
```
kobigurk: This doesn't generate a constraint, `IsZero` should be used here (on the subtraction).

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
 
-  for (var i=0; i < message_length; i++) {
-    signature_verifier.preimage[i] <== decrypted_data.out[i];
+  signal valid_state_leaf_index;
+  valid_state_leaf_index <-- decrypted_command.out[0] <= state_tree_max_leaf_index ? 0 : 1;
```
kobigurk: I didn't know you could chain these :D

This doesn't generate a constraint, `IsZero` should be used here.

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
+
+  signal valid_vote_options_leaf_index;
+  valid_vote_options_leaf_index <-- decrypted_command.out[3] <= vote_options_max_leaf_index ? 0 : 1;
```
kobigurk: This doesn't generate a constraint, `IsZero` should be used here.

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
+  signal valid_update;
+  valid_update = valid_signature + sufficient_vote_credits + correct_nonce + valid_state_leaf_index + valid_vote_options_leaf_index;
+
+
+  // If any of the above checks return false
+  // We update the random leaf instead
+
+  // Note: valid_update 0 is true, 1 is false
+  signal selected_state_tree_hash;
+  selected_state_tree_hash <-- valid_update == 0 ? new_state_tree_leaf.hash : random_leaf;
```
kobigurk: This doesn't generate a constraint, `IsZero` should be used here.

circuits/update_state_tree.circom
kobigurk: This doesn't generate a constraint, `IsZero` should be used here.

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
 
-  // **** 4. If signature valid, update leaf **** //
-  component new_state_tree = MerkleTreeUpdate(levels);
-  new_state_tree.leaf <== encrypted_data_hash.hash;
-  for (var i = 0; i < levels; i++) {
-    new_state_tree.path_elements[i] <== state_tree_path_elements[i];
-    new_state_tree.path_index[i] <== state_tree_path_index[i];
+  signal selected_state_tree_path_index[depth];
+  for (var i = 0; i < depth; i++) {
+    selected_state_tree_path_index[i] <-- valid_update == 0 ? state_tree_path_index[i] : random_leaf_path_index[i];
```
kobigurk: This doesn't generate a constraint, `IsZero` should be used here.

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
+  }
+
+  // selected_state_tree_hash can only be the random leaf
+  // if the no_op flag is set to true (1)
+  signal no_op_valid_update;
+  no_op_valid_update <-- valid_update == 0 ? new_state_tree_leaf.hash : no_op * random_leaf;
```
kobigurk: This doesn't generate a constraint, `IsZero` should be used here.

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
+  
+  selected_state_tree_hash === no_op_valid_update;
```
kobigurk: To select the right hash, you could do something like `no_op * new_hash + (1-no_op) * old_hash`.

circuits/verify_signature.circom
```diff
@@ -1,27 +1,123 @@
...
+  component snum2bits = Num2Bits(253);
+  snum2bits.in <== S;
+
+  component compConstant = CompConstant(2736030358979909402780800718157159386076813972158567259200215660948447373040);
+
+  for (i=0; i<253; i++) {
+    snum2bits.out[i] ==> compConstant.in[i];
+  }
+  compConstant.in[253] <== 0;
+  compConstant.out === 0;
```
kobigurk: This should also be changed to a `no-op` style.

circuits/verify_signature.circom
kobigurk: Similar comment to above - I believe you should use `IsZero` for both addends, this might be susceptible to malicious inputs.

kobigurk:
An intuition - I believe you'd eventually end up with an array of "valid flags", which you'd then check are all  1, and that would feed into your final no_op flag for a leaf.

circuits/update_state_tree.circom
weijiekoh: It's an less-than-or-equal operator, not a chained signal assignment :)

kobigurk: :D Right!

circuits/update_state_tree.circom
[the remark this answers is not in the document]
kendricktan: Sounds good

circuits/update_state_tree.circom
```diff
@@ -1,106 +1,266 @@
...
+    new_vote_options_tree.path_index[i] <== vote_options_tree_path_index[i];
+  }
+
+  // Verify signature against existing public key
+  component signature_verifier = VerifySignature(message_length - 4);
+
+  signature_verifier.from_x <== state_tree_data[0]; // public key x
+  signature_verifier.from_y <== state_tree_data[1]; // public key y
+
+  signature_verifier.R8x <== decrypted_command.out[message_length - 4]; // sig R8x
```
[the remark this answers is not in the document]
kendricktan: Thats a good idea

circuits/update_state_tree.circom
kendricktan: @kobigurk we definitely can, the specs seems to want it tho. Maybe it can help double check if the coordinator managed to get similar decryption results to the circuit?

circuits/verify_signature.circom
kendricktan: Could you elaborate more on this?

circuits/update_state_tree.circom
kobigurk: I don't think it adds a double check, though let's chat about it :)

circuits/verify_signature.circom
kobigurk: Since S is supplied by the user, it can be a malicious input that will not be smaller than the constant and will cause the proof to fail.

PR #38, opened 2020-01-11
Nothing up my sleeve

kendricktan:
https://github.com/barryWhiteHat/maci/issues/36
https://github.com/barryWhiteHat/maci/issues/35


contracts/MerkleTree.sol
```diff
@@ -59,6 +59,12 @@ contract MerkleTree is Whitelist {
       uint256 _zeroValue,
```
weijiekoh: We can remove this parameter since the contract computes the `ZERO_VALUE` anyway.

weijiekoh:
Besides the one comment above, everything looks great!

PR #43, opened 2020-01-14
Batch circuit update for state tree

kendricktan:
A batch circuit update with tests for valid inputs

### DIFF
--- a/circuits/batch_update_state_tree.circom
+++ b/circuits/batch_update_state_tree.circom
@@ -11,7 +11,7 @@ template BatchUpdateStateTree(
   //    batch_size: that we're processing
 
   // Output: New state tree root
-  signal output new_state_tree_root;
+  signal output root;
 
   // Input(s)
   signal input coordinator_public_key[2];
@@ -39,6 +39,12 @@ template BatchUpdateStateTree(
   signal private input random_leaf_path_elements[depth];
   signal private input random_leaf_path_index[depth];
 
+  // Root when random leaf is inserted
+  // As the random leaf is inserted at th end,
+  // the random_leaf_root is the end root after all the commands
+  // has been processed
+  signal private input random_leaf_root;
+
   // State tree
   var state_tree_data_length = 5;
   signal private input state_tree_data_raw[batch_size][state_tree_data_length];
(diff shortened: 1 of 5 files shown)


kendricktan:
I'll wait for @weijiekoh's conversion to Jest and TS before beginning any significant update and changes to the logic (to prevent double work)

circuits/batch_update_state_tree.circom
```diff
@@ -0,0 +1,124 @@
...
+
+  // Input(s)
+  signal input coordinator_public_key[2];
+
+  var message_length = 11;
+  signal input message[batch_size][message_length];
+
+  // Select vote option index's weight
+  // (a.k.a the raw value of the leaf pre-hash)
+  signal private input vote_options_leaf_raw[batch_size];
```
kobigurk: This is also supplied by the coordinator, right?

circuits/batch_update_state_tree.circom
```diff
@@ -0,0 +1,124 @@
...
+  // Vote options tree root (supplied by coordinator)
+  signal private input vote_options_tree_root[batch_size];
+  signal private input vote_options_tree_path_elements[batch_size][vote_options_tree_depth];
+  signal private input vote_options_tree_path_index[batch_size][vote_options_tree_depth];
+  signal input vote_options_max_leaf_index;
+
+  // Message tree
+  signal input msg_tree_root;
+  signal input msg_tree_path_elements[batch_size][depth];
+  signal input msg_tree_path_index[batch_size][depth];
```
kobigurk: How about instead of heaving `msg_tree_path_index`, this would be calculated?
We'd have "initial index" and then we continue one by one until batch size? We'd need to then transform each index into binary, but that feels better than having all of these public inputs

Also, can `msg_tree_path_elements` be private inputs?

circuits/batch_update_state_tree.circom
```diff
@@ -0,0 +1,124 @@
...
+    // Shared Keys
+    new_state_tree[i].ecdh_private_key <== ecdh_private_key;
+
+    new_state_tree[i].ecdh_public_key[0] <== ecdh_public_key[i][0];
+    new_state_tree[i].ecdh_public_key[1] <== ecdh_public_key[i][1];
+  }
+
+  // Update random leaf at index 0
+  component final_state_tree = MerkleTreeUpdate(depth);
+  final_state_tree.leaf <== random_leaf
```
kobigurk: Missing semicolon :)

circuits/batch_update_state_tree.circom
```diff
@@ -0,0 +1,124 @@
...
+
+  // Message tree
+  signal input msg_tree_root;
+  signal input msg_tree_path_elements[batch_size][depth];
+  signal input msg_tree_path_index[batch_size][depth];
+
+  // Random leaf (updated every batch)
+  signal private input random_leaf;
+  signal private input random_leaf_path_elements[depth];
+  signal private input random_leaf_path_index[depth];
```
kobigurk: Isn't random leaf index always the last index? If yes, we can just use a locally calculated value instead of a signal (which might also cause some problems if you use an index that is different than the last one).

circuits/batch_update_state_tree.circom
kendricktan: Yeap! Anything to do with `vote_options` is supplied by the coordinator

circuits/batch_update_state_tree.circom
kendricktan: fair point!

circuits/batch_update_state_tree.circom
kendricktan: Regarding `msg_tree_path_elements`, why does it need to be private inputs? As the msg tree can be reconstructed by anyone

circuits/batch_update_state_tree.circom
```diff
@@ -0,0 +1,124 @@
...
+    new_state_tree[i].vote_options_tree_root <== vote_options_tree_root[i];
+    for (var j = 0; j < vote_options_tree_depth; j++) {
+      new_state_tree[i].vote_options_tree_path_elements[j] <== vote_options_tree_path_elements[i][j];
+      new_state_tree[i].vote_options_tree_path_index[j] <== vote_options_tree_path_index[i][j];
+    }
+    new_state_tree[i].vote_options_max_leaf_index <== vote_options_max_leaf_index;
+
+    // Message Tree
+    new_state_tree[i].msg_tree_root <== msg_tree_root;
+    for (var j = 0; j < depth; j++) {
```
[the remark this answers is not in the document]
kendricktan: It'll be logically cleaner by separating them out imo

circuits/batch_update_state_tree.circom
kobigurk: Yep, it's not for privacy! It's for having less public inputs if not needed. That's pretty expensive on chain

PR #44, opened 2020-01-22
Lerna submodules, Jest, Typescript, and Etherlime

weijiekoh:
This large PR contains:

- Submodules (using Lerna) to separate concerns (e.g contract code and tests are in their own directory, and the same for circuits, crypto, etc)
- A port to Typescript, which allows us to strongly type all our code, and allow us to borrow audited code from Semaphore in the future
- Tests ported to Jest, which has good Typescript support
- Standardised and typed cryptographic operations, which allow us to not worry about low-level cryptographic operations (e.g. `mimicsponge` and `mimc7` functions are encapsulated as just `hashOne`, `hash`, `encrypt`, or `decrypt` functions and the developer doesn't need to think about what goes on under the hood).
- [Domain objects](https://wiki.c2.com/?DomainObject), which are high-level representations of fundamental concepts (e.g. Commands, Messages, and State Leaves) with methods which encapsulate low-level details (e.g. signing, encryption, etc)
- A contract deployment script which uses Etherslime and works on local testnets and Kovan
- More cryptographic and circuit tests
- More documentation
- Updated `LICENSE` file with the current year and authors
- A test for the Hasher contract
- A script (`circuits/ts/buildSnarks.ts`) that performs a trusted setup and Solidity verifier generation. This helps local development and CI. See the readme for the npm commands to use for the two circuits we have.


contracts/ts/__tests__/Maci.test.ts
```diff
@@ -0,0 +1,243 @@
...
+        const contracts = await deployAllContracts(deployer)
+        maciContract = contracts.maciContract
+    })
+
+    it('should be deployed', async () => {
+        expect(maciContract.contractAddress.startsWith('0x')).toBeTruthy()
+    })
+})
+
+//const {
```
barryWhiteHat: probably a good idea to either delete the comments or get that code workingl

weijiekoh: It's a WIP :smile: so we can merge this for it to not be a blocker.

barryWhiteHat: :+1:

kendricktan:
Sorry for the late reply, but it seems running `lerna build` returns a `Cannot find module 'maci-crypto'.`

I think having some documentation and updating the README.md would be nice on how to run the repository :)

PR #47, opened 2020-01-30
[contracts] Sign-up logic

weijiekoh:
According to the spec:

> **`signUp(EddsaPubKey _pubKey) payable`**
> During a sign-up period, any user who owns a recgonised ERC721 token can invoke signUp(). This function tracks the ERC721 token ID and prevents double sign-ups with the same token ID.
> 
> Next, it adds a new leaf to the state tree, starting from index 1 (as index 0 is reserved for invalid leaves). This leaf is the hash of the public key, the user's voice credits, the nonce 0, and the root of an empty vote option tree.
> 
> Additionally, this function requires a deposit of depositAmt ETH to discourage users from sharing their EdDSA private keys with potential bribers.
> 
> The sign-up period ends after a predefined deadline. A later version of MACI will allow ongoing sign-ups where state trees will be merged once per week.

This PR implements the above, except the deposit requirement as that deserves further discussion.

**`SignUpGatekeeper`**

The logic that governs whether a user can sign up is abstracted away into a `SignUpGatekeeper` contract with an abstract `register(address _user, bytes memory _data)`, which can be derived by child contracts such as the `SignUpTokenGatekeeper.sol`. Child contracts which implement `register()` should make it do the following:

1. Prevent double registrations
2. Allow registrations according to custom logic

`SignUpTokenGatekeeper.sol` implements this custom logic by first ABI-decoding `_data` into a `uint256` token ID, and then checks if `_user` owns said token ID.

Very similar (or identical) logic will be implemented in a future `POAPGatekeeper` contract.

 **`MACI.constructor`**

The MACI constructor atomically initialises the command tree and state tree. This is necessary (instead of deploying them separately) as we can replace the MerkleTree whitelist with the OZ Ownable, and also save gas.

It also stores a blank vote option tree root (based on hardcoded values for said blank roots in `MerkleTree.sol`). This saves a lot of gas, but it may be better to just pay for the optics of generating and storing the Merkle root programmatically. Let's discuss!

 **`MACI.signUp`**

This function calls `signUpGatekeeper.register`, and if that function does not revert, inserts a fresh state leaf into the state tree.

**`DomainObjs`**

This contract provides the Solidity equivalent of the `maci-domainobjs` submodule. For instance, it represents each state leaf as a struct `StateLeaf` and provides a convenience function `hashStateLeaf(StateLeaf memory _stateLeaf)` that abstracts over hashing its constituent values.

**Misc**

There are also miscellaneous updates (e.g. `maci-crypto` now exports the nothing-up-my-sleeve value).


contracts/sol/MACI.sol
```diff
@@ -1,191 +1,200 @@
...
+    public {
 
-    // Gets coordinator public key
-    function getCoordinatorPublicKey() public view returns (uint256, uint256) {
-      return (coordinatorPublicKeyX, coordinatorPublicKeyY);
+        // Register the user via the sign-up gatekeeper. This function should
+        // throw if the user has already registered or if ineligible to do so.
+        signUpGatekeeper.register(msg.sender, _signUpGatekeeperData);
+
+        // Create, hahs, and insert a fresh state leaf
```
kobigurk: hash :)

contracts/sol/MerkleTree.sol
```diff
@@ -185,3 +194,50 @@ contract MerkleTree is Ownable, Hasher {
         return nextLeafIndex;
     }
 }
+
+contract EmptyMerkleTreeRoots {
```
kobigurk: How expensive is it to calculate in the contract?
I don't think it's critical, it seems fine to me either way, just not sure it's worth the savings.

weijiekoh: Using a pre-calculated root: 4730825 gas

Using a on-chain calculated root (for a 4-level vote option tree): 5171319 gas

10 levels: 5755717 gas
20 levels: 6729607 gas

Good point, I think it's not worth the savings!

PR #56, opened 2020-02-06
ci

barryWhiteHat:
Adding CI


kendricktan:
Project currently uses typescript and requires some experimental features from `experimental-worker` might need to use a newer version of node (v12.x.x)

PR #59, opened 2020-02-11
[contracts] Batch process message function and tests

weijiekoh:
At the core of this PR is a Solidity implementation of `batchProcessMessage()`.

After all sign-ups and during the first invocation of `publishMessage()`, we store the state tree root as `postSignUpStateRoot` and subsequently ignore the state tree as it is no longer needed. After all, all that `batchProcessMessage()` needs to do is update the state root, not perform insertions.

When the coordinator processes the message, for each batch, if the snark proof is valid, `batchProcessMessage()` will update `postSignUpStateRoot`. 

The contract test includes proof generation (via convenience functions from `libsemaphore`) and off-chain verification.

There is a new submodule (`maci-core`) which includes business logic functions, particularly `processMessage()`. Later PRs will add vote tallying functions to that module. It is a separate module as it's meant to be shared by circuit, contract, integration tests, and the coordinator server.

There are also miscellaneous changes:

- The snark inputs `message` and `msg_tree_path_elements` in the batch UST circuit are now private, saving us a lot of gas
- We now use the audited groth verifier.sol from https://github.com/weijiekoh/snarkjs.git#52e0ceba0256bc9bf94cbc9500139dfc37018f91
- `maci-circuits` now has a `compileAndLoadCircuit()` convenience function
- Shifted the `merkleTrees` entry in config to `maci` as that makes more sense
- There is a new test for the Merkle tree contract's `insertBlankAtZerothLeaf()` function.
- The `Keypair` domain object constructor can now accept a raw private key if the developer wishes to override its default behaviour which is to generate a random one.


contracts/sol/MACI.sol
```diff
@@ -191,20 +239,138 @@ contract MACI is Ownable, DomainObjs {
...
+    function genBatchUstPublicSignals(
+        uint256 _newStateRoot,
+        uint256[] memory _stateTreeRoots,
+        PubKey[] memory _ecdhPubKeys
+    ) public view returns (uint256[19] memory) {
+        uint256[19] memory publicSignals;
+        publicSignals[0] = _newStateRoot;
+        publicSignals[1] = coordinatorPubKey.x;
+        publicSignals[2] = coordinatorPubKey.y;
+        publicSignals[3] = voteOptionsMaxLeafIndex;
```
barryWhiteHat: This does not need to be public input you can just set it in the snark circuit at compile time.

contracts/sol/MACI.sol
```diff
@@ -179,8 +210,25 @@ contract MACI is Ownable, DomainObjs {
...
+            // It is exceedingly improbable that the zero value is a tree root
+            assert(postSignUpStateRoot != stateTree.getRoot());
+
+            postSignUpStateRoot = stateTree.getRoot();
+
+            // This is exceedingly unlikely to occur
+            assert(postSignUpStateRoot != 0);
+
+            // Destroy the state tree contract
+            stateTree.selfDestruct();
```
barryWhiteHat: I would rather not destory the stateTree because in future we will want to allow signup for ever. But its fine for now.

contracts/sol/MACI.sol
```diff
@@ -191,20 +239,138 @@ contract MACI is Ownable, DomainObjs {
...
+
+        require(isValid == true, "MACI: invalid batch UST proof");
+
+        // Increase the message batch start index to ensure that each message
+        // batch is processed in order
+        messageBatchStartIndex += messageBatchSize;
+
+        // The message batch start index should never exceed the maximum
+        // allowed value
+        assert(messageBatchStartIndex <= messageTreeMaxLeafIndex - messageBatchSize);
```
barryWhiteHat: I think its okay to allow the coordinator to process zero leaves at the end if they want. I think its more dangerous having this check here and not being able to process the last few votes.

contracts/sol/MerkleTree.sol
```diff
@@ -78,8 +78,9 @@ contract MerkleTree is Ownable, Hasher {
     }
 
 
-    // Inserts (appends) a new leaf into the
-    // merkle tree
+    /*
+     * Insert a new leaf
+     */
     function insert(uint256 leaf) public onlyOwner {
         require(nextLeafIndex < maxLeafIndex, "Merkle Tree at max capacity");
```
barryWhiteHat: Concerned an attack can fill up the command_tree but probably okay to just make the command tree really big.

But i think this only applies to the deposit and not the message insert.

contracts/sol/MerkleTree.sol
```diff
@@ -173,6 +174,13 @@ contract MerkleTree is Ownable, Hasher {
         emit LeafUpdated(leaf, leafIndex);
     }
 
+    /*
+     * Destroy this contract to free up space
+     */
+    function selfDestruct() public onlyOwner {
+        selfdestruct(0x0000000000000000000000000000000000000000);
```
barryWhiteHat: Scared to use this every :P

contracts/sol/MACI.sol
weijiekoh: Let's say the vote option tree has 4 levels (max 16 leaves). However, there are only 10 options to vote for (e.g. if you want to vote for what to buy at a restaurant, the menu may only have 10 items).

**Scenario A**

Alice votes for option #0 (which is valid) and Bob votes for option 15 (which is also valid). Both votes are counted.

At the end of the day, the coordinator tallies the votes and ignores votes for options 10-15, so only Alice's vote has an effect.

**Scenario B**

Alice votes for option #0 (which is valid) and Bob votes for option 15 (which is invalid).

Alice's vote is counted, but Bob's vote is a no-op.

At the end of the day, the result is the same as Scenario A.

However, Scenario B will allow us to spend less gas, since we can reduce the number of public inputs by 1. As such, in the grand scheme of things, I think we should go with Scenario B.

barryWhiteHat: Well if a user wants to charge you gas they can put noop votes on chain that you are required to process. So don't agree with argument about reducing gas costs.  

NOTE : I think if we want to reduce gas we should make an optimisitic version of MACI.

weijiekoh: Correction: Scenario A (not B) allows us to spend less gas.

But after some discussion, we've decided to go with B.

contracts/sol/MACI.sol
weijiekoh: This check does not prevent the coordinator from processing zero-value leaves at the end. Instead, it prevents this function from being called more times than is acceptable.

E.g. if there is a message tree of maximum index 16, and the batch size is 4, this check ensures that `messageBatchStartIndex` never exceeds 12.

contracts/sol/MerkleTree.sol
weijiekoh: Will remove :smile:

contracts/sol/MACI.sol
weijiekoh: Will remove :smile:

PR #79, opened 2020-04-20
Spec review 1: before state root transition proof circuit

ChihChengLiang:
Try to give an update on spec.

Photo by David Clode on Unsplash

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -174,7 +174,7 @@ It also increments the message tree index by the number of commands whose proces
 
 This function should, however, only do so if the processed message counter indicates that all previous messages have already been processed.
 
-Only the coordinator may invoke this function?
+Although anyone may call this contract function, only the coordinator should know the ECDH shared keys used to encrypt the messages.
 
 #### `proveVoteTallyBatch()`
 
@@ -191,7 +191,7 @@ This allows the coordinator to prove the correctness of their vote tally (in `_f
 
 `_finalSaltedResults` can be any value but for the final batch, it must be the correct quadratic vote tally.
 
-Only the coordinator may invoke this function.
+It does not matter that the contract does or does not restrict access to this function as anyone who can produce a valid proof should be able to tally the votes, and it should not be possible for anyone to tamper with the results anyway.
 
 ### State leaves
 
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -174,7 +174,7 @@ It also increments the message tree index by the number of commands whose proces
 
 This function should, however, only do so if the processed message counter indicates that all previous messages have already been processed.
 
-Only the coordinator may invoke this function.
+Only the coordinator may invoke this function?
```
ChihChengLiang, line 177: Looks like only the coordinator can successfully invoke this function because it requires a proof. But we don't have an explicitly `isCoordinator` modifier in the `MACI.sol` contract though.

SPEC.md
```diff
@@ -187,12 +187,12 @@ uint256[] memory _finalSaltedResults,
 uint256[8] memory _proof
 ```
 
-where `n` is the number of vote options.
```
ChihChengLiang, line 190: Removed for no reference to `n` in this method

SPEC.md
```diff
@@ -187,12 +187,12 @@ uint256[] memory _finalSaltedResults,
...
-
 This allows the coordinator to prove the correctness of their vote tally (in `_finalSaltedResults`). They do this in batches of state leaves. Each batch of state leaves is accumulated into an intermediate state root, and the Merkle root of all the intermediate state roots is the full state root. The proof shows that the result of adding the votes in the current batch to the culmulative results is computed correctly, but hides the results by salting and hashing them.
 
 `_finalSaltedResults` can be any value but for the final batch, it must be the correct quadratic vote tally.
 
+Only the coordinator may invoke this function.
```
ChihChengLiang, line 194: I suppose `proveVoteTallyBatch` is coordinator invokable for the same reason with `batchProcessMessage`

SPEC.md
```diff
@@ -276,41 +276,51 @@ Note that the circuit pseudocode in this specification does not describe zk-SNAR
...
-| `msgTreePathElements` | Private | The Merkle path elements to the message tree root from the message leaf | Coordinator |
-| `randomLeaf` | Private | Random data | Coordinator |
-| `newStateTreePathIndex` | Private |The Merkle path index to the new state root from the new state leaf | Coordinator |
-| `newStateTreePathElements` | Private |The Merkle path elements to the new state root from the new state leaf | Coordinator |
-| `newStateTreePathElementsToZero` | Private |The Merkle path elements to the new state root from leaf 0, **after** the new state leaf has been updated | Coordinator |
-| `userCurrentLeaf` | Private | The user's latest valid state leaf | Coordinator |
-| `command` | Private | The command to process. Includes all the details in the leaf. | Coordinator |
-| `noOp` | Private | The no-op flag | Coordinator |
-| `userPubKey` | Private | The user's public key | Coordinator |
-| `encPubKey` | Private | The ephermeral public key | Contract |
```
ChihChengLiang, line 304: I was confused by this. If I understand correctly,  `userPubKey` is published by the user via contract, collected by the coordinator to generate a shared key. The `encPubKey` should be the coordinator's private key that fed into the circuit as a private message.

SPEC.md
weijiekoh: I'll update the spec to address this point as well.

SPEC.md
weijiekoh: I'll update the spec to address this point!

SPEC.md
weijiekoh: To encrypt the command, the user creates an ephemeral keypair, and generates a shared key using `encPrivKey` and `coordinatorPubKey`. The coordinator can recreate this key using `encPubKey` and `coordinatorPrivKey`.

The user's ephemeral keypair's public key is different from the one which they store in the state tree.

PR #80, opened 2020-04-21
[presentation] Split the spec into smaller chunks

ChihChengLiang:
Not sure if you all like this idea, splitting it into smaller chunks for better readability and maintainability.


weijiekoh:
Thank you so much! This is super helpful as the spec was really long!

Issue #109, opened 2020-05-27
[circuits] Negative votes in MACI


ChihChengLiang:
## Motivation

In the book "Radical Markets" page 119, negative votes are proposed as a cheaper way to let a voter's favorite candidate stands out from a less favorable one.

To show why it's cheaper, consider a 2 candidate situation, and a voter Alice prefers Catherine over Carl. To make Catherine win Carl by 2 votes, Alice could

- vote +2 for Catherine, which costs 4 voice credits, or
- vote +1 for Catherine and -1 for Carl, which costs 2 voice credits only.

The result is useful in the situation that you have 3 candidates, the first two are widely loved by one camp of voters and hated by another, and the third is less hated but also not popular enough.

In the positive vote only situation, the voters would strategically vote the first two but not the third, to prevent the hated one winning.

But with the negative vote situation, the votes expect the first two get almost equally positive and negative votes and the third one could stand out. The voting dynamic is studied in [this paper](https://www.ssrn.com/abstract=3319508).

## Possible design choices in MACI

Thanks to @weijie for writing this section.

Two ways are possible to implement negative votes in MACI. We demonstrate this with examples.

Alice has 99 voice credits
Bob has 99 voice credits

Vote options: party A and party B

### Possibility 1 (hides information. e.g. the fact that a vote is polarised)

Bob votes for party A with 81 positive voice credits.
Party A now has +9 votes.

Alice does not like party A. She votes for party A with 36 negative voice credits.
Party A now has (+9) + (-6) = +3 votes.

Final tally = [3, 0]

### Possibility 2 (shows information, e.g. the fact that a vote is polarised)

Bob votes for party A with 81 positive voice credits.
Party A now has 9 positive votes and 0 negative votes.

Alice does not like party A. She votes for party A with 36 negative voice credits.

Party A now has 9 positive votes and 6 negative votes.

Final tally = [(9, 6), (0, 0)]

### Discussion

In Possibility 1 we have to worry about the overflow/underflow. Say, if party A only has -6 votes, what would the result look like? 

The Possibility 2 we tally the positive vote and negative votes in a different array. We don't have to worry about the overflow/underflow in this case.

### Circuit is happy with negative inputs?

Consider this circuit. Input `{ 'a': -1, 'b': 4 }` would gives witness `[ 1n, -4n, -1n, 4n ]`

```
 template Multiplier() {
       signal private input a;
       signal private input b;
       signal output c;
       
       c <== a*b;
   }

[...]

PR #112, opened 2020-05-31
Negative votes

weijiekoh:
This PR implements negative votes. Addresses #109.

Before this PR, each Command contained a new vote weight field which counts entirely as positive votes. Now, we use this field to encode positive and negative votes. The final vote tally will show the number of positive votes and the number of negative votes per vote option. We include both values (instead of subtracting the negative votes from the positive votes) so that we can retain information about how contentious a vote is. 

### VoteLeaf

There is now a `VoteLeaf` domain object which represents a vote option tree leaf.

Each vote option tree leaf has a `pos` and a `neg` value. This allows the final tally to show not only the vote values, but also how contentious a vote is.

There is a function `pack()` which converts a `VoteLeaf` into a `SnarkBigInt`, and `unpack()` which does the opposite.

`pack()` encodes a VoteLeaf this way:

e.g. if `leaf.pos = 0xabcd` and `leaf.neg = 0x1234`, `leaf.pack()` returns `0xabcd0000000000000000000000000001234`

### Command

The `Command` used to contain a `newVoteWeight` field. That has been replaced by `vote` which is a `VoteLeaf`.

### User

The `User`'s `votes` attribute used to have the `SnarkBigInt[]` type. Now, its type is `VoteLeaf[]`.

### MaciState

`processMessage()` has been modified to account for positive and negative votes.

`genQuadVoteTallyCircuitInputs()` has been modified to pass in the `pack`ed vote leaves.

`genUpdateStateTreeCircuitInputs()` will pass in the `pack`ed vote leaf as the `vote_options_leaf_raw` circuit input.

## Circuits

The UpdateStateTree circuit will perform a no-op if the `pack`ed vote leaf's `pos` and/or `neg` value is out of range.

There is a new `UnpackVoteLeaf` circuit which accepts a `pack`ed value and outputs a `pos` value and a `neg` value. It also ensures that the packed value is within range. This range check should not interfere with the UpdateStateTree circuit's no-op capability.

## CLI

The `publish` subcommand now accepts negative votes using `-v`. The flag for positive votes remains `-w`. The vote option index flag has been renamed to `-vi`.

The JSON file that the `tally` subcommand outputs now contains both the positive and negative results (`"tally": :[ [pos, neg], ... ] ...`)

## Integration tests

Updated the test suite JSON files to contain pos+neg votes. Also updated the `expectedTally` fields.


integrationTests/ts/__tests__/suites/suite0_happy.json
```diff
@@ -6,93 +6,110 @@
...
+        [1, 0],
+        [1, 0],
+        [1, 0],
+        [1, 0],
+        [1, 0],
+        [1, 0],
+        [1, 0],
+        [1, 0],
+        [0, 1],
+        [0, 0]
```
ChihChengLiang, line 113: It would be nice if we highlight that this is the 15th vote option and no user has voted for it yet.
```suggestion
        [0, 0] # No one votes for voteOptionIndex 15
```

integrationTests/ts/__tests__/suites.ts
```diff
@@ -282,8 +285,9 @@ const executeSuite = async (data: any, expect: any) => {
     const salt = bigInt(tallyRegMatch[2])
 
     const finalTallyCommitment = await maciContract.currentResultsCommitment()
+    debugger
```
ChihChengLiang, line 288: Is this accidentally left here?
```suggestion
```

circuits/circom/voteLeaf.circom
```diff
@@ -0,0 +1,55 @@
...
+
+    // Pack p and n and check that they match packedLeaf
+    packedLeaf === p * POW + n;
+
+    // Wire p and n to the outputs
+    pos <== p;
+    neg <== n;
+}
+
+template CalculateSquaredVoteLeaf() {
```
ChihChengLiang, line 41: This CalculateSquaredVoteLeaf solution is beautiful ✨ 
It avoids doing byte manipulation, range checks, and squares in updateStateTree.circom

weijiekoh: Thank you!

integrationTests/ts/__tests__/suites.ts
weijiekoh: Yes! Thank you for spotting it!

integrationTests/ts/__tests__/suites/suite0_happy.json
weijiekoh: Great idea! I've added a comment to the file (using a new field, as JSON doesn't support comments).

weijiekoh:
FYI, I'd like to resolve the conflicts when we implement the other features we've planned for the next version of MACI.

weijiekoh:
Closed in favour of https://github.com/appliedzkp/maci/pull/283

PR #114, opened 2020-06-06
Maci parameters

xuhcc:
At clr.fund we created the [MACI factory](https://github.com/clrfund/monorepo/blob/master/contracts/contracts/MACIFactory.sol) contract to automate MACI deployment.

That required MACI parameter structs which were defined in `MACI.sol` contract, so I moved them to a separate contract in order to make them accessible through inheritance. I also moved `PubKey` struct out of `DomainObjs.sol` to reduce the size of our contracts.


weijiekoh:
Thank you! I will merge this once the CircleCI checks pass. https://circleci.com/gh/appliedzkp/maci/tree/feat%2Fmaci-parameters

PR #121, opened 2020-06-25
Various minor improvements

xuhcc:
Fixed failing tests and lint errors, improved readme, etc.


contracts/scripts/compileSol.sh
```diff
@@ -12,7 +12,7 @@ mkdir -p ./compiled/abis
 
 # Copy the Semaphore contracts from the submodule into solidity/
 
-npx etherlime compile --solcVersion=native --buildDirectory=compiled --workingDirectory=sol --exportAbi --runs 200
+npx etherlime compile --solcVersion=0.5.16 --buildDirectory=compiled --workingDirectory=sol --exportAbi --runs 200
```
weijiekoh, line 15:
```suggestion
if [[ $* == *--native* ]]; then
    npx etherlime compile --solcVersion=native --buildDirectory=compiled --workingDirectory=sol --exportAbi --runs 200
else
    npx etherlime compile --solcVersion=0.5.16 --buildDirectory=compiled --workingDirectory=sol --exportAbi --runs 200
fi
```

I suggest giving the developer the option to use the `native` flag as the solcjs compiler is slower than the solc binary.

circuits/scripts/buildBatchUpdateStateTreeSnark.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 
+set -e
+
 cd "$(dirname "$0")"
 cd ..
 mkdir -p build
 
-node build/buildSnarks.js -i circom/prod/batchUpdateStateTreeVerifier.circom -j build/batchUstCircuit.json -p build/batchUstPk.json -v build/batchUstVk.json -s build/BatchUpdateStateTreeVerifier.sol -vs BatchUpdateStateTreeVerifier -m build/batchUst.params -z ~/.cargo/bin/zkutil
+node build/buildSnarks.js -i circom/prod/batchUpdateStateTreeVerifier.circom -j build/batchUstCircuit.json -p build/batchUstPk.json -v build/batchUstVk.json -s build/BatchUpdateStateTreeVerifier.sol -vs BatchUpdateStateTreeVerifier -m build/batchUst.params
```
weijiekoh, line 9: The `buildSnarks` command will fail without specifying `-z`.

weijiekoh: Ah, never mind, I see that the default is set in the config!

contracts/scripts/compileSol.sh
xuhcc: Ah, sure! I didn't know about that.
Though I'd like to write it differently: 81f85f1af667cb14bfacd058ffd1b8a09e219dad

PR #132, opened 2020-07-12
Fixed incorrect Merkle proof verification logic in VerifyTally.computeMerkleRootFromPath

weijiekoh:
Thanks @NIC619 for spotting this bug!

This PR fixes the way the on-chain tally verification function computes the index per tree level.


weijiekoh:
@xuhcc FYI, this is a crucial bugfix for MACI.

PR #141, opened 2020-07-20
In proveTallyBatch(), range-check each public signal input

weijiekoh:
This is an important security check that was inadvertently left out of `proveTallyBatch()`.


weijiekoh:
@xuhcc FYI - I will also update the npm package.

PR #153, opened 2020-07-29
Don't inherit from Ownable

xuhcc:
There are no `onlyOwner` methods in MACI contract, so I propose to remove `Ownable` from the list of base classes. That would save ~800 bytes in contract size.


weijiekoh:
Thank you!

PR #159, opened 2020-08-27
Upgrade snarkjs, circom, and circomlib versions, the Poseidon hash function implementation, and all submodules to v0.2.0

weijiekoh:
This PR implements two changes:

1. EdDSA signature generation now uses Poseidon instead of MiMC.

2. The Poseidon hash now uses the latest implementation in `circomlib` (commit [`01e3f0d6807137fb01c085b88cfaf63661e4a946`](https://github.com/iden3/circomlib/commit/01e3f0d6807137fb01c085b88cfaf63661e4a946)) from [PR 47](https://github.com/iden3/circomlib/pull/47). Note that this doesn't use the latest `circomlib` commit (which has an `assert` statement in `comparators.circom` as that requires a later version of `circom` and I haven't upgraded yet.

The new Poseidon hash implementation uses the parameters defined [here](https://github.com/tornadocash/circomlib/blob/1490baea4fbb6e58eb8e93ef826fcbb61860e74b/src/poseidon.js#L25).

```
const N_ROUNDS_F = 8;
const N_ROUNDS_P = [56, 57, 56, 60, 60, 63, 64, 63];

const t = inputs.length + 1;
const nRoundsF = N_ROUNDS_F;
const nRoundsP = N_ROUNDS_P[t - 2];
```

Since the new implementation and constants are based on an [updated version of the Poseidon paper](https://eprint.iacr.org/2019/458.pdf), I'm inclined to stick to them instead of the parameters we used previously.

@ChihChengLiang @kobigurk I'd love to hear your thoughts about this. Thank you!


kobigurk:
Sounds good to me!

Are there test vectors from the paper the implementation checks against?

weijiekoh:
The tests are upstream here: https://github.com/iden3/circomlib/blame/01e3f0d6807137fb01c085b88cfaf63661e4a946/test/poseidoncircuit.js

I'll run the reference implementation and see if the hardcoded results match. e.g:

```
const res2 = poseidon([1,2]);
assert.equal("17117985411748610629288516079940078114952304104811071254131751175361957805920", res2.toString());
```

circuits/scripts/installZkutil.sh
```diff
@@ -15,4 +15,4 @@ chmod a+x rust.sh
 sh rust.sh -y
 rm rust.sh
 source $HOME/.cargo/env
-cargo install zkutil --version 0.2.1
+cargo install zkutil --version 0.3.1
```
xuhcc, line 18: zkutil version (0.2.1) is also mentioned in main README.md

weijiekoh:
Thanks @xuhcc!

BTW I'm still trying to get the CircleCI tests to work. Their VM mysteriously kills the circuit compilation step (using `circom`). I tested it via SSH and the VM seems to be throttling processes which consume a lot of CPU for some time (between 1 to 2 minutes).

```
circleci@50a5e6dc6861:~/maci/circuits$ echo 'include "../batchUpdateStateTree.circom"; component main = BatchUpdateStateTree(4, 2, 2, 4);' > circom/prod/test.circom       
circleci@50a5e6dc6861:~/maci/circuits$ NODE_OPTIONS=--max-old-space-size=4096 node ./node_modules/circom/cli.js circom/prod/test.circom -o build/batchUstCircuit.json
```

The above works.

```
circleci@50a5e6dc6861:~/maci/circuits$ echo 'include "../batchUpdateStateTree.circom"; component main = BatchUpdateStateTree(4, 3, 2, 4);' > circom/prod/test.circom       
circleci@50a5e6dc6861:~/maci/circuits$ NODE_OPTIONS=--max-old-space-size=4096 node ./node_modules/circom/cli.js circom/prod/test.circom -o build/batchUstCircuit.json      
Killed
```
The above doesn't work. Still a WIP...

weijiekoh:
FYI, the SageMath reference implementation seems to be incompatible with `circomlib`'s. The former requires the number of input words to be the same as t, the latter requires it to be t-1.

Also, the number of round constants in the [SageMath implementation](https://extgit.iaik.tugraz.at/krypto/hadeshash/-/blob/master/code/poseidonperm_x5_254_3.sage) is different from that in circomlib. `poseidonperm_x5_254_3.sage` has 192 round constants but circomlib has 195 for t = 3.

crypto/ts/index.ts
```diff
@@ -60,49 +58,16 @@ const buffer2BigInt = (b: Buffer): BigInt => {
...
- */
-const hash = (plaintext: Plaintext): SnarkBigInt => {
-
-    return mimcsponge.multiHash(plaintext, 0, 1)
+const poseidonT6 = (inputs: SnarkBigInt[]) => {
+    assert(inputs.length === 5)
+    return poseidon(inputs)
 }
 
 const hash5 = (elements: Plaintext): SnarkBigInt => {
```
ChihChengLiang, line 73: We can remove `hash5` or try to add paddings in it.

ChihChengLiang (approved):
Looks good to me. Commented a small heads-up on hash5.

The new circomlib Poseidon API looks more beautiful now.

weijiekoh:
I got the SageMath reference implementation to output the same hardcoded test values from `circomlib`.

First, I had to generate new round constants and MDS matrices per `t` value. In this example, I used `t=3` and `t=6`.

```
sage generate_parameters_grain.sage 1 0 254 3 8 57 0x30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000001 > t3.txt
sage generate_parameters_grain.sage 1 0 254 6 8 60 0x30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000001 > t6.txt
```

Next, I copied and pasted the respective round constants and MDS matrix to the following Sage files:

`poseidonperm_x5_254_3.sage`
`poseidonperm_x5_254_5.sage`

For each test case, I modified the input words. For instance, to check `circomlib`'s test of `t=3` for `[1, 2]`, I used:

```
input_words = [] :
for i in [1, 2, 0]:
    input_words.append(F(i))
```

The first output word was be `0x25d86fb7c42fd70a7e800e871f22f2f03a282abb18f86c347a1078a92f713f60` which is equal to `17117985411748610629288516079940078114952304104811071254131751175361957805920`.


```
Input:
['0x0000000000000000000000000000000000000000000000000000000000000001', '0x0000000000000000000000000000000000000000000000000000000000000002', '0x0000000000000000000000000000000000000000000000000000000000000000']
Output:
['0x25d86fb7c42fd70a7e800e871f22f2f03a282abb18f86c347a1078a92f713f60', '0x129e7fc58c40a11b183cc4bc8b0891def35dcd162ef30186a0a6f3b73943b613', '0x0908875557d4dbcd02f58f3e181719aed2c220ce3bbdcc936acf925daf1e5fa3']
```

crypto/ts/index.ts
weijiekoh: Will do!

weijiekoh:
FYI I'm still working on this. I encountered a few blockers:

1. Now that the circuit has more constraints (due to the patched Poseidon circuit), CircleCI doesn't like the circuit build process and kills it. I might have to migrate to Travis CI.
2. I'm upgrading `snarkjs`, `circom`, and `circomlib` as only the latest `circomlib` implements the patched Poseidon function (in JS).

weijiekoh:
CircleCI no longer likes us because we run really intense CPU jobs when we do proof generation. Fortunately, Travis works. The CI checks pass in my own account (`weijiekoh`) and I'm in the process of enabling it under `appliedzkp`.

weijiekoh:
**Summary of changes**

- All submodules now have a new semver: 0.2.0. @xuhcc, please note that you'll have to update clrfund's package.json files to point to `^0.2.0` for the updated modules. There is a chance that things break, so please let me know if there are any issues and I will fix them ASAP.
- Updated Poseidon hash functions
- We use the latest version of `snarkjs`, which is currently incompatible with `zkutil`. As such, I've replaced `zkutil` with `snarkjs` operations in `buildSnarks.ts`. The downside is that whenever we build circuits (for development), the script has to download a 500+MB `pot19_final.ptau` file from my Dropbox account.
- We don't use CircleCI any more. Instead, we now use Travis.

weijiekoh:
I also made changes to the circuits to make them work with the latest version of `circom`. Specifically, signals can no longer be assigned using `=`, and "lazy" circuit evaluation no longer works. See: https://github.com/weijiekoh/circom_test

See Jordi's response to the lazy circuit evaluation problem I posted in the iden3 TG chat:

```
Ok @weijiek  I understand your problem.

Various things.

1.- When you run calculateWitness(input) in your code I recomend you to add a second parameter to true: 
calculaWitness(input, true) This second parameter is "sanityCheck" parameter and it's very convenient 
specially when you are debugging/testing.  This will give you better info in what's going on.

I added a [PR](https://github.com/weijiekoh/circom_test/compare/master...jbaylina:patch-1) to your code to add this. 

2.- In your case, when you activete this flag, you will get this error:

`
ERROR:  5 Accessing a not assigned signal main.selectors[0].out[0]
[ERROR] snarkJS: Error: Accessing a not assigned signal main.selectors[0].out[0]
    at error (/Users/jbaylina/git/iden3/circom_runtime/js/witness_calculator.js:51:23)
    at <anonymous>:wasm-function[98]:0x642f
    at <anonymous>:wasm-function[109]:0x6921
    at <anonymous>:wasm-function[88]:0x6229
    at init (<anonymous>:wasm-function[89]:0x62ea)
    at WitnessCalculator._doCalculateWitness (/Users/jbaylina/git/iden3/circom_runtime/js/witness_calculator.js:130:31)
    at WitnessCalculator.calculateWitness (/Users/jbaylina/git/iden3/circom_runtime/js/witness_calculator.js:158:20)
    at Module.wtnsDebug (file:///Users/jbaylina/git/iden3/snarkjs/src/wtns_debug.js:44:24)
    at async Object.wtnsDebug [as action] (file:///Users/jbaylina/git/iden3/snarkjs/cli.js:372:5)
    at async clProcessor (file:///Users/jbaylina/git/iden3/snarkjs/src/clprocessor.js:50:21)
`

The error is clear now.  You are reading a signal that's not assigned yet.

In alder versions of circom I allowd it. But we dissabled it because this adds a lot of complexity to the code generated and also for debugging circuits.

The only condition that you need to now, is that components are calculated when all the input signals are set. More concretelly, when the last input signal is set.

3.- I pushed a new version of circom (v0.5.25) and snarkjs (v0.3.25) to add better information of exactly which signal are you trying to read before being assigned.
```

weijiekoh:
> I got the SageMath reference implementation to output the same hardcoded test values from `circomlib`.

2021 update: https://github.com/appliedzkp/poseidon_in_circomlib_check

PR #164, opened 2020-09-06
Remove a redundant LeafExists check in updateStateTree.circom

### LINKED ISSUE #154: Redundant Merkle check in updateStateTree.circom
It isn't necessary to compute the new state root for each batch of messages as the circuit already does this check when it checks if the new state leaf exists in the tree.

https://github.com/appliedzkp/maci/blob/704247a22f51fe4fc3eba81aea40f28b6b6f8fed/circuits/circom/updateStateTree.circom#L419-L426
https://github.com/appliedzkp/maci/blob/704247a22f51fe4fc3eba81aea40f28b6b6f8fed/circuits/circom/trees/incrementalMerkleTree.circom#L61-L80

weijiekoh:
In `updateStateTree.circom`, there is a `LeafExists` check which ensures that `new_state_tree.root` is correct. This check, however, is unnecessary, as the value of `new_state_tree.root` is generated using the same inputs (`state_tree_path_elements`) and `state_tree_path_index`) which are used to generate `new_state_tree.root` in the first place. This PR just removes the redundant `LeafExists` to save on constraints. Fixes #154.


weijiekoh:
Seems that even though some Travis CI tests fail, there are some green checkmarks. Will figure it out.

PR #168, opened 2020-09-17
Upgraded circom to v0.5.24 and rebuilt snarks

weijiekoh:


xuhcc:
Message processing slowed down ~5x times for (8, 11, 3) after upgrade to 0.2.7.

(It's not a problem. Just sharing my test results :)

weijiekoh:
Thanks for the info! I will add https://github.com/appliedzkp/maci/issues/169 to my task list. Going back to `zkutil` will speed up proof generation.

PR #173, opened 2020-09-22
Migrate to node.js 12

### LINKED ISSUE #172: Migrate to node.js 12
According to the readme, MACI requires node.js 11:

```
You should have Node 11.14.0 installed. Use nvm to install it.
```

But this version is now unsupported and the increasing number of npm packages are starting to require `=10 || >=12`.

xuhcc:
~I ran the tests on node.js 12 and all of them passed.~

Resolves #172

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 11
+  - 12
 
 install:
   - npm i -g npm
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ We welcome contributions to this project. Please join our
 
 ### Requirements
 
-You should have Node 11.14.0 installed. Use
+You should have Node 12 installed. Use
 [`nvm`](https://github.com/nvm-sh/nvm) to install it.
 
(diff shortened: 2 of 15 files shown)


xuhcc:
CI build fails.

The cause of failure is outdated `scrypt` package which is a dependency of etherlime (https://github.com/LimeChain/etherlime/issues/314). It looks like etherlime is only used in `maci-contracts` package and probably can be replaced.

xuhcc:
I've removed etherlime from tests but it is still used in `contracts/scripts/compileSol.sh` to compile contracts.

weijiekoh:
Thanks for this PR!

May I suggest the solc binary instead of etherlime?

Here's an attempt at doing so: https://github.com/appliedzkp/maci/pull/174

weijiekoh:
Just found out that switching to the `solc` binary will take more work the deploy script will have to read the ABI and .bin files via `fs.readFileSync`. Previously, Etherlime took care of that.

weijiekoh:
Continued in #174.

PR #174, opened 2020-09-23
Upgrade to Node 12 and remove dependence on Etherlime

weijiekoh:
This PR follows up from @xuhcc's PR: https://github.com/appliedzkp/maci/pull/173

The main change is to upgrade to Node 12.

Since Etherlime requires Node 11 due to an upstream `scrypt` dependency, this PR also removes MACI's dependence on Etherlime. Now, the compile and deploy scripts use the native `solc` binary.


contracts/sol/IncrementalMerkleTree.sol
```diff
@@ -23,7 +23,7 @@ pragma solidity ^0.5.0;
 
 import { SnarkConstants } from "./SnarkConstants.sol";
 import { Hasher } from "./Hasher.sol";
-import { Ownable } from "@openzeppelin/contracts/ownership/Ownable.sol";
+import { Ownable } from "./openzeppelin-solidity//contracts/ownership/Ownable.sol";
```
xuhcc, line 26: Can we import these contracts from node_modules instead? That would be easier to maintain.

```
import { Ownable } from "../node_modules/openzeppelin-solidity/contractscontracts/ownership/Ownable.sol"
```

contracts/scripts/compileSol.sh
```diff
@@ -8,16 +8,18 @@ echo 'Building contracts'
...
-    # You need Solidity 0.5.16 (https://github.com/ethereum/solidity/releases/tag/v0.5.16) installed in your PATH
-    SOLC_VERSION=native
+if [[ -z "${SOLC_PATH}" ]]; then
+    # Assumes that you have solc 0.5.17 (https://github.com/ethereum/solidity/releases/tag/v0.5.17) installed in your PATH
+    solcBin="solc"
 else
-    SOLC_VERSION=0.5.16
+    # Otherwise, you can specify the path to solc 0.5.17
+    solcBin="${SOLC_PATH}"
 fi
```
xuhcc, line 17: This will require developer to install solc manually. Can we automate this by downloading the solc binary like we do in CI environment?

Another option would be using solc-js but it requires additional work because solc-js has different interface (I tried it yesterday but didn't make much progress).

contracts/sol/IncrementalMerkleTree.sol
weijiekoh: Yes! I hadn't figured out how to do that but I came across a fix and will push it.

contracts/scripts/compileSol.sh
weijiekoh: > Can we automate this by downloading the solc binary like we do in CI environment?

Great idea! Will do.

PR #176, opened 2020-09-24
Fix OZ import paths in solidity files

### LINKED ISSUE #172: Migrate to node.js 12

xuhcc:
This prevents import errors when yarn package manager is used:

```
File maci-contracts/node_modules/openzeppelin-solidity/contracts/ownership/Ownable.sol, imported from maci-contracts/sol/IncrementalQuinTree.sol, not found.
```

I also switched to `@openzeppelin/contracts` package as recommended at https://github.com/OpenZeppelin/openzeppelin-contracts

Resolves #172

### DIFF
--- a/contracts/package-lock.json
+++ b/contracts/package-lock.json
@@ -619,6 +619,11 @@
 				"@types/yargs": "^13.0.0"
 			}
 		},
+		"@openzeppelin/contracts": {
+			"version": "2.5.1",
+			"resolved": "https://registry.npmjs.org/@openzeppelin/contracts/-/contracts-2.5.1.tgz",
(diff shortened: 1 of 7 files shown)


weijiekoh:
Thank you!!

xuhcc:
@weijiekoh Could you publish a new version to NPM?

weijiekoh:
Sorry for the delay! Just published 0.2.9.

Issue #170, opened 2020-09-21
[Feature Request] sign up and vote


barryWhiteHat:
## Is your feature request related to a problem? Please describe.

At the moment when users sign up they have to wait until [`isAfterSignUpDeadline`](https://github.com/appliedzkp/maci/blob/master/contracts/sol/MACI.sol#L318) before they can submit their commands. 

This is bad UX where users have to take action, wait until the sign up period is over and then return in order to vote. 

The reason it is like this is to prevent users from being bribed as follows 

1. Sign up
2. Make bribe command
3. Command to set private key to 0x0. 

This way a user at sign up time can prove how they voted. To prevent this we prevent users from doing step 2 until after the sign up period. 

## Describe the solution you'd like

I would like messages that are published during the sign up period to be inserted into the command que to be processed after all other commands. So if a user does step 1 and 2 and 3 at sign up time. Step 1 is executed immediately but step 2 and 3 are deferred until the end of the voting period. Giving the bribee time to return later and change their vote. 

The change should be quite simple. 
Replace
```
        messageTree.insertLeaf(leaf);
```
With 

```
     if signupperiod:
        finalmessageTree.insertLeaf(leaf)
    else:
        messageTree.insertLeaf(leaf);
```

weijiekoh:
Could we do it this way instead?

```
function signUp(PubKey _pubKey, Message _msg) public {
    StateLeaf leaf = hash(_pubKey, voiceCredits, ...);
    stateTree.insertLeaf(leaf);

    finalMessageTree.insertLeaf(hash(_msg));
}

function publishMessage(Message _msg) public {
    penultimateMessageTree.insertLeaf(hash(_msg));
}
```

And during message processing, process messages in `finalMessageTree` before processing those in `penultimateMessageTree`?

weijiekoh:
@xuhcc @barryWhiteHat please let me know your thoughts on which approach we should take.

I think the latter approach is better UX as there won't be a signup period any more.

xuhcc:
@weijiekoh In your approach a bribee is able to send overriding commands immediately after signup, right?

If that's the case, I agree that this is a preferred solution.

barryWhiteHat:
> And during message processing, process messages in penultimateMessageTree before processing those in finalMessageTree?

[...]

PR #180, opened 2020-09-29
Signup-and-vote, process messages in reverse order, and enable zkutils

weijiekoh:
Addresses #170. This PR removes the signup period. Now anyone can sign up and publish messages before the voting period ends. To prevent an attack where a briber makes someone sign up and change their key immediately, this PR also reverses the order of message processing.

It also includes #181, which uses zkutil instead of snarkjs for the trusted setup (for testing only) and proof generation.


contracts/sol/MACI.sol
```diff
@@ -263,7 +263,9 @@ contract MACI is DomainObjs, ComputeRoot, MACIParameters, VerifyTally {
     function signUp(
         PubKey memory _userPubKey,
         bytes memory _signUpGatekeeperData,
-        bytes memory _initialVoiceCreditProxyData
+        bytes memory _initialVoiceCreditProxyData,
+        Message memory _message,
+        PubKey memory _encPubKey
```
xuhcc, line 268: The user may want to vote for multiple options, so if we allow them to publish a message using `signUp` method, it would make sense to accept an array of messages here.

On the other hand, it is not necessary for `signUp` to accept messages at all, because an external contract can be created which allows to do `signUp` and make multiple `publishMessage` calls in one transaction (this is the approach I would prefer).

weijiekoh: > On the other hand, it is not necessary for signUp to accept messages at all, because an external contract can be created which allows to do signUp and make multiple publishMessage calls in one transaction (this is the approach I would prefer).

Perfect, this will make it easier to implement! I will revert this commit.

weijiekoh:
Includes PR #181 which enables zkutil, so there is no longer a need to download the large ptau file.

contracts/sol/MACI.sol
```diff
@@ -315,14 +317,11 @@ contract MACI is DomainObjs, ComputeRoot, MACIParameters, VerifyTally {
         Message memory _message,
         PubKey memory _encPubKey
     ) 
-    isAfterSignUpDeadline
```
xuhcc, line 318: There are still distinct signup and voting periods, but now users can also vote during signup period?

weijiekoh: Yup, they can.

contracts/sol/MACI.sol
```diff
@@ -417,7 +418,7 @@ contract MACI is DomainObjs, ComputeRoot, MACIParameters, VerifyTally {
     }
 
     function hasUnprocessedMessages() public view returns (bool) {
-        return currentMessageBatchIndex < numMessages;
+        return allMessagesProcessed == false;
```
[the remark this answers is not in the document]
weijiekoh: Will do!

circuits/scripts/installZkutil.sh
```diff
@@ -15,4 +15,4 @@ chmod a+x rust.sh
 sh rust.sh -y
 rm rust.sh
 source $HOME/.cargo/env
-cargo install zkutil --version 0.3.1
+cargo install zkutil --version 0.3.2
```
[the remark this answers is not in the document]
weijiekoh: Sure thing :+1:

README.md
```diff
@@ -27,7 +27,24 @@ npm run build
...
+
+Install Rust:
+
+```bash
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+```
+
+Also install [`zkutil`](https://github.com/poma/zkutil) v0.2.1 and ensure that
+the `zkutil` binary is in the `~/.cargo/bin/` directory. You can configure the
+path to this binary via `maci-config` (see `config/test.yaml` for an example).
```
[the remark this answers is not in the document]
xuhcc: btw, here's how I usually set the path to binary:

```
export NODE_CONFIG='{"zkutil_bin": "/usr/bin/zkutil"}'
```

.travis.yml
```diff
@@ -7,14 +7,15 @@ install:
...
-  - ./contracts/scripts/compileSol.sh
   - cd crypto && npm run test
   - cd ../domainobjs && npm run test
   - cd ../core && npm run test
   - cd ../circuits && npm run buildBatchUpdateStateTreeSnark && npm run buildQuadVoteTallySnark
   # - ./scripts/buildSnarksSmall.sh
   - ./scripts/runTestsInCircleCi.sh
-  - cd ../contracts && ./scripts/runTestsInCircleCi.sh
+  - cd ../contracts && ./scripts/compileSol.sh
+  - ./scripts/runTestsInCircleCi.sh
```
xuhcc, line 20: I tried to run contracts tests locally, but they are failing. One of the errors I see is this:

```
ENOENT: no such file or directory, open '.../maci/circuits/build/1601562991582.publicSignals.json'
```

weijiekoh: I see - please try the following:

```
cd circuits
./scripts/buildBatchUpdateStateTreeSnark.sh && ./scripts/buildQuadVoteTallySnark.sh
```

xuhcc: Ah I forgot to set zkutil binary path :sweat_smile: 

Created PR to better handle such cases -> #182

PR #181, opened 2020-09-30
Enable zkutil again

weijiekoh:
`circom` now works with `zkutil`. This PR will reduce the trusted setup time for testing purposes and will remove the need for a large .ptau file.


weijiekoh:
`zkutil` is only slightly faster than the `snarkjs` prover (260 vs 277s).

Nevertheless, not having to generate / download the ptau file is good for developer experience.

weijiekoh:
Merged into #180.

PR #188, opened 2020-10-17
Branch that uses Solidity 0.7.3

weijiekoh:


contracts/sol/IncrementalMerkleTree.sol
```diff
@@ -19,11 +19,14 @@
...
  */
 
-pragma solidity ^0.5.0;
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.7.3;
+
+import "./SnarkConstants.sol";
+import "./Hasher.sol";
+import "../node_modules/@openzeppelin/contracts/access/Ownable.sol";
```
xuhcc, line 28: Imports like this may not work for projects which use `maci-contracts` package because `@openzeppelin` can be in different location.

PR #189, opened 2020-10-23
Crypto: fix `hash5`

mhchia:
In `hash5`, `elements` is passed to `poseidonT6` instead of `elementsPadded`, which is the one ought to be used.


weijiekoh:
Thanks @mhchia for spotting this bug!!

PR #193, opened 2020-11-17
Remove verifiers for 'small' circuits

xuhcc:
The users are supposed to generate their own verifiers so we can remove these files from the repo.


weijiekoh:
Thanks @xuhcc !

PR #203, opened 2021-02-04
Add snarkParamsPath config parameter

xuhcc:
- Added config parameter `snarkParamsPath` which should resolve #197.
- Changed default directory where snark build artifacts are collected to `circuits/params`, so now they are stored separately from JS files (`circuits/build`).

PR #204, opened 2021-02-05
Fix resolution of relative snark params path

xuhcc:
I get this error with `maci-circuits 0.5.6`:

```
Error: ENOENT: no such file or directory, open `./project-dir/circuits/params/1612548070018.input.json'
```

The packaged script treats `snarkParamsPath` as relative to current working directory but it must treat it as relative to script itself. This patch should fix it.

### DIFF
--- a/circuits/ts/index.ts
+++ b/circuits/ts/index.ts
@@ -14,7 +14,7 @@ import { config } from 'maci-config'
 const zkutilPath = config.zkutil_bin
 const snarkParamsPath = path.isAbsolute(config.snarkParamsPath)
     ? config.snarkParamsPath
-    : path.resolve(config.snarkParamsPath)
+    : path.resolve(__dirname, config.snarkParamsPath)
 
 /*
  * @param circuitPath The subpath to the circuit file (e.g.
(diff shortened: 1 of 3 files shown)


xuhcc:
Another problem with 0.5.6 is that snark parameters are stored in `maci-circuits/build`, but they should be moved to `maci-circuits/params` because the default path was changed in #203.

weijiekoh:
Thanks for the PR! I've published v0.5.7 to NPM.

xuhcc:
@weijiekoh 
0.5.7 still stores parameters for 'test' circuits in `build` directory

xuhcc:
```
$ ls node_modules/maci-circuits/build/
batchUstCircuit.r1cs  batchUst.params  batchUstPk.json  batchUst.sym  batchUstVk.json  batchUst.wasm  buildSnarks.js  buildSnarks.js.map  genVerifier.js  genVerifier.js.map  index.js  index.js.map  qvtCircuit.r1cs  qvt.params  qvtPk.bin  qvt.sym  qvtVk.json  qvt.wasm
$ ls node_modules/maci-circuits/params/
ls: cannot access 'node_modules/maci-circuits/params/': No such file or directory
```

xuhcc:
If you'd like, you can just remove these files (260 Mbytes), because now we have a setting for snark params path.
Though having them packaged could be convenient for beginners.

PR #209, opened 2021-02-22
Replace Ownable with coordinatorAddress for coordinatorReset

weijiekoh:
This makes it easier for the clrfund contracts to work with the MACI contract.


weijiekoh:
@xuhcc I've published 0.5.10 :smile:

PR #213, opened 2021-03-03
Use hardhat to compile contracts

corydickson:
Right now the Poseidon build scripts can work for the existing structure of hardhat projects, but in order to avoid brittleness we may want to deprecate the use of Artifactor as if the language changes around directories it will require a maintenance to build scripts.

Exploring the use of https://hardhat.org/plugins/hardhat-abi-exporter.html in the future may be beneficial


corydickson:
Let me know if it'd be better not to track changes in the artifacts folder

PR #225, opened 2021-03-17
MACI 0.6.0

weijiekoh:
This is a new semver as it will cause some breaking changes.

The circuits have been upgraded to prevent a DoS vector. Commands with fields above `2 ^ 32` could cause proofs to not be able to be generated. The fix was to use 252-bit circomlib range proof checks.

The `download`, `replay`, and `replayCsv` subcommands help with stress-testing especially on live networks like xDAI.

There is a new debug mode which disables the signup and voting periods. This makes it unnecessary to wait for the periods to pass before processing messages.


contracts/sol/MACI.sol
```diff
@@ -71,7 +71,11 @@ contract MACI is DomainObjs, ComputeRoot, MACIParameters, VerifyTally {
     // immediately upon deployment.
     uint256 public signUpTimestamp;
 
-    // Duration of the sign-up and voting periods, in seconds
+    // Duration of the sign-up and voting periods, in seconds. If these values
+    // are set to 0, the contract will be in debug mode - that is, only the
+    // coordinator may sign up and publish messages. This makes it possible to
+    // submit a large number of signups and messages without having to do so
+    // before the signup and voting deadlines.
```
xuhcc, line 78: Is debug mode supposed to be used in production?

contracts/sol/QuadVoteTallyVerifierMedium.sol
```diff
@@ -165,21 +165,21 @@ contract QuadVoteTallyVerifierMedium {
...
+        vk.IC[6] = Pairing.G1Point(uint256(1265862285389912583354988517188240046545245531899484378775932237585919113405),uint256(19533887002306214906401588301741821164134268055785360059716974096750249574592));
+        vk.IC[7] = Pairing.G1Point(uint256(9366383303037262111495176025900763714531644320956547136527308721507028908668),uint256(12342056557140844024941014535132352840593653924946843771623130612151334486161));
+        vk.IC[8] = Pairing.G1Point(uint256(377648027856046402067856432085235248427633732903195773131155890402237030240),uint256(13859189692549416893530466171559207483847454708965810295259420276885186194111));
+        vk.IC[9] = Pairing.G1Point(uint256(12137642909261162531614282794577276988027441374171514839983505649086044434171),uint256(18386481830849357503396854902661055139744772247970949541726767736810048014090));
+        vk.IC[10] = Pairing.G1Point(uint256(11168589902532857753647687088832653482107030818840713069138157464421493318222),uint256(14379320319015633530046109708947298315994836438043323861191851455013758947415));
```
xuhcc, line 182: I think verifiers can be removed & gitignored because the maci-circuits package does not contain corresponding snark parameters.

cli/0x2C2B9C9a4a25e24B174f26114e8926a9f2128FE4.json
```diff
@@ -0,0 +1,106 @@
...
+        "17963171418159573443212685830696624129145451976211369253153464087551051564356",
+        "18070900453130066079508991220033940484915745849682676584495920689808056185273",
+        "15281542088681398561172262336380123657983489309022367846766670534480332247250",
+        "15658976556044653017639073481670465795456937976326439654217215964137770110765",
+        "16804117294521997319830729223093417549221914353374389420572241640856758290235",
+        "12273892878991597886379704231168287877218070804087754512256378117154414171928"
+      ]
+    }
+  ]
+}
```
xuhcc, line 106: It seems that this file was added inadvertently.

contracts/sol/MACI.sol
weijiekoh: No, since in production the signup and voting periods will be greater than 0, debug mode will be off.

contracts/sol/QuadVoteTallyVerifierMedium.sol
weijiekoh: Absolutely! I'll remove them and  update .gitignore.

cli/0x2C2B9C9a4a25e24B174f26114e8926a9f2128FE4.json
weijiekoh: Oops! I'll remove it.

contracts/sol/MACI.sol
xuhcc: If it is meant to be used solely during development, could it be implemented in a separate contract (e.g. `MACIDebug.sol`) to prevent MACI contract from increasing in size?

xuhcc: (The MACI contract itself is not that big yet, but in clrfund there's a MACI factory contract which size is close to EIP170 limit.)

PR #230, opened 2021-03-18
Enable programmatic access to getProofs and proveOnChain commands

xuhcc:
...and other minor fixes & improvements


weijiekoh:
Awesome, thank you!

PR #233, opened 2021-03-25
Use ethers contract factory and update tests

corydickson:
AccQueue tests are failing due to some changes in the `crypto` lib that I didn't pick up until rebasing with the latest changes but otherwise this should lay the groundwork to correctly deploy the test suite


contracts/ts/deploy.ts
```diff
@@ -342,18 +337,28 @@ const deployMaci = async (
...
 
     // VkRegistry
-    const vkRegistryContract = await deployVkRegistry(deployer)
+    const vkRegistryContract = await deployVkRegistry()
     await vkRegistryContract.deployTransaction.wait()
 
+    const signer = await getDefaultSigner()
     log('Transferring VkRegistry ownership to MACI', quiet)
-    await (await (vkRegistryContract.transferOwnership(maciContract.address))).wait()
+    await (await (vkRegistryContract.transferOwnership(signer.address))).wait()
```
corydickson, line 361: So I think this might have been a mistake, but since we are deploying all contracts with a single signer, when the initialization function for MACI gets called it does a check that the owner of the VkRegistry is equal to the owner of the MACI instance (the default signer). 

Either this logic is correct or this check in the contract should be updated such that `VkRegistry.owned() === address(this)`. Though this has the side effect of when the test suite calls certain methods to register a verifying key, `ether.js` uses the signer by default to execute the transactions and does not support overriding the `from` field. This implies that we would need to deploy the contract from `MACI.sol` thus not using a signed contract factory.

weijiekoh: Thank you! This line of code was left over from a previous iteration where I made the MACI contract set verifying keys in the VkRegistry.

I've removed it. The owner of VkRegistry is automatically the deployer (aka the `signer`).

PR #234, opened 2021-03-25
Large circuits (requires g++)

weijiekoh:
Supports 4095 users, 32768 messages, and 125 vote options.

Requires the coordinator to install these dependencies:

```
sudo apt-get install -y libgmp-dev nlohmann-json3-dev nasm g++ build-essential
```


xuhcc:
>sudo apt-get install -y libgmp-dev nlohmann-json3-dev nasm g++ build-essential

What would you recommend to Windows and macOS users?

What are the system requirements for coordinator? I'm especially interested in memory requirements.

weijiekoh:
I'm not sure about Windows or macOS. I think it'll be difficult to support them.

I just pushed a commit containing a Dockerfile and a docker-compose script. They should help anyone who needs to run MACI if they aren't on Debian/Ubuntu.

PR #236, opened 2021-04-01
Use Poseidon instead of SHA256 to hash messages + fix the UnpackElement circuit

weijiekoh:
1. I initially thought that it was a good tradeoff to use SHA256 to hash messages so users can save gas. I came to realise that this would make it difficult for projects like clrfund to increase the message batch size for the processing circuit, as the circuit would require a lot more memory to compile.

2. The UnpackElement circuit used the right-shift `<<` operator to unpack an input value into multiple 50-bit outputs. I used a constraint to check that the outputs were correct but closer inspection by @han0110 revealed that there was actually no constraint created in the compiled circuit. As such, I've modified this circuit to use `Num2Bits_strict` to convert the input to bits, and then `n` `Bits2Num(50)` circuits to convert each 50-bit chunk to the desired output. The new circuit requires only 500+ constraints, which is well within acceptable limits.

### DIFF
--- a/circuits/circom/hasherPoseidon.circom
+++ b/circuits/circom/hasherPoseidon.circom
@@ -45,20 +45,16 @@ template Hasher5() {
     hash <== hasher.out;
 }
 
-template Hasher11() {
-    // Hasher2_2(
-    //     Hasher2_1 (
-    //         Hasher5_1(in[0], in[1], in[2], in[3], in[4]),
-    //         Hasher5_2(in[5], in[6], in[7], in[8], in[9])
-    //     ),
-    //     in[10]
+template Hasher10() {
+    // Hasher2(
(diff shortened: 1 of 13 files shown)


circuits/circom/hasherPoseidon.circom
```diff
@@ -45,20 +45,16 @@ template Hasher5() {
...
-    //         Hasher5_2(in[5], in[6], in[7], in[8], in[9])
-    //     ),
-    //     in[10]
+template Hasher10() {
+    // Hasher2(
+    //     Hasher5_1(in[0], in[1], in[2], in[3], in[4]),
+    //     Hasher5_2(in[5], in[6], in[7], in[8], in[9])
     // )
 
-    signal input in[11];
```
corydickson, line 57: was this last element of the array ever used?

PR #243, opened 2021-05-10
Add process message check in local state

corydickson:
just a small missing commit


weijiekoh:
Thanks @corydickson !

PR #245, opened 2021-06-02
Generate users for happy path testing

corydickson:
Going to split up the PRs for the happy path and the bribery test cases


weijiekoh (approved):
Happy path works!

PR #251, opened 2021-06-20
Poseidon encryption + fixes to vote tallying in MaciState.ts

weijiekoh:
TODO for @weijie: prevent the DOS vector where a malicious message prevents a proof from being generated


circuits/circom/messageToCommand.circom
```diff
@@ -1,14 +1,14 @@
 include "./ecdh.circom";
-include "./decrypt.circom";
 include "./unpackElement.circom";
 include "../node_modules/circomlib/circuits/bitify.circom";
+include "../node_modules/circomlib/circuits/poseidon.circom";
 
 template MessageToCommand() {
-    var MSG_LENGTH = 8;
+    var MSG_LENGTH = 7;
```
corydickson, line 7: why is the message length only seven? isn't suppose to be 9

weijiekoh: With MiMC7 encryption, the initialisation vector was part of the message. Poseidon encryption doesn't have an IV any more though.

PR #252, opened 2021-06-22
Allow to publish keychange message

corydickson:
Supports key change messages in test DSL


integrationTests/ts/__tests__/suites/suite3_keychange.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Happy path for user key change",
+    "description": "Single user vote with new keys",
+    "numUsers": 1,
+    "numVotesPerUser": 1,
```
corydickson, line 5: @weijiekoh I'm going to refactor how we publish messages in the next PR but this should allow us to start writing a few tests. 

One of the main problems is that `numVotesPerUser` is a bit misleading, since technically if a user changes keys they can vote for another option with a different weight.

weijiekoh:
Thanks for the PR! Running Travis CI tests in a private repo at the moment. (I think we are out of credits in this public repo)

PR #263, opened 2021-08-12
Fix .sol compilation to work in directories which are symlinks

aspiers:
If a `.sol` file being compiled lives in a directory path which contains a symlink component, then the `--allowed-paths` option of `solc` only works when the symlinks are fully resolved.  So use `pwd -P` to obtain the absolute paths instead of just `pwd`.


weijiekoh:
Thanks for this improvement!

PR #264, opened 2021-08-12
Clarify how to run tests

aspiers:
The existing instructions did not work, because they instructed the user to start their own Ganache instance, which would then conflict with the ones started by the scripts.  So clarify this, and also explain how to run individual tests one by one, which is what the `README` already recommended anyway to prevent incorrect nonce errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,14 +116,29 @@ cd contracts
 npm run ganache
 ```
 
-In another terminal, run the tests individually:
+In another terminal, run any of the tests found in `contracts/ts/__tests__/`
+via pattern matching, e.g.:
+
+```bash
+cd contracts
+npx jest IncrementalMerkleTree
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,14 +116,29 @@ cd contracts
...
+npx jest IncrementalMerkleTree
+```
+
+would run `IncrementalMerkleTree.test.ts`.
+
+N.B. `npx jest Tree` would run that and `IncrementalQuinTree.test.ts`
+in parallel, causing incorrect nonce errors.
+
+Alternatively you can run all unit tests as follows, but you should
+stop your Ganache instance first as this will start its own instance
```
weijiekoh, line 133:
```suggestion
stop your Hardhat instance first as this will start its own instance
```

PR #265, opened 2021-08-12
Change signups check to new calculation

### LINKED ISSUE #260: Properly enforce upper bound of sign ups
Check if that the number of sign ups does not exceed `STATE_TREE_ARITY ** stateTreeDepth` in MACI.sol

corydickson:
Closes #260 

I'd include a testcase and revert reason, but it takes far too long to create so many users given our current implementation.


weijiekoh (approved):
Thanks for the fix!

PR #267, opened 2021-08-13
Update readme files for some modules

weijiekoh:
Addresses #226

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -10,303 +10,89 @@ transform said data. This makes it easier to reason about the system, write
 tests, and implement functionality. It also allows us to implement the smart
 contracts in discrete components which are easy to test.
 
-To this end, we this submodule exposes a `MaciState` class and a `User` class.
-
-## **`Poll`**
-
-Each `Poll object has the following attributes:
-
-`coordinatorKeypair: Keypair`: The coordinator's keypair.
-`messages: Message[]`: An array of all published messages.
-`ballots: Ballot[]`: An array of all ballots.
-
-TODO
-
-### Functions
-
-#### **`publishMessage`**
-
-Function signature:
-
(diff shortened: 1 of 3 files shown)


crypto/README.md
```diff
@@ -49,3 +84,31 @@ Produces a signature of the given message using the private key.
...
+**`hash3 = (elements: Plaintext): BigInt`**: the Poseidon hash function for 3
+inputs.
+
+**`hash4 = (elements: Plaintext): BigInt`**: the Poseidon hash function for 4
+inputs.
+
+**`hash5 = (elements: Plaintext): BigInt`**: the Poseidon hash function for 5
+inputs.
+
+**`hash12 = (elements: Plaintext): BigInt`**: the Poseidon hash function for 12
```
corydickson, line 112: Does this implicitly use hashLeftRight by first using hash5 then hash6 on the remaining inputs? If so we may want to include this here so users are aware of the ordering.

corydickson (changes requested):
Looking good, just a couple small things

core/README.md
```diff
@@ -10,303 +10,89 @@ transform said data. This makes it easier to reason about the system, write
...
-#### **`copy`**
-
-Function signature:
-
-```ts
-(): Poll
-```
-
-Deep-copies and returns this object.
+To this end, we this submodule exposes a `MaciState` class and a `Poll` class.
```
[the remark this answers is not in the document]
weijiekoh: Fixed

crypto/README.md
weijiekoh: Thanks for the suggestion. I've added the pseudocode in a new commit.

corydickson (approved):

PR #271, opened 2021-08-16
CircleCI config for v1

corydickson:


weijiekoh:
We can actually delete `.circleci/` entirely as we don't use it.

PR #275, opened 2021-08-18
Fix logic for offchain batch calculations

corydickson:


weijiekoh (approved):
Travis tests pass now!

PR #278, opened 2021-09-02
Contract initialize tally

corydickson:
Just ignoring the fact that we would want to actually provide the merkle roots of spentVO and credits, this example is to show that the would need to be done causes the contract not to deploy as we've reached the size limit. The biggest culprit is the `Hasher` interface

@weijiekoh looks like we will just ensure the check in the circuit but not sure if the other avenue is available to us anyway

### DIFF
--- a/circuits/circom/tallyVotes.circom
+++ b/circuits/circom/tallyVotes.circom
@@ -132,12 +132,15 @@ template TallyVotes(
         voteTree[i].root === ballots[i][BALLOT_VO_ROOT_IDX];
     }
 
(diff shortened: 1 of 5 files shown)


weijiekoh:
Hi! Just to clarify, if we set the initial tally commitment to some value computed in the contract (using Poseidon via `Hasher.sol`), we will run into contract size problems?

Therefore, a better way is to not to make the change at the contract level, but instead account for this case in the circuit?

corydickson:
Exactly, just posting this an example

PR #287, opened 2021-09-24
Add dockerfile as multibuild job

corydickson:
Run ```docker build -t maci . ```

To provide some context on why we want this structure for each of the submodules, when we implement github actions for different parts of the code base we can stop the build process to specific tagged intermediate images see https://docs.docker.com/develop/develop-images/multistage-build/

Essentially this will allow us to copy over cached artifacts to different images that need to be rebuilt to test new code changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:15.8.0-buster AS builder
+FROM node:16-bullseye AS builder
 
 WORKDIR /maci
 COPY . /maci/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM node:16-bullseye AS builder
```
corydickson, line 1: Had to bump the tag version to a supported Node docker image

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+WORKDIR /maci
+COPY . /maci/
+
+RUN npm i && \
+    npm run bootstrap && \
+    npm run build
```
corydickson, line 8: Build everything for now but eventually we'd build each package separately copying over the lerna generated artifacts and running `npx lerna link`

weijiekoh (approved):

PR #294, opened 2021-10-12
update Dockerfile

johnson86tw:


corydickson:
@chnejohnson we are going to be going in a different direction with the docker image in the latest release, making things into a multi-stage job.

PR #295, opened 2021-10-12
Add korean translation README.md

dvlprsh:


cloudinertia:
I just reviewed this Korean README. LGTM.

But I have a concern.
https://github.com/appliedzkp/maci/pull/296/files

This PR would update original README so this Korean translation would be outdated.

PR #296, opened 2021-10-15
Update README

corydickson:
Using a lot of the same language of the previous readme but added some of the common updates to v1 with our new tool chain. Let me know if there is anything else you'd like to see included

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,156 @@ https://appliedzkp.github.io/maci/
 
 We welcome contributions to this project. Please join our
 [Telegram group](https://t.me/joinchat/LUgOpE7J2gstRcZqdERyvw) to discuss.
+
+## Local Development and testing
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,3 +10,156 @@ https://appliedzkp.github.io/maci/
...
+cd maci && \
+npm i && \
+npm run bootstrap && \
+npm run build
+```
+
+For development purposes, you can generate the proving and verifying keys for
+the zk-SNARK circuits, along with their Solidity verifier contracts as such.
+
+Install Rust:
```
weijiekoh, line 42: For v1, we should add instructions for rapidsnark instead of zkutil.

corydickson: Ah yes good catch, do you know of a way to preserve the contribution history as well?

PR #298, opened 2021-10-16
Upgrade circuits to circom 2

weijiekoh:
- circom circuits are now compatible with circom 2 (which compiles 10x faster)
- upgraded to the latest versions of circom-helper and zkey-manager which use circom 2

### DIFF
--- a/circuits/circom/test/ecdh_test.circom
+++ b/circuits/circom/test/ecdh_test.circom
@@ -1,3 +1,4 @@
-include "../ecdh.circom"
+pragma circom 2.0.0;
+include "../ecdh.circom";
 
-component main = Ecdh();
+component main {public [pubKey]} = Ecdh();
--- a/circuits/circomHelperConfig.json
+++ b/circuits/circomHelperConfig.json
@@ -1,8 +1,6 @@
 {
-    "circom": "./node_modules/circom/cli.js",
+    "circom": "../../.local/bin/circom",
     "snarkjs": "./node_modules/snarkjs/build/cli.cjs",
-    "circom_runtime": "./node_modules/circom_runtime",
-    "ffiasm": "./node_modules/ffiasm",
     "circuitDirs": [
         "./circom/test/"
     ]
--- a/circuits/circom/ecdh.circom
+++ b/circuits/circom/ecdh.circom
@@ -1,9 +1,10 @@
+pragma circom 2.0.0;
 include "../node_modules/circomlib/circuits/bitify.circom";
 include "../node_modules/circomlib/circuits/escalarmulany.circom";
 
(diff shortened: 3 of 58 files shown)


docs_src/src/installation.md
```diff
@@ -57,6 +64,29 @@ Install dependencies for `circom-helper` and `zkey-manager`:
...
+```json
+    "circom": "RELATIVE_PATH_TO_CIRCOM",
+    "snarkjs": "./node_modules/snarkjs/build/cli.cjs",
+    "circuitDirs": [
+        "./circom/test/"
+    ]
+}
+```
+
+Edit `cli/zkeys.config.yml` to include the relative path to the
```
corydickson, line 81: Are we no longer pulling this file from the Azure instance?

circuits/circom/test/ecdh_test.circom
```diff
@@ -1,3 +1,4 @@
-include "../ecdh.circom"
+pragma circom 2.0.0;
+include "../ecdh.circom";
 
-component main = Ecdh();
+component main {public [pubKey]} = Ecdh();
```
corydickson, line 4: Could you elaborate a little bit on this syntax?

corydickson (approved):
Just a few small questions, if we are still planning to use Azure to get the zkey files, then we should update the filenames in `integrationTests/scripts/runTestsInCi.sh`

circuits/circom/test/ecdh_test.circom
weijiekoh: Sure! In circom 2, all input signals are private by default. The only way to declare a signal as public is in its `component main ...` instantiation.

docs_src/src/installation.md
weijiekoh: Just uploaded the zkey files to a new Azure container, and pushed a new commit with an updated download script.

weijiekoh:
> Just a few small questions, if we are still planning to use Azure to get the zkey files, then we should update the filenames in `integrationTests/scripts/runTestsInCi.sh`

Done :D

circuits/circomHelperConfig.json
```diff
@@ -1,8 +1,6 @@
 {
-    "circom": "./node_modules/circom/cli.js",
+    "circom": "../../.local/bin/circom",
```
corydickson, line 2: I just noticed that `circom` package has been removed. Is there a reason we would want to have people build this locally from now on?

weijiekoh: Yes, circom v2 is a Rust binary that devs will have to build separately (https://docs.circom.io/getting-started/installation/)

PR #301, opened 2021-11-02
Add view functions to caculate state commitment

corydickson:
Since we no longer separate commitments for voice options, this generates the commitment for the state tree. There are additional view functions to get the message leaf but this causes and exceeded contract size for mainnet. 

Not sure if this is a problem since deploying on xDai/L2 would be strongly recommended

### DIFF
--- a/contracts/contracts/Poll.sol
+++ b/contracts/contracts/Poll.sol
@@ -375,7 +375,10 @@ contract Poll is
     }
 
     /*
-    * Verify the number of spent voice credits given a vote option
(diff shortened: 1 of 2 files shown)


corydickson:
@daodesigner let me know if you have any questions or want to me to address any changes

daodesigner:
> @daodesigner let me know if you have any questions or want to me to address any changes

Could we run some quick gas estimates on the functions?

daodesigner:
lgtm

PR #310, opened 2021-11-18
Use a faster Merkle tree implementation

weijiekoh:
This PR is for pre-1.0 MACI. It switches out the `IncrementalQuinTree` Typescript code for https://npmjs.com/optimisedmt, which stores all non-zero nodes in memory, rather than just the filled subtrees. As such, it is much more efficient at processing messages and generating the circuit inputs and witnesses for proofs.

There is a new subcommand: `fetchLogs`, which downloads event logs from the chain and saves it to disk for consumption by `genProofs`.

The `genProofs` subcommand now saves a serialised MaciState to disk so it can be loaded from disk during retries. This is faster than regenerating the MaciState from the logs all over again.


daodesigner:
@corydickson we should merge this into master with the circom2 upgrades

samajammin:
> This PR is for pre-1.0 MACI

Not sure what happened to this PR but seems it can be closed out given we're past 1.0 MACI...?

samajammin:
I'm going to close this out. Feel free to re-open if you still think it's useful!

PR #312, opened 2021-11-23
Multiple state AccQueues for more than one poll 

corydickson:
Need to include updates to the MACI state domain object but before I do that want to get feedback on the approach. 

Another way to fix this would be to require a given poll to be completed/state merged etc. before handling future polls.


weijiekoh:
Does this PR take a different approach from #317?

corydickson:
As discussed previously since we don't imagine a use case where multiple polls will be run concurrently, this PR should be closed and revisited with a greater rewrite around managing the signups `StateAcc`

PR #313, opened 2021-11-28
use uint16 for treeArity in PollFactory.deploy

exfinen:
In maci/contracts/contracts/Poll.sol,

since the type of `treeArity` is `uint8`,
```
        uint8 treeArity = 5;   // line 63
```
in 
```
        require(   // line 73
            _maxValues.maxMessages <= treeArity ** _treeDepths.messageTreeDepth &&
            ...
```
`treeArity ** _treeDepths.messageTreeDepth` seems to overflow when `_treeDepths.messageTreeDepth` is greater than `3`, i.e. `5^3 = 125` but `5^4 = 625` which is greater than `256`, and that makes the require statement fail.

The type of `treeArity` should be a type that can store larger values without overflow.

### DIFF
--- a/contracts/contracts/Poll.sol
+++ b/contracts/contracts/Poll.sol
@@ -60,7 +60,7 @@ contract PollFactory is Params, IPubKey, IMessage, Ownable, Hasher, PollDeployme
         IMACI _maci,
         address _pollOwner
     ) public onlyOwner returns (Poll) {
-        uint8 treeArity = 5;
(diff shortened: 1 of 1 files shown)


weijiekoh:
Thanks @exfinen! I pushed a new commit that uses `uint256` for `treeArity` and casts `messageTreeDepth` and `voteOptionTreeDepth` to uin256 instead.

PR #317, opened 2021-11-30
Fix proof generation for subsequential polls

corydickson:
Instead of allowing concurrent polls where users must signup for each poll, this PR aims to allow on chain proof verification for a multiple poll instances with the same set of users.

Testing run `cli/testMultiplePoll.sh`

### DIFF
--- a/cli/testMultiplePoll.sh
+++ b/cli/testMultiplePoll.sh
@@ -1,0 +1,87 @@
+# one signup and one valid message
+node build/index.js deployVkRegistry  && \
+node build/index.js setVerifyingKeys -s 10 -i 1 -m 2 -v 2 -b 1 \
+    -p ./zkeys/ProcessMessages_10-2-1-2.test.0.zkey \
+    -t ./zkeys/TallyVotes_10-1-2.test.0.zkey \
+    -k 0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0 && \
+node build/index.js create \
+    -r 0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0 && \
+node ./build/index.js deployPoll -x 0xf204a4Ef082f5c04bB89F7D5E6568B796096735a \
+    -pk macipk.c974f4f168b79727ac98bfd53a65ea0b4e45dc2552fe73df9f8b51ebb0930330 \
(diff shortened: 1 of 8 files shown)


cli/testMultiplePoll.sh
```diff
@@ -0,0 +1,87 @@
...
+
+node build/index.js publish \
+    -p macipk.3e7bb2d7f0a1b7e980f1b6f363d1e3b7a12b9ae354c2cd60a9cfa9fd12917391 \
+    -sk macisk.fd7aa614ec4a82716ffc219c24fd7e7b52a2b63b5afb17e81c22fe21515539c \
+    -x 0xf204a4Ef082f5c04bB89F7D5E6568B796096735a \
+    -i 1 -v 0 -w 7 -n 1 -o 1
+
+node build/index.js timeTravel -s 140 && \
+node build/index.js mergeMessages -x 0xf204a4Ef082f5c04bB89F7D5E6568B796096735a -o 1 && \
+node build/index.js mergeSignups -x 0xf204a4Ef082f5c04bB89F7D5E6568B796096735a -o 1 && \
```
corydickson, line 62: Without changing how the sbCommitments are updated, this command is still necessary even though the StateAq subtrees are merged.

weijiekoh:
Hi! Does this include https://github.com/appliedzkp/maci/pull/311 or should that be merged separately?

weijiekoh:
`PollFactory` compiles to 25Kb which exceeds the mainnet limit

weijiekoh (dismissed):

weijiekoh:
Suggestions to reduce the size of Poll.sol:

- Comment out `batchEnqueueMessage` which isn't used - size is now 24.57 Kb
- Have a separate ResultVerifier contract that implements `verifySpentVoiceCredits`, `verifyPerVOSpentVoiceCredits`, `verifyTallyResult`, and `computeMerkleRootFromPath`, instead of implementing these functions in Poll.sol.

weijiekoh (changes requested):
Looks good and the test script works on my end too. We just need to reduce the size of Poll.sol - see the previous comment on how we can do this.

Edit: let's do the size reduction in a different PR.

PR #324, opened 2021-12-08
Changed the Merkle tree impl to optimisedmt for performance speedup

weijiekoh:
This PR tweaks MACI 1.0 to use https://www.npmjs.com/package/optimisedmt, which makes the generation of circuit inputs for message processing and vote tallying faster, as tree updates are no longer O(numLeaves * treeDepth).

Addresses https://github.com/appliedzkp/maci/issues/319


circuits/ts/__tests__/TallyVotes.test.ts
```diff
@@ -192,6 +190,8 @@ describe('TallyVotes circuit', () => {
             const witness = await genWitness(circuit, generatedInputs)
             expect(witness.length > 0).toBeTruthy()
 
+            debugger
```
corydickson, line 193:
```suggestion
```

PR #326, opened 2021-12-08
Save each proof in its own file, rather than one big proofs.json

weijiekoh:
When there are many messages, there will be many batches of proofs to store. Previously, MACI saved them in a single JSON file, but clrfund round 8 taught us that if this file gets too large, we won't be able to easily parse it due to the string length limit in Nodejs.

As such, this PR makes genProofs save each proof in its own file in a user-specified proofs directory.

Addresses https://github.com/appliedzkp/maci/issues/316

### DIFF
--- a/integrationTests/ts/__tests__/suites.ts
+++ b/integrationTests/ts/__tests__/suites.ts
@@ -258,7 +258,7 @@ const executeSuite = async (data: any, expect: any) => {
     }
     console.log(e.stdout)
 
-    const removeOldProofs = `rm -f tally.json proofs.json`
+    const removeOldProofs = `rm -f tally.json proofs`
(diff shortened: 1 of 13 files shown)


integrationTests/ts/__tests__/suites.ts
```diff
@@ -258,7 +258,7 @@ const executeSuite = async (data: any, expect: any) => {
     }
     console.log(e.stdout)
 
-    const removeOldProofs = `rm -f tally.json proofs.json`
+    const removeOldProofs = `rm -f tally.json proofs`
```
corydickson, line 261: nit: 
```suggestion
    const removeOldProofs = `rm -rf tally.json proofs`
```

PR #327, opened 2021-12-10
coordinator servcie for maci v1.0.4

chaosma:


cli/package.json
```diff
@@ -27,6 +27,7 @@
         "maci-core": "^1.0.4",
         "maci-crypto": "^1.0.4",
         "maci-domainobjs": "^1.0.4",
+        "maci-common": "^1.0.0",
```
corydickson, line 30: nit: mind using `lerna` so all submodules share the same package version

cli/test.sh
```diff
@@ -35,13 +38,31 @@ node build/index.js genProofs -x 0xf204a4Ef082f5c04bB89F7D5E6568B796096735a \
...
 
 node build/index.js verify \
-    -x 0xf204a4Ef082f5c04bB89F7D5E6568B796096735a \
     -o 0 \
     -t tally.json \
-    -q 0xEcFcaB0A285d3380E488A39B4BB21e777f8A4EaC
+
+node build/index.js mergeMessages -o 1 && \
+node build/index.js mergeSignups -o 1 
+rm -f tally.json proofs.json && \
```
corydickson, line 50: Sorry for the merge conflict but now proofs can span across multiple files for speed improvements

cli/ts/utils.ts
```diff
@@ -1,4 +1,5 @@
 import * as prompt from 'prompt-async'
+import * as fs from 'fs'
```
corydickson, line 2: I think this import is unused

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    && cd maci \
+    && git checkout ${BRANCH_NAME} \
+    && npm i 
+RUN cd ~/maci && npm run bootstrap
+RUN cd ~/maci && npm run build
+RUN cd ~/maci/contracts && npm run compileSol
+
+# HTTP port
+EXPOSE 8080
+EXPOSE 8545
```
corydickson, line 53: Are we assuming this runs on a docker network? This may cause a port collision if one was trying to run the integration tests locally

corydickson:
@chaosma Just couple small things, if you could also rebuild the package-lock files. 

Otherwise this looks great, thank you for the contribution and the clever scripting. 

I will be testing some of the endpoints to see if I run into any issues; we may want to make the commands issued to server verbose by default or give a dev option for debugging.

cli/package.json
corydickson: We probably won't publish this so not sure if it's necessarily required unless you imagine using it for other things in the `server` package

chaosma: some functions are used in both server and cli package, so I put the common functions in a separate folder. What I thought is to put the commonly used utility functions in this folder.

cli/ts/utils.ts
chaosma: removed.

docker/Dockerfile
chaosma: we can either run hardhat inside docker network or on local machine. The hardhat url will be different.

cli/test.sh
chaosma: I have rebased the v1 code. No conflicts now.

daodesigner:
🔥🔥

PR #338, opened 2022-01-05
coordinator service for v0.10.x version

chaosma:


cli/ts/publish.ts
```diff
@@ -179,7 +187,7 @@ const publish = async (args: any) => {
     if (args.prompt_for_eth_privkey) {
         ethSk = await promptPwd('Your Ethereum private key')
     } else {
-        ethSk = args.eth_privkey
+        ethSk = args.eth_privkey?args.eth_privkey:DEFAULT_ETH_SK
```
corydickson, line 190: needs a space between the conditional operator

corydickson:
@chaosma we are gonna have to include the commits that contain the new coordinator service package in this PR as well.

cli/ts/publish.ts
chaosma: updated, thanks!

PR #343, opened 2022-01-12
Remove duplicate function for totalSupply

corydickson:


corydickson:
@kcharbo3

kcharbo3:
Can probably remove the curTokenId field as well and replace it in the giveToken function with a call to totalSupply()?

    // remove: uint256 curTokenId = 1;

    // Gives an ERC721 token to an address
    function giveToken(address to) public onlyOwner {
        _mint(to, totalSupply());
    }

This would start the first token to have id of 0. This is what I usually see, but if you want the first token id to be 1, then you could easily adjust that.

PR #348, opened 2022-01-25
Add new circuit with build script

corydickson:


daodesigner:
think we also need to bump the intermediate state tree depth to increase the tally batch size which I think is calculated as 2**[intermediateStateTreeDepth]
https://github.com/corydickson/maci/blob/increase-batch-size/circuits/circom/quadVoteTally.circom#L29 .

The message batch size affects the BatchUst verifier batch size [when messages are processed](https://github.com/corydickson/maci/blob/increase-batch-size/cli/ts/proveOnChain.ts#L269). The tally batch size affects the QVT verifer [when we submit proofs on chain](https://github.com/corydickson/maci/blob/increase-batch-size/cli/ts/proveOnChain.ts#L319) we should ideally make circuits that double both batch sizes.

circuits/scripts/buildSnarks32BatchSize16.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+cd "$(dirname "$0")"
+cd ..
+mkdir -p params
+
+echo "Building batchUpdateStateTree_32_batch16"
+NODE_OPTIONS=--max-old-space-size=16384 node --stack-size=1073741 build/buildSnarks.js -i circom/prod/batchUpdateStateTree_32_batch16.circom -j params/batchUst32.r1cs -c params/batchUst32.c -y params/batchUst32.sym -p params/batchUstPk32.json -v params/batchUstVk32.json -s params/BatchUpdateStateTreeVerifier32.sol -vs BatchUpdateStateTreeVerifier32 -pr params/batchUst32.params -w params/batchUst32 -a params/batchUst32.wasm
```
daodesigner, line 10: nit: change to 
`-s params/batchUpdateStateTree_32_batch16.sol`
so the names of smart contracts don't overlap

daodesigner (changes requested):

corydickson:
@daodesigner thanks for catching this!

PR #366, opened 2022-02-23
Check parameters and auto-calc intermediate state tree depth

corydickson:


corydickson:
@daodesigner should avoid any confusion can just run: 

```
./scripts/buildCustomSnarks.sh -s 32 -m 32 -v 3 -b 8
```

PR #372, opened 2022-03-14
fix some typos and bugs

chaosma:


circuits/circom/processMessages.circom
```diff
@@ -295,8 +295,7 @@ template ProcessMessages(
             }
         }
 
-        stateRootQip.leaf <== stateRootHasher.hash;
-        stateLeafQip.root === stateRoots[i + 1];
```
corydickson, line 299: I believe we want this check to make sure the state root proofs include the next state root to be processed

corydickson:
@chaosma Good catch overall, if you can just rebase so the package-lock files don't get changed.

circuits/circom/processMessages.circom
chaosma: the stateLeafQip is not defined and only used here, so I removed it. Is it needed here? stateRootQip.leaf changed to stateRootQip[i].leaf.

corydickson: Yes it should be set in the top level of the loop

PR #411, opened 2022-04-18
Topups circuit add state root check

daodesigner:


corydickson:
This commit will be added in a following PR for the topups feature, closing for now to avoid conflicts

PR #434, opened 2022-05-18
Check if `.dat` file exists when running `genProofs` command

gurrpi:
This PR add checks for `.dat` file (one of the artifact of compiled circuit and it is required to execute witness generation) and print error if it doesn't exist.

Currently, when user execute `genProofs` command without `.dat` file,  they will got ambiguous error doesn't address explicitly they missed `.dat` file.

### DIFF
--- a/cli/ts/utils.ts
+++ b/cli/ts/utils.ts
@@ -1,3 +1,4 @@
+import * as fs from 'fs'
 import * as prompt from 'prompt-async'
 
 prompt.colors = false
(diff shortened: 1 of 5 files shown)


corydickson:
The build failing looks unrelated, but looks good!

cli/ts/utils.ts
```diff
@@ -132,6 +133,16 @@ const delay = (ms: number): Promise<void> => {
...
 
+const isPathExist = (paths: Array<string>): [boolean, string] => {
+    for (const path of paths) {
+        if (!fs.existsSync(path)) {
+            return [false, path]
+        }
+    }
+    return [true, null]
+}
+
```
corydickson, line 144: nit: formatting

daodesigner:
🎉 2.0.4 circom

PR #406, opened 2022-04-14
Subsidy calculation

chaosma:

PR #454, opened 2022-06-08
Topup feature

chaosma:
* TopupCredit.sol: 1 token = 1 vote in credit balance.
* Command now has two type: PCommand for vote type message; TCommand for topup message
* Message object has a msgType field: BigInt(1) corresponds to vote type, BigInt(2) corresponds to topup type
* for any type (vote or topup) message, MaciState will generates circuit two part of inputs: topup and vote type. 
* processMessage circuit will process each message as both topup and vote type and get final state change based on message type.
* cli repo added top-up related commands


gurrpi:
Will patch the CI to use the latest code when testing each sub-packages. It is the cause of the failure of test that CI uses dependencies from npm, which doesn’t reflect the change of PR.

core/ts/MaciState.ts
```diff
@@ -337,89 +359,191 @@ class Poll {
...
+                        }
+                        topupStateIndexes.unshift(stateIndex)
+                        topupAmounts.unshift(amount)
+                        topupStateLeaves.unshift(this.stateLeaves[Number(stateIndex)].copy())
+                        topupStateLeavesPathElements.unshift(
+                            this.stateTree.genMerklePath(Number(stateIndex)).pathElements
+                        )
+
+                        const newStateLeaf = this.stateLeaves[Number(stateIndex)].copy()
+                        newStateLeaf.voiceCreditBalance = BigInt(newStateLeaf.voiceCreditBalance) + BigInt(amount)
```
gurrpi, line 513: It causes [build failure]((https://github.com/privacy-scaling-explorations/maci/runs/6802556146?check_suite_focus=true#step:6:60)):
```
Error: ts/MaciState.ts(513,66): error TS2345: Argument of type 'BigInt' is not assignable to parameter of type 'string | number | bigint | boolean'.
```

 Seems its the same case as previous error (see [this](https://github.com/privacy-scaling-explorations/maci/pull/406#discussion_r851608061)).

package.json
```diff
@@ -14,8 +14,14 @@
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "^2.31.0",
     "@typescript-eslint/parser": "^2.31.0",
+    "cz-conventional-changelog": "^3.3.0",
```
gurrpi, line 17: +1

chaosma:
> Seems current implementation presumes that the topup feature is default to MACI setup. If so, merging the code of `cli/ts/deployTopupCredit.ts` into `cli/ts/create.ts` would decrease operational complexity in user level.

Good idea. I had moved the deployTopupCredit into create.ts

PR #463, opened 2022-06-21
Topup clean

daodesigner:


daodesigner:
@gurrpi @chaosma can't figure out why test is failing, looks like it's failing to deploy the ppt contract? 
errors out at this line on CI.  thoughts?
`3 |     const pptAddress = deployPollRegMatch[1]`

PR #468, opened 2022-06-24
Bump the circom to v2.0.5

gurrpi:
Related release note: https://github.com/iden3/circom/commit/8a97f51dfb955b867f33a447b707dd445e5e15b1


gurrpi:
> I think it would be a good idea when these updates happen to keep a copy of some of the build artifacts in the repo just so we can compare updates.
> 
> For example there is a tool that's used to manage zkeys, and while I doubt this would break, it'd be nice to have a file that verifies things work beforehand.

It make sense. It would be good to track build time and artifact size by version of build tool.

PR #474, opened 2022-07-21
Trying to solve two TS type errors in `crypto/`

0xisk:
- Argument of type 'BigIntConstructor' is not assignable to parameter of type '(value: BigInt, index: number, array: BigInt[]) => bigint'.
- Type 'BigInt' is not assignable to type 'string | number | bigint | boolean'.

### DIFF
--- a/crypto/ts/IncrementalQuinTree.ts
+++ b/crypto/ts/IncrementalQuinTree.ts
@@ -75,7 +75,7 @@ class IncrementalQuinTree {
 
         this.depth = Number(_depth)
         this.nextIndex = 0
-        this.zeroValue = BigInt(_zeroValue)
+        this.zeroValue = BigInt(_zeroValue as bigint)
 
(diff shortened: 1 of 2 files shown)


iskdrews:
I am not sure about the first error fixing: 
```
const level: BigInt[] = pathElements[i].map((i: BigInt): BigInt => {
     return BigInt(i as bigint);
}) 
```

ctrlc03:
> I am not sure about the first error fixing:
> 
> ```
> const level: BigInt[] = pathElements[i].map((i: BigInt): BigInt => {
>      return BigInt(i as bigint);
> }) 
> ```

Thank you so much for opening this, and sorry it took us this long to get back to you. As discussed on discord, we'll be closing this PR the code has been updated since then, and some refactoring work tackling similar issues is about to be merged into the repo, thank you for making this problem evident.

PR #520, opened 2022-09-06
fix: errors in 02_contract.md

stanbar:
I've spotted a few minor errors in the document.

Line 166: if e,d, and c are valid then the weight sums up to 11, not 10.

Line 168: (b) gives a nonce of (1) not (2)

Line 175: There is no vote (c), only (a) and (b). Rename (c) to (b).

I've also unified two styles used to list votes `a.` and `a)` to `(a)` (so it is consistent with the description). Moreover, I have added `- (a)` for a better post-rendering list 
readability.

### DIFF
--- a/website/versioned_docs/version-v0.x/contract.md
+++ b/website/versioned_docs/version-v0.x/contract.md
@@ -156,29 +156,29 @@ The last valid message per user should have a nonce of `1`. Each valid message t
 
 For example, Alice publishes 5 messages, all of which vote for the same option:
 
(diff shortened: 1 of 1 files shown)


Abuchtela (approved):

samajammin:
Thanks for this! Looks good, just some merge conflicts now that @ctrlc03 will address 👍

Abuchtela (approved):
contract.md #520

baumstern (approved):
LGTM

PR #558, opened 2022-12-02
feat(contracts): made `nextPollId` public 

ctrlc03:
Changed visibility of `nextPollId` to allow easier integration in third party apps. Also removed redundant variables initialization

#fix 557


gurrpi:
LGTM!

PR #562, opened 2022-12-05
Fix/server

ctrlc03:
Added a whitelist of allowed input from HTTP requests that are inserted dynamically into a cli command


gurrpi (approved):
LGTM!

Issue #457, opened 2022-06-09
Lint errors in `master`


baumstern:
There are some errors when running `npm run lint` :
```
✖ 119 problems (79 errors, 40 warnings)
  67 errors and 0 warnings potentially fixable with the `--fix` option.
```

### Lint log

```
% npm run lint

> maci@1.1.1 lint
> eslint './**/**/*.ts'


/Users/daehyun/pse/playground/maci/circuits/ts/__tests__/Hasher.test.ts
  210:30  warning  'pubKey' is assigned a value but never used  @typescript-eslint/no-unused-vars

/Users/daehyun/pse/playground/maci/circuits/ts/__tests__/MessageToCommand.test.ts
  21:26  warning  'pubKey' is assigned a value but never used  @typescript-eslint/no-unused-vars

/Users/daehyun/pse/playground/maci/circuits/ts/__tests__/ProcessMessages.test.ts
  322:15  warning  'voteOptionIndex' is assigned a value but never used  @typescript-eslint/no-unused-vars

/Users/daehyun/pse/playground/maci/circuits/ts/__tests__/TallyVotes.test.ts
  42:7  warning  'testProcessVk' is assigned a value but never used  @typescript-eslint/no-unused-vars
  53:7  warning  'testTallyVk' is assigned a value but never used    @typescript-eslint/no-unused-vars

/Users/daehyun/pse/playground/maci/cli/ts/airdrop.ts
   60:9   error  'contractAddrs' is never reassigned. Use 'const' instead      prefer-const
  127:13  error  'MAXIMUM_ALLOWANCE' is never reassigned. Use 'const' instead  prefer-const

/Users/daehyun/pse/playground/maci/cli/ts/checkVerifyingKey.ts
    1:13  warning  'fs' is defined but never used                            @typescript-eslint/no-unused-vars
    2:13  warning  'shelljs' is defined but never used                       @typescript-eslint/no-unused-vars
  113:9   error    'contractAddrs' is never reassigned. Use 'const' instead  prefer-const

/Users/daehyun/pse/playground/maci/cli/ts/create.ts
  61:9  error  'contractAddrs' is never reassigned. Use 'const' instead  prefer-const

/Users/daehyun/pse/playground/maci/cli/ts/deployPoll.ts
  123:9  error  'contractAddrs' is never reassigned. Use 'const' instead  prefer-const
  196:2  error  Mixed spaces and tabs                                     no-mixed-spaces-and-tabs

[...]

PR #585, opened 2022-12-22
refactor(cli): print relevant values when comparing fails

baumstern:
## Better Logging

Before: only error message has been printed
After: error message and relevant values will be printed

### DIFF
--- a/cli/ts/proveOnChain.ts
+++ b/cli/ts/proveOnChain.ts
@@ -10,6 +10,7 @@ import {
 } from 'maci-contracts'
 
 import {
+    compareOnChainValue,
     validateEthAddress,
     contractExists,
     delay,
@@ -304,18 +305,14 @@ const proveOnChain = async (args: any) => {
             currentMessageBatchIndex -= messageBatchSize
         }
 
-        const txErr = 'Error: processMessages() failed'
         const { proof, circuitInputs, publicInputs } = data.processProofs[i]
 
         // Perform checks
-        if (circuitInputs.pollEndTimestamp !== pollEndTimestampOnChain.toString()) {
-            console.error('Error: pollEndTimestamp mismatch.')
-            return 
+        if (!compareOnChainValue("pollEndTimestamp", pollEndTimestampOnChain.toString(), circuitInputs.pollEndTimestamp)) {
+            return
         }
-
-        if (BigInt(circuitInputs.msgRoot).toString() !== messageRootOnChain.toString()) {
-            console.error('Error: message root mismatch.')
-            return 
+        if (!compareOnChainValue("message root", messageRootOnChain.toString(), BigInt(circuitInputs.msgRoot).toString())) {
+            return
         }
 
         let currentSbCommitmentOnChain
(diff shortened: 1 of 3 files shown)


gurrpi:
> @gurrpi what's the status on this? Do you think this is worth finishing/merging? Or should we close this out?

I've just triggered a new CI run. Once the checks pass, I think this is worth merging.

samajammin:
Nudge on this @baumstern

baumstern:
@ctrlc03 PTAL

cli/ts/utils.ts
```diff
@@ -142,6 +142,21 @@ const isPathExist = (paths: Array<string>): [boolean, string] => {
     return [true, null]
 }
 
+const CompareOnchainValue = (
```
samajammin, line 145:
```suggestion
const compareOnchainValue = (
```
General convention - functions should always be camelCase, ya?

Noting this commit suggestion would require updating all uses of this function...

cli/ts/utils.ts
```diff
@@ -142,6 +142,21 @@ const isPathExist = (paths: Array<string>): [boolean, string] => {
...
+    name: string,
+    onChainValue: any,
```
samajammin, line 147: Another naming convention nitpick - here you capitalize the "C" in `onChain` but you didn't in the function name. I suggest we be consistent (i.e. use capital "C" for both)

cli/ts/utils.ts
samajammin: Edit - also capitalized the "C". See comment below for context

samajammin:
@baumstern any particular reason you PR'd from your fork vs. creating a branch directly off the repo?

cli/ts/proveOnChain.ts
```diff
@@ -380,6 +375,7 @@ const proveOnChain = async (args: any) => {
             return 
         }
 
+        const txErr = 'Error: processMessages() failed'
```
samajammin, line 378: Another consistency nitpick - looks like you're using a mix of single quotes & double quotes. Do we have a convention on this? I don't have a strong preference... just that we be consistent

cli/ts/utils.ts
```diff
@@ -142,6 +142,21 @@ const isPathExist = (paths: Array<string>): [boolean, string] => {
...
+    offChainValue: any,
+): boolean => {
+    if (onChainValue !== offChainValue) {
+        console.error(`Error: ${name} mismatch.`)
+        console.error("     onChainValue: " + onChainValue)
```
samajammin, line 152: Why the spacing on these logs?

samajammin (changes requested):
A few styling nitpicks - see my comments

cli/ts/utils.ts
baumstern: > General convention - functions should always be camelCase, ya?
> 
> Noting this commit suggestion would require updating all uses of this function...

Good catch! Reflected in [974ab46](https://github.com/privacy-scaling-explorations/maci/pull/585/commits/974ab462a55ad39f0df76ce16972c53446480589)

cli/ts/utils.ts
baumstern: Having seen `onchain` used more commonly, I've decided to go with that. Thanks for the input!

cli/ts/utils.ts
baumstern: The spacing in these logs is intended to make the output more noticeable and easier to read.

baumstern:
Inconsistency has inherited from legacy. Filed an issues for that
* https://github.com/privacy-scaling-explorations/maci/issues/811
* https://github.com/privacy-scaling-explorations/maci/issues/810

> @baumstern any particular reason you PR'd from your fork vs. creating a branch directly off the repo?

I think I didn't dare to add a branch to the upstream repo at that moment 😂

baumstern:
> A few styling nitpicks - see my comments

The inconsistency has been inherited from legacy code. I believe it would be better to address the codebase's overall consistency, so I have filed separate issues to tackle this:

    [Issue 811](https://github.com/privacy-scaling-explorations/maci/issues/811)
    [Issue 810](https://github.com/privacy-scaling-explorations/maci/issues/810)

samajammin:
> The inconsistency has been inherited from legacy code

I'm just referring to your changes in this PR - you use both `Onchain` & `OnChain`, as well as single/double quotes. So I'd start by just addressing this PR. i.e. be consistent with your individual commits.

But agree, would be good to unify across the codebase as well 👍

baumstern:
> > The inconsistency has been inherited from legacy code
> 
> I'm just referring to your changes in this PR - you use both `Onchain` & `OnChain`, as well as single/double quotes. So I'd start by just addressing this PR. i.e. be consistent with your individual commits.
> 
> But agree, would be good to unify across the codebase as well 👍

That's reasonable. Fixed the inconsitency of this PR in 2e26126 to uses double quotes and `OnChain` across the change.

Lint errors such as those in https://github.com/privacy-scaling-explorations/maci/issues/457 should be fixed and automatically enforced linting in pre-commit hook would ensure us to have consistent style after that.

samajammin (approved):

PR #599, opened 2023-01-31
Contract refactor

chaosma:
* separation of MessageProcessor, Tally and Subsidy contracts out according to the circuits
* various minor optimizations on contracts
* fixed broken unit tests


contracts/contracts/Tally.sol
```diff
@@ -0,0 +1,248 @@
...
+            _proof,
+            numSignUps,
+            batchStartIndex,
+            tallyBatchSize,
+            _newTallyCommitment
+        );
+        require(isValid, ERROR_INVALID_TALLY_VOTES_PROOF);
+
+        // Update the tally commitment and the tally batch num
+        tallyCommitment = _newTallyCommitment;
```
ctrlc03, line 133: Move these two lines up at the top (checks effect interaction pattern). You can store `tallyBatchNum` in a local variable and use that for calculations:

```js
uint256 _tallyBatchNum = tallyBatchNum;
unchecked {
    tallyBatchNum++;
}
```

contracts/contracts/Tally.sol
```diff
@@ -0,0 +1,248 @@
...
+     */
+    function verifyTallyResult(
+        Poll _poll,
+        uint256 _voteOptionIndex,
+        uint256 _tallyResult,
+        uint256[][] memory _tallyResultProof,
+        uint256 _spentVoiceCreditsHash,
+        uint256 _perVOSpentVoiceCreditsHash,
+        uint256 _tallyCommitment
+    ) public view returns (bool) {
```
ctrlc03, line 232: Don't see this function being used anywhere in the contract, so can be made external

contracts/contracts/Tally.sol
```diff
@@ -0,0 +1,248 @@
...
+        (, , , uint8 voteOptionTreeDepth) = _poll.treeDepths();
+        uint256 computedRoot = computeMerkleRootFromPath(
+            voteOptionTreeDepth,
+            _voteOptionIndex,
+            _tallyResult,
+            _tallyResultProof
```
ctrlc03, line 238: Is this correct? passing the proof here?

contracts/contracts/Tally.sol
```diff
@@ -0,0 +1,248 @@
...
+import {VkRegistry} from "./VkRegistry.sol";
+
+
+contract Tally is
+    Ownable,
+    SnarkCommon,
+    Hasher
+{
+    // Error codes
+    string constant ERROR_VOTING_PERIOD_NOT_PASSED = "TallyE01";
```
ctrlc03, line 21: I would say it might be worth to migrating to custom errors throughout the code base, and remove all require statements for if checks. What you think? should help save some gas

ctrlc03: i.e.

```js
error VOTING_PERIOD_NOT_OVER();
error PROCESSING_NOT_COMPLETE();
error INVALID_TALLY_VOTES_PROOF();
error ALL_BALLOTS_TALLIED();
```

contracts/contracts/Tally.sol
```diff
@@ -0,0 +1,248 @@
...
+        // Require that the voting period is over
+        uint256 secondsPassed = block.timestamp - deployTime;
+        require(secondsPassed > duration, ERROR_VOTING_PERIOD_NOT_PASSED);
+        _;
+    }
+
+    /*
+     * Pack the batch start index and number of signups into a 100-bit value.
+     */
+    function genTallyVotesPackedVals(
```
ctrlc03, line 65: Let's add Natspec to all functions

contracts/contracts/Tally.sol
```diff
@@ -0,0 +1,248 @@
...
+    // The final commitment to the state and ballot roots
+    uint256 public sbCommitment;
+
+    Verifier public verifier;
+
+    constructor(Verifier _verifier) {
+        verifier = _verifier;
+    }
+
+    modifier votingPeriodOver(Poll _poll) {
```
ctrlc03, line 53: let's move this to a private function, and add where appropriate in the functions. will be sligthly cheaper

contracts/contracts/Tally.sol
```diff
@@ -0,0 +1,248 @@
...
+
+    /*
+     * Pack the batch start index and number of signups into a 100-bit value.
+     */
+    function genTallyVotesPackedVals(
+        uint256 _numSignUps,
+        uint256 _batchStartIndex,
+        uint256 _tallyBatchSize
+    ) public pure returns (uint256) {
+        // TODO: ensure that each value is less than or equal to 2 ** 50
```
ctrlc03, line 70: let's implement this check and get rid of the TODO?

contracts/contracts/Tally.sol
```diff
@@ -0,0 +1,248 @@
...
+        input[1] = sbCommitment;
+        input[2] = tallyCommitment;
+        input[3] = _newTallyCommitment;
+        uint256 inputHash = sha256Hash(input);
+        return inputHash;
+    }
+
+    // TODO: make sure correct mp address is passed or change to private function
+    // TODO: reuse tally.sol for multiple polls
+    function updateSbCommitment(MessageProcessor _mp) public {
```
ctrlc03, line 99: let's make this private for now? it's only called in tallyVotes and tallyVotes is owner only so should be ok

contracts/contracts/Utility.sol
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: MIT
```
ctrlc03, line 1: To keep the repo a bit more organized, should we move utilities contract like this one to their own folder named for instance utilities/helpers?

contracts/contracts/Tally.sol
```diff
@@ -0,0 +1,248 @@
...
+    // )
+    //
+    // Where each salt is unique and the merkle roots are of arrays of leaves
+    // TREE_ARITY ** voteOptionTreeDepth long.
+    uint256 public tallyCommitment;
+
+    uint256 public tallyBatchNum;
+
+    // The final commitment to the state and ballot roots
+    uint256 public sbCommitment;
```
ctrlc03, line 45: If we want to use one Tally contract for multiple polls, we could implement a mapping poll => sbCommitment and same for the tallyCommitment and batch num

contracts/contracts/Subsidy.sol
```diff
@@ -0,0 +1,162 @@
...
+{
+
+    uint256 public rbi; // row batch index
+    uint256 public cbi; // column batch index
+    // The final commitment to the state and ballot roots
+    uint256 public sbCommitment;
+    uint256 public subsidyCommitment;
+
+    // Error codes
+    string constant ERROR_VOTING_PERIOD_NOT_PASSED = "SubsidyE01";
```
ctrlc03, line 26: Let's move to custom errors: error VOTING_PERIOD_NOT_PASSED(); etc.

contracts/contracts/Subsidy.sol
```diff
@@ -0,0 +1,162 @@
...
+    string constant ERROR_ALL_SUBSIDY_CALCULATED = "SubsidyE04";
+    string constant ERROR_VK_NOT_SET = "SubsidyE05";
+
+    Verifier public verifier;
+
+    constructor(Verifier _verifier) {
+        verifier = _verifier;
+    }
+
+    modifier votingPeriodOver(Poll _poll) {
```
ctrlc03, line 38: could move this to a internal/private function. I see the same modifier is in Tally.sol so we could move it to an utilities contract (CommonUtilities) as internal and inherit from it and use it that way

contracts/contracts/Subsidy.sol
```diff
@@ -0,0 +1,162 @@
...
+            .getDeployTimeAndDuration();
+        // Require that the voting period is over
+        uint256 secondsPassed = block.timestamp - deployTime;
+        require(secondsPassed > duration, ERROR_VOTING_PERIOD_NOT_PASSED);
+        _;
+    }
+
+    // TODO: make sure correct mp address is passed or change to private function
+    // TODO: reuse subsidy.sol for multiple polls
+    function updateSbCommitment(MessageProcessor _mp) public {
```
ctrlc03, line 49: This should be internal/private or moved to an helpers contract so that we can re use in Tally.sol

contracts/contracts/Subsidy.sol
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+
+    function genSubsidyPackedVals(uint256 _numSignUps)
+        public
+        view
+        returns (uint256)
+    {
+        // TODO: ensure that each value is less than or equal to 2 ** 50
+        uint256 result = (_numSignUps << uint256(100)) +
```
ctrlc03, line 64: The cast to uint256 here seems redundant

contracts/contracts/Subsidy.sol
```diff
@@ -0,0 +1,162 @@
...
+        Poll _poll,
+        MessageProcessor _mp,
+        uint256 _newSubsidyCommitment,
+        uint256[8] memory _proof
+    ) public onlyOwner votingPeriodOver(_poll) {
+        updateSbCommitment(_mp);
+
+        (uint8 intStateTreeDepth, , , uint8 voteOptionTreeDepth) = _poll
+            .treeDepths();
+        uint256 subsidyBatchSize = 5**intStateTreeDepth; // treeArity is fixed to 5
```
ctrlc03, line 95: let's not use hardcoded values like this, but constants instead

contracts/contracts/Subsidy.sol
```diff
@@ -0,0 +1,162 @@
...
+        (uint256 numSignUps, ) = _poll.numSignUpsAndMessages();
+        uint256 numLeaves = numSignUps + 1;
```
ctrlc03, line 97: we can use an unchecked block here as these cannot overflow 5 ** uint8 cannot overflow an uint256
and numSignups + 1 also very very unlikely to overflow

contracts/contracts/Subsidy.sol
```diff
@@ -0,0 +1,162 @@
...
+
+        bool isValid = verifySubsidyProof(
+            _poll,
+            _proof,
+            numSignUps,
+            _newSubsidyCommitment
+        );
+        require(isValid, ERROR_INVALID_SUBSIDY_PROOF);
+        subsidyCommitment = _newSubsidyCommitment;
+        increaseSubsidyIndex(subsidyBatchSize, numLeaves);
```
ctrlc03, line 117: would be great to move this state change up in the function. It will require some caching to local variable

contracts/contracts/Subsidy.sol
```diff
@@ -0,0 +1,162 @@
...
+        uint256 inputHash = sha256Hash(input);
+        return inputHash;
+    }
+
+    function updateSubsidy(
+        Poll _poll,
+        MessageProcessor _mp,
+        uint256 _newSubsidyCommitment,
+        uint256[8] memory _proof
+    ) public onlyOwner votingPeriodOver(_poll) {
```
ctrlc03, line 90: we can make this external as it's not called anywhere within this contract

contracts/contracts/Subsidy.sol
```diff
@@ -0,0 +1,162 @@
...
+            _proof,
+            numSignUps,
+            _newSubsidyCommitment
+        );
+        require(isValid, ERROR_INVALID_SUBSIDY_PROOF);
+        subsidyCommitment = _newSubsidyCommitment;
+        increaseSubsidyIndex(subsidyBatchSize, numLeaves);
+    }
+
+    function increaseSubsidyIndex(uint256 batchSize, uint256 numLeaves)
```
ctrlc03, line 120: Let's add documentation to all functions please, as some more comments within the code, for instance this function needs some explainer

contracts/contracts/MessageProcessor.sol
```diff
@@ -0,0 +1,250 @@
...
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+import {Poll} from "./Poll.sol";
+import {SnarkCommon} from "./crypto/SnarkCommon.sol";
+import {CommonUtilities} from "./Utility.sol";
+import {Verifier} from "./crypto/Verifier.sol";
+import {VkRegistry} from "./VkRegistry.sol";
+
+contract MessageProcessor is Ownable, SnarkCommon, CommonUtilities {
+
+    string constant ERROR_VOTING_PERIOD_PASSED = "ProcessE01";
```
ctrlc03, line 15: As detailed in the other comments, I think it'd be great to move to custom errors and replace require statements with if checks

contracts/contracts/MessageProcessor.sol
```diff
@@ -0,0 +1,250 @@
...
+   // The commitment to the state and ballot roots
+    uint256 public sbCommitment;
+
+    Verifier public verifier;
+
+    constructor(Verifier _verifier) {
+        verifier = _verifier;
+    }
+
+    modifier votingPeriodOver(Poll _poll) {
```
ctrlc03, line 41: Let's move this to the CommonUtilities contract as an internal function so we can use it everywhere else

contracts/contracts/MessageProcessor.sol
```diff
@@ -0,0 +1,250 @@
...
+        (uint256 deployTime, uint256 duration) = _poll
+            .getDeployTimeAndDuration();
+        // Require that the voting period is over
+        uint256 secondsPassed = block.timestamp - deployTime;
```
ctrlc03, line 45: We can move this calculation into an unchecked block, as the current time will always be greater than the deploy time

contracts/contracts/MessageProcessor.sol
```diff
@@ -0,0 +1,250 @@
...
+     * @param _poll The poll to update
+     * @param _newSbCommitment The new state root and ballot root commitment
+     *                         after all messages are processed
+     * @param _proof The zk-SNARK proof
+     */
+    function processMessages(
+        Poll _poll,
+        uint256 _newSbCommitment,
+        uint256[8] memory _proof
+    ) public onlyOwner votingPeriodOver(_poll) {
```
ctrlc03, line 61: we can make this external as it's not called within this contract

contracts/contracts/MessageProcessor.sol
```diff
@@ -0,0 +1,250 @@
...
+            _poll,
+            currentMessageBatchIndex,
+            messageRoot,
+            sbCommitment,
+            _newSbCommitment,
+            _proof
+        );
+        require(isValid, ERROR_INVALID_PROCESS_MESSAGE_PROOF);
+
+        {
```
ctrlc03, line 114: let's move state update to the top of the function. Will require caching the state variables as local variables so they can be used in the function body

contracts/contracts/Poll.sol
```diff
@@ -3,18 +3,18 @@ pragma solidity ^0.8.10;
...
 import {AccQueue, AccQueueQuinaryMaci} from "./trees/AccQueue.sol";
 import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
 import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import {VkRegistry} from "./VkRegistry.sol";
+import {Verifier} from "./crypto/Verifier.sol";
 import {EmptyBallotRoots} from "./trees/EmptyBallotRoots.sol";
 import {TopupCredit} from "./TopupCredit.sol";
+import {Utilities} from "./Utility.sol";
+import {MessageProcessor} from "./MessageProcessor.sol";
```
ctrlc03, line 17: this is not used anywhere

contracts/contracts/MessageProcessor.sol
```diff
@@ -0,0 +1,250 @@
...
+        (uint24 mbs, , ) = _poll.batchSizes();
+        uint256 messageBatchSize = uint256(mbs);
+
+        uint256 batchEndIndex = _currentMessageBatchIndex + messageBatchSize;
+        if (batchEndIndex > numMessages) {
+            batchEndIndex = numMessages;
+        }
+
+        uint256 result = maxVoteOptions +
+            (_numSignUps << uint256(50)) +
```
ctrlc03, line 230: casting here is not needed

contracts/contracts/MessageProcessor.sol
```diff
@@ -0,0 +1,250 @@
...
+    /*
+     * Returns the SHA256 hash of the packed values (see
+     * genProcessMessagesPackedVals), the hash of the coordinator's public key,
+     * the message root, and the commitment to the current state root and
+     * ballot root. By passing the SHA256 hash of these values to the circuit
+     * as a single public input and the preimage as private inputs, we reduce
+     * its verification gas cost though the number of constraints will be
+     * higher and proving time will be higher.
+     */
+    function genProcessMessagesPublicInputHash(
```
ctrlc03, line 176: Let's add proper Natspec documentation to each function

contracts/contracts/MessageProcessor.sol
```diff
@@ -0,0 +1,250 @@
...
+
+        uint256 result = maxVoteOptions +
+            (_numSignUps << uint256(50)) +
+            (_currentMessageBatchIndex << uint256(100)) +
+            (batchEndIndex << uint256(150));
+
+        return result;
+    }
+
+    function updateMessageProcessingData(
```
ctrlc03, line 237: Natspec missing

contracts/contracts/MessageProcessor.sol
```diff
@@ -0,0 +1,250 @@
...
+import {AccQueue} from "./trees/AccQueue.sol";
+import {IMACI} from "./IMACI.sol";
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+import {Poll} from "./Poll.sol";
+import {SnarkCommon} from "./crypto/SnarkCommon.sol";
+import {CommonUtilities} from "./Utility.sol";
+import {Verifier} from "./crypto/Verifier.sol";
+import {VkRegistry} from "./VkRegistry.sol";
+
+contract MessageProcessor is Ownable, SnarkCommon, CommonUtilities {
```
ctrlc03, line 13: Let's add some comments to explain the purpose of this contract

contracts/contracts/Tally.sol
chaosma: changed public to external. Just keep it for now in case we need it later.

contracts/contracts/Tally.sol
chaosma: haven't checked or test this function, it was never used.

chaosma: we might delete it if we don't need it

contracts/contracts/Subsidy.sol
chaosma: let me change it to internal to avoid deploy another contract

contracts/contracts/Tally.sol
chaosma: yes, that's also my thought. I plan test it later, not in this PR.

contracts/contracts/Subsidy.sol
chaosma: can modifier be inherited as well?

ctrlc03: you can yes, but we can also make it an internal function and inherit that way.

daodesigner (changes requested):
[#630](https://github.com/privacy-scaling-explorations/maci/issues/630)

contracts/contracts/Tally.sol
chaosma: Removed. It's never been called

contracts/contracts/Tally.sol
chaosma: Sounds good. But let's put it into the next PR to separate refactor and optimization.

contracts/contracts/Tally.sol
chaosma: added

contracts/contracts/Tally.sol
chaosma: After this modification, it will cause stack too deep in this function. Since onlyOwner can call this function, it should be okay. What do you think?

ctrlc03: Right ok, it should  be fine to move only the commitment update on top and keep the counter increased down given that it's protected by onlyOwner. We can always try and refactor it later.

contracts/contracts/Subsidy.sol
ctrlc03: Can't you call `updateSubsidy` from proveOnChain.ts? Or does this function need to be called in other occasions too?

contracts/contracts/Tally.sol
chaosma: do you mean by this:
```
private modifier votingPeriodOver(Poll _poll) {
}
```

contracts/contracts/Tally.sol
chaosma: it has been called in proveOnChain.ts. The reason is that after the deployment of the new Tally.sol contract. We have to update the sbCommitment for the first time by calling this function. So if we change it to private, we cannot call it. Let me rephrase the TODO here.

contracts/contracts/Tally.sol
ctrlc03: sorry what I meant is rather than using a modifier we could use a private/internal function which does the checks of the modifier:

```js
function _votingPeriodOver(Poll _poll) private view {
   if (x != y) revert SomeError();
}
```

up to you if you want to go down this route and change all modifiers to private functions. Contract size will be smaller and cheaper to deploy, might result in a small const increase on the user side though.

contracts/contracts/Utility.sol
chaosma: sounds good!

contracts/contracts/Tally.sol
chaosma: Yeah, that should work.

contracts/contracts/Tally.sol
chaosma: yeah, use the internal function instead based on our discussion offline.

contracts/contracts/Subsidy.sol
chaosma: added docs for this function

contracts/contracts/Subsidy.sol
chaosma: done

contracts/contracts/Poll.sol
chaosma: it is used in contract Poll (the inheritance part)

contracts/contracts/MessageProcessor.sol
chaosma: done

contracts/contracts/MessageProcessor.sol
chaosma: done

contracts/contracts/MessageProcessor.sol
chaosma: yeah

contracts/contracts/MessageProcessor.sol
chaosma: but is it possible to overflow uint256

chaosma:
thanks for the comments. many good optimizations

contracts/contracts/Subsidy.sol
chaosma: removed uncheck for potential overflow

daodesigner (approved):

PR #708, opened 2023-06-27
Feat/elgamal

aleksandar-veljkovic:
**Milestone 2**
- Key deactivation functionality (smart contracts/circuits/CLI commands)
- Docs
- Tests


aleksandar-veljkovic:
All green!
Of 97 tests, 5 contract tests are skipped due to changes in the protocol implementation which affects the original tests Namely, merging state tree produces error if it is requested on the already merged state tree - which is now possible in the key deactivation phase. End-2-end tests are passing, which proves that the original protocol is not affected. Additional tests will be added in the milestone 3.

daodesigner (commented):
lgtm

PR #747, opened 2023-10-18
Test case to show anyone can deploy poll

yuetloo:
The test case duplicates the existing deployPoll.ts script and change it to accept a new argument, a private key of a signer which will deploy the poll.

The test case first run the first test case from the existing integration test suite to setup the environment (like deploying MACI and Poll contracts). After the first poll is tallied and finalized, the test case start a new poll using a randomly created wallet.

How to run the test:
1) Setup MACI as described in the readme file
2) run the test
```
cd integrationTests
npm run download-zkeys
npm run test-deployPoll
```

The test should pass, which means a new poll was successfully created by a random wallet.

### DIFF
--- a/cli/ts/deployPollWithSigner.ts
+++ b/cli/ts/deployPollWithSigner.ts
@@ -1,0 +1,244 @@
+const { ethers } = require('hardhat')
+import {
+    parseArtifact,
+    deployVerifier,
+    deployPpt,
+    getDefaultSigner,
+} from 'maci-contracts'
+
+import {
+    PubKey,
+} from 'maci-domainobjs'
+
+import {contractExists} from './utils'
+import {readJSONFile, writeJSONFile} from 'maci-common'
+import {contractFilepath} from './config'
+
+const configureSubparser = (subparsers: any) => {
+    const createParser = subparsers.addParser(
+        'deployPollWithSigner',
+        { addHelp: true },
+    )
+
+    createParser.addArgument(
+        ['-s', '--signer-private-key'],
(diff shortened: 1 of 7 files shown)


ctrlc03:
> The test case duplicates the existing deployPoll.ts script and change it to accept a new argument, a private key of a signer which will deploy the poll.
> 
> The test case first run the first test case from the existing integration test suite to setup the environment (like deploying MACI and Poll contracts). After the first poll is tallied and finalized, the test case start a new poll using a randomly created wallet.
> 
> How to run the test:
> 
> 1. Setup MACI as described in the readme file
> 2. run the test
> 
> ```
> cd integrationTests
> npm run download-zkeys
> npm run test-deployPoll
> ```
> 
> The test should pass, which means a new poll was successfully created by a random wallet.

Thanks for opening this. I'm thinking now, we have two ways to go:
1. we can merge this PR (please first change target branch to dev) and I will add the rest 
2. you can add the onlyOwner modifier too in this PR and possibly change the tests that might be affected (if any on the contract side). 

Let me know how you'd like to proceed

yuetloo:
@ctrlc03 , I changed the target branch to dev.  I'm not sure if you want to keep the `download_zkeys.sh` script or not. I created it because I couldn't download the zkeys in the `scripts/runTestsInCi.sh`.  I noticed that the CI job that runs agains the master branch generates the `zkeys` instead..

ctrlc03 (approved):
LGTM, will merge and then merge #765

Issue #729, opened 2023-10-15
Documentation formatting bug?


samajammin:
Not certain but I suspect there's bugs around displaying mathematical formulas...

e.g. on https://privacy-scaling-explorations.github.io/maci/primitives.html

![Image 2023-10-15 at 12 32 06](https://github.com/privacy-scaling-explorations/maci/assets/8097623/eaaccb5c-488f-4d29-8d81-5aac28c82895)

Is the documentation supposed to be able to render LaTeX or something similar?

samajammin:
FWIW we've had success in past projects (e.g. https://github.com/ethereum/cryptography-research-website) with KaTeX:
https://katex.org/

ctrlc03:
> FWIW we've had success in past projects (e.g. https://github.com/ethereum/cryptography-research-website) with KaTeX: https://katex.org/

suggesting we use this issue to track researching how to preview this documentation before a PR is merged, so errors like this are spotted before merging.

PR #751, opened 2023-10-19
docs(trusted setup) - small refresh of the trusted setup documentation

ctrlc03:


docs/trustedsetup.md
```diff
@@ -10,28 +10,25 @@ to manage a large amount of value, and if their users accept that the risk of
...
-
-You should perform at least one contribution to each circuit, even if you
-choose not to perform a multi-party trusted setup.
-
-We don't recommend a browser-based trusted setup (which [Tornado
-Cash](https://ceremony.tornado.cash/) and [Zkopru](https://mpc.zkopru.network/)
-used) for the MACI circuits as they are too large to be feasibly processed in
-the browser.
-
+## How to run your own trusted setup
```
0xjei, line 24: Why are we quoting this one? Isn't a cross-reference in the paragraph above sufficient?

It seems to me "too trusted setup tooling centric" and that it has little to do with MACI tbh.

0xjei (changes requested):

docs/trustedsetup.md
0xjei: Ah, in that case even less imho. They could use another tool and there's no point in us telling them how they should launch with p0tion here if we already link above docs. It seems like repetition to me and I would keep the docs concise.

0xjei (approved):

docs/trustedsetup.md
```diff
@@ -10,28 +10,14 @@ to manage a large amount of value, and if their users accept that the risk of
 coordinator misbehaviour is insufficient to justify doing the work nof a
 trusted setup. After all, MACI's security model presumes a trusted coordinator.
 
+The PSE team is making available a trusted set of zKeys for MACI's circuits,
+which will be available as soon as the first ceremony is completed and
```
baumstern, line 14:
```suggestion
The PSE team is in the process of creating a trusted setup for MACI's circuits. The public parameters will be available as soon as the first ceremony is completed and
```

I'd recommend clarifying the terminology for better technical accuracy. Instead of "trusted set," the more standard term is "trusted setup." If you're referring to the output of this setup, it might be better to use "public parameters."

docs/trustedsetup.md
```diff
@@ -10,28 +10,14 @@ to manage a large amount of value, and if their users accept that the risk of
...
+will be visible on
+[p0tion's website](https://ceremony.pse.dev/projects/Maci%20v1%20Trusted%20Setup%20Ceremony).
+For more info on the ceremony please refer to p0tion's [docs](https://p0tion.super.site/).
```
baumstern, line 17: > A project might want to run their own trusted setup for a different param combination. what you think of that?

I agree with this @ctrlc03. But the purpose of p0tion needs to be explicitly defined. It's ambiguous whether p0tion serves as an archive page displaying the results of the ceremony, a page where one can view the ongoing status, or a tool designed for users to easily conduct their own ceremonies with desired parameters. This distinction should be made clear in this document.

docs/trustedsetup.md
ctrlc03: a set of zKeys here simply means 2+ zKeys, I could have used the word pair in this case probably (one for process messages and one for tally votes). Anyways, the maci maci ceremony is nown ended, so might as well change the wording of this doc and add links to the keys 😂

docs/trustedsetup.md
```diff
@@ -10,28 +10,28 @@ to manage a large amount of value, and if their users accept that the risk of
 coordinator misbehaviour is insufficient to justify doing the work nof a
 trusted setup. After all, MACI's security model presumes a trusted coordinator.
 
+The PSE team is making available a trusted set of zKeys for MACI's circuits,
```
samajammin, line 13:
```suggestion
The [PSE team](https://pse.dev/) is making available a trusted set of zKeys for MACI's circuits,
```
For those who don't know who "PSE" is...

samajammin (dismissed):
Suggested an added link but overall LGTM!

Side note, is there a way to preview changes to our doc pages? Would be nice to have preview deploys of that web app...

ctrlc03:
> Suggested an added link but overall LGTM!
> 
> Side note, is there a way to preview changes to our doc pages? Would be nice to have preview deploys of that web app...

committed that suggestion, thanks! sure, we could have a look, added as a comment on this issue so we can track there #729

docs/trustedsetup.md
baumstern: ahh I see XD

baumstern (approved):
LGTM

samajammin (approved):

PR #758, opened 2023-10-20
feat(genLocalState) - debugging utility for log fetching and local maci state generation

### LINKED ISSUE #447: Couldn't get a large amount of logs from Alchemy
Since [Alchemy has a limit to retrieve a logs]( https://docs.alchemy.com/alchemy/apis/ethereum/eth-getlogs), `genMaciStateFromContract()` might fails when a MACI instance has a large amount of logs; a single `eth_getLogs` call would fail when MACI instance has over 10k sign-ups or 10k publishMessages.

#### Possibly failing point
v1:
https://github.com/privacy-scaling-explorations/maci/blob/v1/contracts/ts/genMaciState.ts#L67-L85, https://github.com/privacy-scaling-explorations/maci/blob/v1/contracts/ts/genMaciState.ts#L235-L258

v0.10.x:
https://github.com/privacy-scaling-explorations/maci/blob/master/cli/ts/utils.ts#L63-L71

Its not a MACI-side bug, but think it would be a common pitfall and think we could have a better way to handle this (Note that sending a request of an interval of 2,000 blocks to Alchemy endpoint would fails with 429 Too Many Requests error).

### LINKED ISSUE #723: Missing MACI CLI, fetchLog, in v1.1.1
In v0.10, the command line `fetchLog` was introduced because `genProofs` would fail (in production) when retrieving event logs from too many blocks.

The `fetchLog` command can retrieve logs in batches which is needed to avoid errors from the node getting too many blocks. See the code here.

https://github.com/clrfund/maci/blob/maci-cli-0.10.2/cli/ts/fetchLogs.ts#L109

ctrlc03:
fix #723 
fix #447  

- [x] Fix bug of double fetching
- [x] Add fetching in blocks for generateMaciState
- [x] Implement serialization and deserialization for all DomainObjects as well as MaciState and Poll
- [x] In genLocalState save the serialized MaciState and Polls to a JSON file
- [x] in genProofs add the option to read the state from a json file -> deserialize the json file into a valid MaciState and Polls

How to test:

Run `tests/vanilla/testPreFetch.sh`


cli/ts/genProofs.ts
```diff
@@ -162,16 +162,21 @@ const configureSubparser = (subparsers: any) => {
...
-    // TODO: support resumable proof generation
-    //parser.addArgument(
-        //['-r', '--resume'],
-        //{
-            //action: 'storeTrue',
-            //help: 'Resume proof generation from the last proof in the specified output file',
-        //}
-    //)
+    parser.addArgument(
+        ['-eb', '--end-block'],
```
baumstern, line 174: do we need this?

baumstern:
Isn't https://github.com/privacy-scaling-explorations/maci/issues/723 a request to add the fetchLogs command back?

contracts/ts/genMaciState.ts
```diff
@@ -51,38 +53,70 @@ const genMaciStateFromContract = async (
...
+    let initLogs: any[] = []
+    let signUpLogs: any[] = []
+    let mergeStateAqSubRootsLogs: any[] = []
+    let mergeStateAqLogs: any[] = []
+    let deployPollLogs: any[] = []
+    
+    const lastBlock = endBlock ? endBlock : await provider.getBlockNumber()
+    
+    // Fetch event logs in batches
+    for (let i = fromBlock; i < lastBlock; i += blocksPerRequest + 1) {
```
baumstern, line 65: I believe it's worthwhile to implement test code to avoid issues like [the previous case of duplicate logs](https://github.com/privacy-scaling-explorations/maci/issues/723#issuecomment-1771599471). Please note that such duplicate logs resulted in generating an incorrect proof.

ctrlc03:
```
    // init() should only be called up to 1 time
    assert(
        initLogs.length <= 1, 
        'More than 1 init() event detected which should not be possible',
    )
```   
 If there are duplicate logs, this will throw.

cli/ts/genProofs.ts
ctrlc03: good shout, let me fix that

ctrlc03:
> Isn't #723 a request to add the

sure, but there was missing context about that, `fetchlogs` was replaced by this `genMaciStateFromContract` in v1. Only thing missing was the ability to fetch in batches

contracts/ts/genMaciState.ts
baumstern: There are signUpLogs, mergeStateAqSubRootsLogs, mergeStateAqLogs, deployPollLogs, publishMessageLogs...etc. Without adequate tests, confidently predicting the impact of these changes on production becomes challenging. What are your thoughts on this?

ctrlc03: this code incorporates the fix needed to ensure we do not fetch duplicate logs - the same fix that the project encountering the issue proposed + We do have e2e and integration tests which use this genProof and have been working correctly.

cli/ts/genProofs.ts
```diff
@@ -336,23 +360,34 @@ const genProofs = async (args: any) => {
...
-    if (txHash) {
-        const txn = await signer.provider.getTransaction(txHash);
-        fromBlock = txn.blockNumber
+    if (args.state_file) {
+        // @todo actually read the file first 
+        const content = JSON.parse(fs.readFileSync(args.state_file, 'utf-8').toString())
+        maciState = MaciState.fromJSON(content)
+        // ensure we merge all messages
+        maciState.polls.forEach((poll: Poll) => poll.mergeAllMessages()) 
+    } else {
```
baumstern, line 371: This change introduce to reading from a state file if `args.state_file` is provided, and then generating the proof based on the JSON content. However, there appears to be no test coverage to verify the correctness of the serialization/deserialization process and the subsequent proof generation using the parsed data. Could we introduce unit test that validate the correct functioning of the proof generation when using parsed data?

cli/ts/genProofs.ts
```diff
@@ -336,23 +360,34 @@ const genProofs = async (args: any) => {
...
-
-    // some rpc endpoint like bsc chain has limitation to retreive history logs
-    let fromBlock = 0
-    const txHash = args.transaction_hash
-    if (txHash) {
-        const txn = await signer.provider.getTransaction(txHash);
-        fromBlock = txn.blockNumber
+    if (args.state_file) {
+        // @todo actually read the file first 
+        const content = JSON.parse(fs.readFileSync(args.state_file, 'utf-8').toString())
```
baumstern, line 367: This assumes that the input file format is correct. It's important to handle potential exceptions that could arise from incorrect file formats. Adding error handling for the JSON parsing process could prevent runtime errors and improve the robustness of the script.

cli/ts/fetchLogs.ts
```diff
@@ -0,0 +1,226 @@
...
+    maciPrivkeyGroup.addArgument(
+        ['-dsk', '--prompt-for-maci-privkey'],
+        {
+            action: 'storeTrue',
+            help: 'Whether to prompt for your serialized MACI private key',
+        }
+    )
+
+    maciPrivkeyGroup.addArgument(
+        ['-sk', '--privkey'],
```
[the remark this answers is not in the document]
ctrlc03: this is needed to generate the maci state :)

cli/ts/fetchLogs.ts
```diff
@@ -0,0 +1,226 @@
...
+        maciAddress,
+        maciContractAbi,
+        provider,
+    )
+
+    const pollContract = new ethers.Contract(
+        pollAddress,
+        pollContractAbi,
+        provider
+    )
```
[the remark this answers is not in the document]
ctrlc03: removed thanks

ctrlc03:
> The `fetchLogs` command, as it's currently implemented, seems to be misnamed as it does more than just fetching logs—it generates and outputs the MACI state, which is an entirely different responsibility. This oversteps the expected functionality of a log-fetching utility, leading to unnecessary complexity and potential confusion:
> 
> * The command requires contract ABIs and the coordinator's private key, which are beyond the scope of simply fetching logs. These requirements suggest operations relevant to MACI state generation and proof creation, not log retrieval.
> * It performs operations with significant time complexity, such as array pushes and sorts on the entire log dataset, which are superfluous for log fetching; these operations incurs O(n) and O(n log n) time complexities.
> * The inclusion of the coordinator's private key raises security concerns, as it's not necessary for fetching logs and should not be requested without cause
> 
> This conflation of concerns could be misleading and might introduce inefficiencies or security vulnerabilities. A command should do one thing and do it well—fetching logs shouldn't involve generating state. Consider refactoring to separate these concerns, ensuring the command name and its functionality are in alignment, and that it adheres to the principle of least privilege by not requiring sensitive information where it's not needed.

Ok so the file now has a different name, `generateLocalState` to avoid confusion. It is not a 100% replacement for fetchLogs which was present in v0.x but it is a new utility for Maci v.1x - it provides support for fetching logs in chunks, sleep option for avoiding rate limiting of RPC providers, and the ability to save the full state in a nice and optimal JSON representation. This allows for better debugging, as it can be analyzed separately if any issues arise when generating proofs.

cli/tests/prepare_test.sh
```diff
@@ -51,6 +51,55 @@ clean() {
...
+    $MACI_CLI mergeMessages \
+        --poll-id $1
+    
+    echo "merge signups ..."
+    $MACI_CLI mergeSignups \
+        --poll-id $1
+
+    echo "gen proofs ..."
+
+    # generate the local state
```
baumstern, line 65: Nice demonstration of how to use new command!

cli/tests/prepare_test.sh
```diff
@@ -51,6 +51,55 @@ clean() {
...
+        $MACI_CLI genProofs \
+            --privkey macisk.49953af3585856f539d194b46c82f4ed54ec508fb9b882940cbe68bbc57e59e \
+            --poll-id $1 \
+            --process-wasm ./zkeys/ProcessMessages_"$PROCESS_MESSAGES_PARAMS"_js/ProcessMessages_"$PROCESS_MESSAGES_PARAMS".wasm \
+            --tally-wasm ./zkeys/TallyVotes_"$TALLY_VOTES_PARAMS"_js/TallyVotes_"$TALLY_VOTES_PARAMS".wasm \
+            --process-zkey "$ZKEYS_DIR"/ProcessMessages_"$PROCESS_MESSAGES_PARAMS".0.zkey \
+            --tally-zkey "$ZKEYS_DIR"/TallyVotes_"$TALLY_VOTES_PARAMS".0.zkey \
+            --tally-file tally.json \
+            --output proofs/ \
+            --state-file localState.json \
```
baumstern, line 84: Nice demonstration of how to use new command +2

baumstern:
Could you please add a description for the newly introduced command in the [cli.md file](https://github.com/privacy-scaling-explorations/maci/blob/dev/docs/cli.md#subcommands)? Once that's included, I think this PR will be ready for merging. Thanks for your hard work on this!

contracts/ts/genMaciState.ts
```diff
@@ -55,38 +59,81 @@ const genMaciStateFromContract = async (
...
+        if (sleepAmount) await sleep(sleepAmount)
+
+        signUpLogs = signUpLogs.concat(tmpSignUpLogs)
+    
+        const tmpMergeStateAqSubRootsLogs = await provider.getLogs({
+            ...maciContract.filters.MergeStateAqSubRoots(),
+            fromBlock: i,
+            toBlock,
+            address: address
+        })
```
yuetloo, line 102: Since we're querying the maci contract for multiple events, it might be more performant if we call getLogs() with multiple event topics, (e.g. don't need to sleep in between each batch of events). 

https://docs.ethers.org/v5/concepts/events/#events--logs-and-filtering

```
getLogs({
  address: addr,
  fromBlock: i,
  toBlock,
  topics: [
    [
      maci.filters.Init().topics?.[0],
      maci.filters.SignUp().topics?.[0],
      ...
    ]
  ]
})
```

core/ts/MaciState.ts
```diff
@@ -1334,6 +1330,82 @@ class Poll {
...
+    /**
+     * Serialize the Poll object to a JSON object
+     * @returns a JSON object
+     */
+    toJSON() {
+        return {
+            duration: this.duration,
+            pollEndTimestamp: this.pollEndTimestamp.toString(),
+            coordinatorKeypair: this.coordinatorKeypair.toJSON(),
+            treeDepths: this.treeDepths,
```
yuetloo, line 1343: Hmm.. I'd prefer not to write the coordinator keypair to the file.

cli/ts/genProofs.ts
```diff
@@ -380,23 +415,38 @@ const genProofs = async (args: any) => {
...
-    if (txHash) {
-        const txn = await signer.provider.getTransaction(txHash);
-        fromBlock = txn.blockNumber
+    if (args.state_file) {
+        const content = JSON.parse(fs.readFileSync(args.state_file, 'utf-8').toString())
+        try {
+            maciState = MaciState.fromJSON(content)
+        } catch (error: any) {
+            console.log('The provided state file is invalid. Please check that the correct path was provided and try again.')
+            return
```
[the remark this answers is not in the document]
ctrlc03: Definitely, there's an open PR with the refactoring of the CLI. In general, I've opted for throwing when there's an error and returning data (if any) when everything is successful, but open to suggestions. Returning data makes it very easy to test the cli properly

contracts/ts/genMaciState.ts
ctrlc03: Thanks for the suggestion, I'll take a look these days

core/ts/MaciState.ts
ctrlc03: Fair point, I think we can skip saving that on a file and have it passed as argument to the 'fromJSON' method

contracts/ts/genMaciState.ts
ctrlc03: @yuetloo Would you have an example? the topics seems to be related to a specific event, it seems I cannot add topics of different events in that array. However, I did remove the many sleeps in favour of a sleep at the end of each loop iteration to be more performant.

contracts/ts/genMaciState.ts
```diff
@@ -253,36 +277,66 @@ const genMaciStateFromContract = async (
...
+            toBlock,
+            address: pollContract.address
+        })
+        mergeMessageAqLogs = mergeMessageAqLogs.concat(tmpMergeMessageAqLogs)
 
-    const mergeMessageAqLogs = await provider.getLogs({
-        ...pollContract.filters.MergeMessageAq(),
-        fromBlock: fromBlock,
-    })
+        if (sleepAmount) await sleep(sleepAmount)
```
baumstern, line 338:
```suggestion
        if (sleepAmount) {
            await sleep(sleepAmount)
        }
```

nit: while concise, one-liners like this can compromise code readability and may lead to overlooked errors during future edits. It's generally a good practice to expand these into full if statements with braces for clarity.

baumstern (changes requested):
I've noticed some concerns with the `fromJson` method in handling `coordinatorPrivateKey`. 
1. Scope Issue: The method extends beyond its intended purpose by managing coordinatorPrivateKey, which is not typically within the scope of a fromJson function.
2. Incomplete Instance Risk: Separating the handling of coordinatorPrivateKey from fromJson could lead to the creation of incomplete MaciState instances.
 
Since this issue seems tangential to the main focus of our current PR, I suggest we address the `fromJson` functionality in a separate PR.

core/ts/MaciState.ts
```diff
@@ -1361,11 +1360,15 @@ class Poll {
      * @param maciState the reference to the MaciState Class
      * @returns a new Poll instance
      */
-    static fromJSON(json: any, maciState: MaciState): Poll {
+    static fromJSON(
+        json: any, 
+        maciState: MaciState, 
+        coordinatorPrivateKey: string
+    ): Poll {
```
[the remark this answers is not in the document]
ctrlc03: @baumstern  are you suggesting we reconstruct the object without a key (by passing a random one) and then set it after `fromJSON` finishes, and before doing any processing? this could work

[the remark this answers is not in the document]
ctrlc03: @baumstern reflected in the latest commits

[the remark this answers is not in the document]
baumstern: > @baumstern are you suggesting we reconstruct the object without the actual valid key (by passing a random one) and then set it after `fromJSON` finishes, and before doing any processing? this could work

Thank you for incorporating the feedback. While using `new Keypair()` in fromJSON and setting the `coordinatorKeypair` later does technically work, it introduces a requirement for users to be aware of an additional step (`setCoordinatorKeypair()`) to fully initialize the `Poll` object. This could potentially lead to confusion or errors if the user is not aware that the `Poll` instance is incomplete after `fromJSON`.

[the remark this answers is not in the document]
baumstern: I think fully addressing this issue is pretty tricky and I want to avoid prolonging this PR too much, as it is not main focus of this PR. We could handle this matter separately, possibly in a dedicated issue or follow-up PR. How would you like to proceed?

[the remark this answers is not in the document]
baumstern: Also..should we merge this PR first or wait until after merging the cli refactoring PR?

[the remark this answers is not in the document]
ctrlc03: @baumstern the only other solution I can think of, which could be much easier and clear, is to actually save the coordinator key in the serialized object - `toJSON` only converts to JSON, it does not save to disk. However, when used in the cli, before saving to disk, we remove the privateKey entry from the object, save it without, and while deserializing, we amend the json object created extracted from the local file with the actual key (which would be an input anyways to `genProofs`)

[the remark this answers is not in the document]
ctrlc03: > Also..should we merge this PR first or wait until after merging the cli refactoring PR?

I think we could merge this first while I work on the feedback of the CLI one, then incorporate this one command into the new cli. lmk what you think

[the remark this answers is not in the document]
baumstern: > @baumstern the only other solution I can think of, which could be much easier and clear, is to actually save the coordinator key in the serialized object - technically speaking `toJSON` only converts to JSON, it does not save to disk (it is up to the user to actually store it on disk later if they wish so). However, when used in the cli, before saving to disk, we remove the privateKey entry from the object, save it without, and while deserializing, we amend the json object created extracted from the local file with the actual key (which would be an input anyways to `genProofs`)

Sounds like a hacky solution haha. It's also technically make sense...but..since `toJSON` is kinda universal method so actually we can't prevent user to export their coordinator key (maybe accidently). We can remove coordinator key before write JSON into disk in our cli implementation but it is possible that third-party application that importing our package and exporting actual coordinator key into their disk. So I think it could be seen as a making attack space(maybe precisely making space that user could make mistake). Since coordinator key is definetely sensitive information, we should be careful and minimize the security risk.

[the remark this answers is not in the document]
baumstern: > > Also..should we merge this PR first or wait until after merging the cli refactoring PR?
> 
> I think we could merge this first while I work on the feedback of the CLI one, then incorporate this one command into the new cli. lmk what you think

Sounds good!

baumstern (approved):
LGTM!

PR #763, opened 2023-10-22
fix(cli) - fix wrong condition check in contracts deployment

### LINKED ISSUE #762: The 'create' cli fails if initialVoiceCreditProxy is passed as an argument
The follow line always evaluates to true because initialVoiceCredits is never undefined; it will be set to a default if the value is not provided from the command line.

https://github.com/privacy-scaling-explorations/maci/blob/master/cli/ts/create.ts#L77

ctrlc03:
fix #762


cli/ts/create.ts
```diff
@@ -67,20 +67,20 @@ const create = async (args: any) => {
...
         console.error('Error: only one of the following can be specified: the initial voice credit proxy or the amount of initial voice credits.')
         return 1
     }
 
-    let initialVoiceCreditProxyContractAddress
-    if (initialVoiceCreditProxy == undefined) {
+    let initialVoiceCreditProxyContractAddress: string 
+    if (initialVoiceCreditProxy === undefined) {
+        // check if we have the amount of credits to set, or use the default
+        const initialVoiceCredits = args.initial_voice_credits ? args.initial_voice_credits : DEFAULT_INITIAL_VOICE_CREDITS
```
baumstern, line 82: Keep in mind that `DEFAULT_INITIAL_VOICE_CREDITS` is being imported from`defaults.ts`. Even though it's intended to serve as a default value here, there's no guarantee that it will always be defined in the external file. It's worth ensuring that there's a safeguard or check in place to handle potential undefined values from external imports.

ctrlc03: if someone goes out of their way to remove the entry from `defaults.ts` then the code will not compile as create.ts it's importing something that does not exist

baumstern: yeah this isn't a major issue, but optionally, for the sake of robustness, you might consider adding a safeguard for `DEFAULT_INITIAL_VOICE_CREDITS`. Even if `defaults.ts` is present and being imported, there's no strict guarantee that `DEFAULT_INITIAL_VOICE_CREDITS` will be consistently defined within that file. A safeguard or a check could make the code more resilient and aligned with best practices of treating external dependencies with caution.

baumstern:
Re-triggered the failed e2e test. Once that passes, this PR should be good to go for merging.

ctrlc03:
> Re-triggered the failed e2e test. Once that passes, this PR should be good to go for merging.

I'm debugging the failure, so will probably fail again. will let you know once it's fixed.

baumstern (approved):
LGTM

PR #765, opened 2023-10-22
fix(contracts) - add onlyOwner to deployPoll

ctrlc03:


baumstern:
Great work! Out of curiosity, what are your thoughts on adding test cases for this chage?

ctrlc03:
> re your thoughts on adding test cases for this chage?

let's merge this first please -> https://github.com/privacy-scaling-explorations/maci/pull/747 which has a test case, then I'll add a unit test too in maci.test

baumstern (approved):
LGTM!

PR #766, opened 2023-10-23
fix(cli) - fix incosistencies on return codes

ctrlc03:


baumstern:
[Given that the calling functions don't check the return codes](https://github.com/privacy-scaling-explorations/maci/blob/dev/cli/ts/index.ts#L203-L243), standardizing them might not add value. For a cleaner approach, I suggest simply using return without any values.

baumstern:
Noticed that e2e integration test failed. Just trigerred to re-run..

ctrlc03:
> [Given that the calling functions don't check the return codes](https://github.com/privacy-scaling-explorations/maci/blob/dev/cli/ts/index.ts#L203-L243), standardizing them might not add value. For a cleaner approach, I suggest simply using return without any values.

yea fair point, I did think that too at first but refrained from changing too much.. if we both agree though I'll just remove the return code everywhere

baumstern (approved):
LGTM

PR #749, opened 2023-10-19
feat(refactor) - refactor the crypto package

ctrlc03:
Refactor the crypto package:
- [x] shift from one file to multiple organized files
- [x] add more typedoc docs
- [x] ensure generateRandomBabyJubValue does not throw but runs until a valid value is generated
- [x] separate types
- [x] separate tests and move supporting functions to utils
- [x] fix inconsistencies between BigInt and bigint types
- [x] fix various (repo-wide) ts bugs due to missing types declarations

PR #767, opened 2023-10-23
feat(domainobjs) - code and tests refactoring

ctrlc03:
depends on #749 (that should be merged first)

Refactoring of the domainobjs package:

- [x] separate classes into their own files
- [x] separate tests
- [x] typedoc 
- [x] BigInt -> bigint
- [x] optimizedmt d.ts declaration


domainobjs/ts/ballot.ts
```diff
@@ -0,0 +1,127 @@
...
+     * @param b - The ballot to compare with
+     * @returns whether the two ballots are equal
+     */
+    public equals(b: Ballot): boolean {
+        for (let i = 0; i < this.votes.length; i++) {
+            if (b.votes[i] !== this.votes[i]) {
+                return false
+            }
+        }
+        return b.nonce === this.nonce && this.votes.length === b.votes.length
```
0xmad, line 96:
```suggestion
       const isEqualVotes = this.votes.every((vote, index) => vote === b.votes[index])
       return isEqualVotes ?
         b.nonce === this.nonce && this.votes.length === b.votes.length : 
         false
```

domainobjs/ts/ballot.ts
```diff
@@ -0,0 +1,127 @@
...
+     * @returns a random Ballot
+     */
+    public static genRandomBallot(
+        _numVoteOptions: number,
+        _voteOptionTreeDepth: number
+    ) {
+        const ballot = new Ballot(_numVoteOptions, _voteOptionTreeDepth)
+        ballot.nonce = genRandomSalt() as bigint
+        return ballot;
+    }
```
0xmad, line 112:
```suggestion
    /**
     * Generate a random ballot
     * @param numVoteOptions How many vote options are available
     * @param voteOptionTreeDepth How deep is the merkle tree holding the vote options
     * @returns a random Ballot
     */
    public static genRandomBallot(
        numVoteOptions: number,
        voteOptionTreeDepth: number
    ) {
        const ballot = new Ballot(numVoteOptions, voteOptionTreeDepth)
        ballot.nonce = genRandomSalt() as bigint
        return ballot;
    }
```

domainobjs/ts/ballot.ts
```diff
@@ -0,0 +1,127 @@
...
+     * @param _voteOptionTreeDepth How deep is the merkle tree holding the vote options
+     * @returns a Blank Ballot object
+     */
+    public static genBlankBallot(
+        _numVoteOptions: number,
+        _voteOptionTreeDepth: number
+    ) {
+        const ballot = new Ballot(_numVoteOptions, _voteOptionTreeDepth)
+        return ballot
+    }
```
0xmad, line 126:
```suggestion
    /**
     * Generate a blank ballot
     * @param numVoteOptions How many vote options are available
     * @param voteOptionTreeDepth How deep is the merkle tree holding the vote options
     * @returns a Blank Ballot object
     */
    public static genBlankBallot(
        numVoteOptions: number,
        voteOptionTreeDepth: number
    ) {
        return new Ballot(numVoteOptions, voteOptionTreeDepth)
    }
```

domainobjs/ts/commands.ts
0xmad, line 1: This file need prettier run

domainobjs/ts/commands.ts
```diff
@@ -0,0 +1,275 @@
+import { decrypt, encrypt, genRandomSalt, hash4, sign, verifySignature, Signature, Ciphertext, EcdhSharedKey } from "maci-crypto"
+import * as assert from "assert"
+import { PubKey } from "./publicKey"
+import { PrivKey } from "./privateKey"
+import { Message } from "./message"
+
+/**
+ * @notice Base class for Commands
+ */
+export class Command {
```
0xmad, line 10: It can be interface or abstract class.

domainobjs/ts/message.ts
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+    /**
+     * Hash the message data and a public key
+     * @param _encPubKey the public key that is used to encrypt this message
+     * @returns the hash of the message data and the public key
+     */
+	public hash = (_encPubKey: PubKey): bigint => {
+		return hash13([...[this.msgType], ...this.data, ..._encPubKey.rawPubKey]) as bigint 
+	}
```
0xmad, line 58:
```suggestion
    /**
     * Hash the message data and a public key
     * @param encPubKey the public key that is used to encrypt this message
     * @returns the hash of the message data and the public key
     */
	public hash = (encPubKey: PubKey): bigint => {
		return hash13([...[this.msgType], ...this.data, ...encPubKey.rawPubKey]) as bigint 
	}
```

domainobjs/ts/message.ts
```diff
@@ -0,0 +1,92 @@
...
+			return false
+		}
+
+		for (let i = 0; i < this.data.length; i++) {
+			if (this.data[i] !== m.data[i]) {
+				return false
+			}
+		}
+
+		return true
```
0xmad, line 90:
```suggestion
		return this.data.every((data, index) => data === m.data[index])
```

domainobjs/ts/privateKey.ts
```diff
@@ -0,0 +1,71 @@
...
+
+		let validValue = false
+		try {
+			const value = BigInt('0x' + x)
+			validValue = value < SNARK_FIELD_SIZE
+		} catch {
+			// comment to make linter happy
+		}
+
+		return correctPrefix && validValue
```
0xmad, line 69:
```suggestion
		try {
	            const value = BigInt(`0x${x}`)
		    return correctPrefix && value < SNARK_FIELD_SIZE
		} catch {
                    return false;
		}
```

domainobjs/ts/publicKey.ts
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Return this public key as smart contract parameters
+     * @returns the public key as smart contract parameters
+     */
+	public asContractParam = (): any => {
+		return {
+			x: this.rawPubKey[0].toString(),
+			y: this.rawPubKey[1].toString(),
+		}
+	}
```
0xmad, line 47:
```suggestion
	public asContractParam = (): IContractParam => {
	        const [x, y] = this.rawPubKey;
		return {
			x: x.toString(),
			y: y.toString(),
		}
	}
```

domainobjs/ts/publicKey.ts
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Generate a serialized public key from this public key object
+     * @returns the string representation of a serialized public key
+     */
+	public serialize = (): string => {
+		// Blank leaves have pubkey [0, 0], which packPubKey does not support
+		if (
+			BigInt(`${this.rawPubKey[0]}`) === BigInt(0) &&
+			BigInt(`${this.rawPubKey[1]}`) === BigInt(0)
+		) {
```
0xmad, line 74:
```suggestion
                const { x, y } = this.asContractParam();
		if (
			BigInt(x) === BigInt(0) &&
			BigInt(y) === BigInt(0)
		) {
```

domainobjs/ts/publicKey.ts
```diff
@@ -0,0 +1,132 @@
...
+     * Check whether a serialized public key is serialized correctly
+     * @param s the serialized public key
+     * @returns whether the serialized public key is valid
+     */
+	public static isValidSerializedPubKey = (s: string): boolean => {
+		const correctPrefix = s.startsWith(SERIALIZED_PUB_KEY_PREFIX)
+
+		try {
+			PubKey.unserialize(s)
+            return correctPrefix && true
```
0xmad, line 127:
```suggestion
            return correctPrefix
```

domainobjs/ts/stateLeaf.ts
```diff
@@ -0,0 +1,155 @@
...
+     * @returns a deserialized state leaf
+     */
+    static unserialize = (serialized: string): StateLeaf => {
+        const base64 = serialized.replace(/-/g, '+').replace(/_/g, '/');
+        const j = JSON.parse(Buffer.from(base64, 'base64').toString('utf8'));
+
+        return new StateLeaf(
+            PubKey.unserialize(j[0]),
+            BigInt('0x' + j[1]),
+            BigInt('0x' + j[2])
```
0xmad, line 152:
```suggestion
            BigInt(`0x${j[1]}`),
            BigInt(`0x${j[2]}`)
```

domainobjs/ts/verifyingKey.ts
```diff
@@ -0,0 +1,159 @@
...
+     * @param _delta2 the delta2 point
+     * @param _ic the ic points
+     */
+	constructor(
+		_alpha1: G1Point,
+		_beta2: G2Point,
+		_gamma2: G2Point,
+		_delta2: G2Point,
+		_ic: G1Point[]
+	) {
```
0xmad, line 27: I think it's better to have one object param. Usually if there are more that 3 params or these params have the same time, it's a sign to use one object instead. Not sure if it has compatibility issues but if not, it will be useful to implement it.

0xmad (commented):
@ctrlc03 thanks, just left some suggestions and improvements.

domainobjs/ts/commands.ts
ctrlc03: good shout, amended

domainobjs/ts/ballot.ts
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    /**
+     * Generate an hash of this ballot
+     * @returns The hash of the ballot
+     */
+    public hash = (): bigint => {
+        const vals = this.asArray()
+        return hashLeftRight(vals[0], vals[1]) as bigint
```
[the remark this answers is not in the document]
ctrlc03: seems like it's complaining if I spread it directly: "A spread argument must either have a tuple type or be passed to a rest parameter."

domainobjs/ts/verifyingKey.ts
ctrlc03: What do you mean by "these params have the same time"? Is this pattern something you'd use in function signatures too, or only for classes constructors?

domainobjs/ts/ballot.ts
```diff
@@ -0,0 +1,127 @@
...
+     * @returns the ballot as a bigint array
+     */
+    public asArray = (): bigint[] => {
+        let lastIndexToInsert = this.votes.length - 1
+        while (lastIndexToInsert > 0) {
+            if (this.votes[lastIndexToInsert] !== BigInt(0)) {
+                break
+            }
+            lastIndexToInsert--;
+        }
```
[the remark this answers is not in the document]
ctrlc03: Loving your suggestions to make the code more concise, thank you! unfortunately as I rebased loads of these were outdated and GH wouldn't let me commit them from the web UI, so had to add them manually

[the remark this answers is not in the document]
ctrlc03: This one however will not work if the votes array is empty. For instance, when initializing a ballot with 0 vote options,  and running the asArray function, the lastIndex will be set to zero; the loop will run at least once because `i = 0; i <= lastIndexToInsert` -> and try to add an undefined value into the tree -> causing this `TypeError: Cannot convert undefined to a BigInt`
Currently, empty ballots are created here to generate the zero values for the incremental merkle trees: https://github.com/privacy-scaling-explorations/maci/blob/dev/contracts/ts/genEmptyBallotRootsContract.ts

What I did, in case the index is not found, we keep -1 so the loop does not run, and in any other cases it runs up until (and including) the last index to include

domainobjs/ts/ballot.ts
```diff
@@ -0,0 +1,127 @@
...
+        for (let i = 0; i < _numVoteOptions; i++) {
+            this.votes.push(BigInt(0))
+        }
+    }
+
+    /**
+     * Generate an hash of this ballot
+     * @returns The hash of the ballot
+     */
+    public hash = (): bigint => {
```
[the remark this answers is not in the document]
0xmad: I can do it automatically in terms of linter integration.

domainobjs/ts/ballot.ts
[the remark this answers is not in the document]
0xmad: Ok, let's leave it as-is.

domainobjs/ts/ballot.ts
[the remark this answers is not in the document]
ctrlc03: sick!

domainobjs/ts/verifyingKey.ts
0xmad: Sorry, I mean `type`. This works for functions and constructors as well.

domainobjs/ts/ballot.ts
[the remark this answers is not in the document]
0xmad: Good catch. So let's use this line instead:
```
const lastIndexToInsert = foundIndex < 0 ? foundIndex : lastIndex - foundIndex
```

domainobjs/ts/verifyingKey.ts
ctrlc03: Could you please share how you would rework this? would love to see this pattern in action and in other codebases too if you have any examples

0xmad: Sure.

```
interface IVerifyingKeyArgs {
  alpha1: G1Point;
  beta2: G2Point;
  gamma2: G2Point;
  delta2: G2Point;
  ic: G1Point[];
}

...

	constructor({
	  alpha1,
          beta2,
          gamma2,
          delta2,
          ic,
	}) {
          ...
        }
...
```

ctrlc03: I see, React props way 😃

ctrlc03: Let's keep this suggestion as a discussion item, and once we all agree on a way to go, we can update throughout the code?

0xmad: Ok, this works for me.

domainobjs/ts/commands.ts
```diff
@@ -0,0 +1,366 @@
...
+
+    constructor(cmdType: bigint) {
+        this.cmdType = cmdType;
+    }
+
+    public copy(): Command {
+        return this;
+    }
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    public equals(command: Command) {}
```
0xmad, line 31:
```suggestion
    abstract equals(command: Command): boolean;
```

domainobjs/ts/commands.ts
```diff
@@ -0,0 +1,366 @@
...
+            cmdType: this.cmdType.toString(),
+        };
+    }
+
+    /**
+     * Deserialize into a Command instance
+     * @param json
+     */
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    static fromJSON(json: any) {}
```
0xmad, line 47: Maybe it's better to get rid of static method in favour of instance one?

```suggestion
    abstract fromJSON(json: unknown): Command;
```

domainobjs/ts/message.ts
```diff
@@ -0,0 +1,115 @@
...
+            return false;
+        }
+
+        for (let i = 0; i < this.data.length; i++) {
+            if (this.data[i] !== m.data[i]) {
+                return false;
+            }
+        }
+
+        return true;
```
0xmad, line 94:
```suggestion
        return this.data.every((data, index) => data === m.data[index]);
```

domainobjs/ts/privateKey.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        let validValue = false;
+        try {
+            const value = BigInt(`0x${x}`);
+            validValue = value < SNARK_FIELD_SIZE;
+        } catch {
+            return false;
+        }
+
+        return correctPrefix && validValue;
```
0xmad, line 73:
```suggestion
        try {
            const value = BigInt(`0x${x}`);
            return correctPrefix && value < SNARK_FIELD_SIZE;
        } catch {
            return false;
        }
```

domainobjs/ts/__tests__/commands.test.ts
0xmad, line 1: Can you run prettier for this file to add new line at the end?

0xmad (commented):
@ctrlc03 thanks, just some missing comments.

domainobjs/ts/keyPair.ts
```diff
@@ -0,0 +1,89 @@
...
+            this.pubKey.rawPubKey[0] === keypair.pubKey.rawPubKey[0] &&
+            this.pubKey.rawPubKey[1] === keypair.pubKey.rawPubKey[1];
+
+        // If this assertion fails, something is very wrong and this function
+        // should not return anything
+        // XOR is equivalent to: (x && !y) || (!x && y )
+        const x = equalPrivKey && equalPubKey;
+        const y = !equalPrivKey && !equalPubKey;
+
+        assert((x && !y) || (!x && y));
```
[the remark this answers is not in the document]
0xmad: This one should work as bit equivalence (opposite result of XOR).
So this formula `([A && B] && !(!A && !B)) || (!(A && B) && (!A && !B))` can be simplified to this `A <-> B`.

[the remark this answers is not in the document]
ctrlc03: Typescript complains with: "The '^' operator is not allowed for boolean types. Consider using '!==' instead."

domainobjs/ts/commands.ts
ctrlc03: I'm not sure that it makes sense for `fromJSON` to be an instance method, as usually its objective is to generate a new instance of a class. 
Perhaps we get rid of the abstract class altogether? or do you see any benefits of keeping it here?

0xmad: Ok, so in this case I'd add `throw new Error("not implemented")` for this method to prevent any calls from abstract class.
And you can use interface instead of abstract class but interfaces don't have static members. I think some parent interface or class should be to not expose implementation details (like T or P command), so we can work only with `Command`.

0xmad: Ok, good. After this PR is merged I'll start adding linter config for this package.

0xmad (dismissed):
@ctrlc03 thanks!

ctrlc03:
> @ctrlc03 thanks!

thanks to you for the review and great suggestions, much appreciated

domainobjs/ts/types.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @notice An interface representing a zk-SNARK proof
+ */
+export interface Proof {
+	a: G1Point;
+	b: G2Point;
+	c: G1Point;
+}
+
+/**
+ * @notice An interface representing a MACI state lead
```
samajammin, line 14:
```suggestion
 * @notice An interface representing a MACI state leaf
```

domainobjs/ts/__tests__/verifyingKey.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import * as fs from "fs";
+import { VerifyingKey } from "../";
+import { expect } from "chai";
+
+describe("verifyingKey", () => {
+    it("Should convert a JSON file from snarkjs to a VerifyingKey", async () => {
+        const file = path.join(__dirname, "./artifacts/test_vk.json");
+        const j = fs.readFileSync(file).toString();
+        const d = JSON.parse(j);
+        const vk = VerifyingKey.fromJSON(j);
```
samajammin, line 11: Interesting that these seems to be the only place we use `VerifyingKey.fromJSON`... I wonder why it's needed 🤔

samajammin: I suppose many classes have a `fromJSON` method for deserialization

domainobjs/ts/verifyingKey.ts
```diff
@@ -0,0 +1,159 @@
...
+			copyG2(this.gamma2),
+			copyG2(this.delta2),
+			this.ic.map(
+				(c: any) => new G1Point(BigInt(c.x.toString()), BigInt(c.y.toString()))
+			)
+		)
+	}
+
+    /**
+     * Convert a json representation to a VerifyingKey
```
samajammin, line 124:
```suggestion
     * Deserialize into a VerifyingKey instance
```
Noticing in basically all other `fromJSON` methods, the comment is usually "Deserialize into a ___ instance" - would that also be appropriate here?

domainobjs/ts/ballot.ts
```diff
@@ -0,0 +1,156 @@
...
+    genRandomSalt,
+    hash5,
+    hashLeftRight,
+    IncrementalQuinTree,
+} from "maci-crypto";
+
+/**
+ * A Ballot represents a User's votes in a Poll, as well as their next valid
+ * nonce.
+ * @param _voiceCreditBalance The user's voice credit balance
```
samajammin, line 12: Noob question - what do params of a Class represent?

I figured it was describing the parameters that the constructor expects when creating an instance of the Ballot class, but this appears to not be the case

domainobjs/ts/keyPair.ts
```diff
@@ -0,0 +1,86 @@
+import * as assert from "assert";
+import { genEcdhSharedKey, genKeypair, genPubKey } from "maci-crypto";
+import { PrivKey } from "./privateKey";
+import { PubKey } from "./publicKey";
+
+/**
+ * @notice A KeyPair is a pair of public and private keys
+ * @author PSE
```
samajammin, line 8: These seems to be the only place we mention `@author PSE` - why?

domainobjs/ts/__tests__/DomainObjs.test.ts
```diff
@@ -1,286 +0,0 @@
...
-        });
-
-        it("the Keypair constructor should generate the correct public key given a private key", () => {
-            const rawKeyPair = genKeypair();
-            const k = new Keypair(new PrivKey(rawKeyPair.privKey));
-            expect(rawKeyPair.pubKey[0]).to.eq(k.pubKey.rawPubKey[0]);
-            expect(rawKeyPair.pubKey[1]).to.eq(k.pubKey.rawPubKey[1]);
-        });
-
-        it("PrivKey.serialize() and unserialize() should work correctly", () => {
```
samajammin, line 111: Looks like some tests (e.g. this one) is no longer present - why?

I thought the main goal here was separating the tests into independent files, not actually changing the tests. Is that right?

samajammin: Ah, I see - you also created `privateKey` & `publicKey` test files - nice 😄

domainobjs/ts/__tests__/DomainObjs.test.ts
```diff
@@ -1,286 +0,0 @@
...
-            expect(s.startsWith("macipk.")).to.be.true;
-
-            const d = s.slice(7);
-            const unpacked = unpackPubKey(Buffer.from(d, "hex"));
-
-            expect(unpacked[0].toString()).to.eq(pk1.rawPubKey[0].toString());
-            expect(unpacked[1].toString()).to.eq(pk1.rawPubKey[1].toString());
-        });
-
-        it("PrivKey.copy() should produce a deep copy", () => {
```
samajammin, line 162: This test looks to be missing from`privateKey.test.ts`

domainobjs/ts/__tests__/DomainObjs.test.ts
```diff
@@ -1,286 +0,0 @@
...
-            const pk3 = k1.pubKey;
-            const pk4 = pk3.copy();
-            expect(pk3.rawPubKey.toString()).to.eq(pk4.rawPubKey.toString());
-            pk4.rawPubKey = [BigInt(0)];
-            expect(pk3.rawPubKey.toString()).not.to.eq(
-                pk4.rawPubKey.toString()
-            );
-        });
-
-        it("Keypair.copy() should produce a deep copy", () => {
```
samajammin, line 206: Same here

domainobjs/ts/publicKey.ts
```diff
@@ -0,0 +1,155 @@
...
+    hashLeftRight,
+    packPubKey,
+    unpackPubKey,
+    PubKey as RawPubKey,
+} from "maci-crypto";
+import { SERIALIZED_PUB_KEY_PREFIX } from "./constants";
+
+/**
+ * @notice A class representing a public key
+ * @note This is a MACI public key, which is not to be
```
samajammin, line 13: `@note` vs `@notice`?

samajammin: Overall I like this explanation though (that MACI key != EVM key)! Would it make sense to add this comment to the `PrivKey` class as well?

samajammin: I think worth adding these explanations to the README as well (e.g. https://github.com/privacy-scaling-explorations/maci/tree/dev/domainobjs#pubkey)

domainobjs/ts/ballot.ts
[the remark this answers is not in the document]
samajammin: > This one however will not work if the votes array is empty

@ctrlc03 should this ever be the case? Why would we allow empty user ballots 🤔

samajammin (changes requested):
Nice work! A few questions & suggestions

domainobjs/ts/publicKey.ts
```diff
@@ -0,0 +1,155 @@
...
+ * confused with an EVM public key.
```
baumstern, line 14: What do you mean by that EVM public key? Technically EVM public key is not an established term

domainobjs/ts/ballot.ts
ctrlc03: I believe this is a repetition, I have removed them. Seems like the norm is to describe the parameters on the constructor

domainobjs/ts/__tests__/DomainObjs.test.ts
ctrlc03: added back, good catch

domainobjs/ts/publicKey.ts
ctrlc03: Adding to the other classes too and README

domainobjs/ts/publicKey.ts
ctrlc03: What would you use instead? this is your Ethereum public key, which works on any EVM chains

cli/ts/genLocalState.ts
```diff
@@ -1,5 +1,5 @@
 import * as ethers from "ethers";
-import fs from "fs";
+import * as fs from "fs";
```
0xmad, line 2: Why did you return this import?

domainobjs/ts/ballot.ts
```diff
@@ -0,0 +1,128 @@
+import assert = require("assert");
```
0xmad, line 1:
```suggestion
import assert from "assert";
```

domainobjs/ts/publicKey.ts
```diff
@@ -0,0 +1,143 @@
+import assert = require("assert");
```
0xmad, line 1:
```suggestion
import assert from "assert";
```

0xmad (commented):
@ctrlc03 thanks, just few comments about code style and imports

domainobjs/ts/ballot.ts
0xmad: These imports still have strange mixture of require and import.

domainobjs/ts/verifyingKey.ts
```diff
@@ -0,0 +1,142 @@
...
+   * @param _delta2 the delta2 point
+   * @param _ic the ic points
+   */
+  constructor(_alpha1: G1Point, _beta2: G2Point, _gamma2: G2Point, _delta2: G2Point, _ic: G1Point[]) {
+    this.alpha1 = _alpha1;
+    this.beta2 = _beta2;
+    this.gamma2 = _gamma2;
+    this.delta2 = _delta2;
+    this.ic = _ic;
+  }
```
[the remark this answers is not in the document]
0xmad: These changes are not applied

0xmad (commented):
@ctrlc03 seems changes haven't been applied. Is it something related to rebasing?

domainobjs/ts/publicKey.ts
samajammin: I'd say "Ethereum public key" might be best

domainobjs/ts/ballot.ts
samajammin: > These imports still have strange mixture of require and import.

Our convention should be to always use import, ya?

ctrlc03:
> @ctrlc03 seems changes haven't been applied. Is it something related to rebasing?

Yes, very sorry about that, I incorporated changes via the ui, then rebased locally and lost those 😅

samajammin (approved):
Good stuff, thanks for making those updates!

domainobjs/ts/publicKey.ts
baumstern: fyi: since there is an ECDSA public key using the secp256k1 curve, the Ethereum address of an EOA is generated by taking a keccak hash in 64-character hex format of the public key and dropping the first 24 characters. Note that the EVM doesn't directly store the public key, although they can be derived from the r, s, v fields in a transaction (https://ethereum.stackexchange.com/questions/13778/get-public-key-of-any-ethereum-account)

PR #777, opened 2023-10-24
feat(state-tree-depth) - ensure stateTreeDepth is not hardcoded 

ctrlc03:
We do not want to hardcode the stateTreeDepth in the smart contracts, but instead we want to pass them as arguments.

Changes:

* in the smart contract this is an immutable param thus passed in the constructor
* MaciState.ts and Poll.ts take them as constructor argument
* in tests we hardcode to 10 for the sake of testing
* contracts package.json's script `compileSol` takes an additional parameter so that we can compile the trees contracts with the correct params

### DIFF
--- a/circuits/ts/__tests__/utils.ts
+++ b/circuits/ts/__tests__/utils.ts
@@ -1,4 +1,4 @@
-const str2BigInt = (s: string): BigInt => {
+export const str2BigInt = (s: string): BigInt => {
     return BigInt(parseInt(
         Buffer.from(s).toString('hex'), 16
     ))
@@ -10,7 +10,7 @@ const str2BigInt = (s: string): BigInt => {
 // ffjavascript has no types so leave circuit with untyped
 type CircuitT = any;
 
-async function getSignal(circuit: CircuitT, witness: bigint[], name: string): Promise<bigint> {
+export const getSignal = async (circuit: CircuitT, witness: bigint[], name: string): Promise<bigint> => {
     const prefix = "main"
     // E.g. the full name of the signal "root" is "main.root"
     // You can look up the signal names using `circuit.getDecoratedOutput(witness))`
@@ -26,7 +26,4 @@ async function getSignal(circuit: CircuitT, witness: bigint[], name: string): Pr
     return BigInt(witness[indexInWitness]);
 }
 
-export {
-    str2BigInt,
-    getSignal
-}
+export const STATE_TREE_DEPTH = 10
--- a/contracts/contracts/MACI.sol
+++ b/contracts/contracts/MACI.sol
@@ -22,7 +22,8 @@ import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
 contract MACI is IMACI, DomainObjs, Params, SnarkCommon, Ownable {
     // The state tree depth is fixed. As such it should be as large as feasible
     // so that there can be as many users as possible.  i.e. 5 ** 10 = 9765625
-    uint8 public constant override stateTreeDepth = 10;
+    // this should also match the parameter of the circom circuits.
+    uint8 public immutable stateTreeDepth;
 
     // IMPORTANT: remember to change the ballot tree depth
     // in contracts/ts/genEmptyBallotRootsContract.ts file
@@ -128,7 +129,8 @@ contract MACI is IMACI, DomainObjs, Params, SnarkCommon, Ownable {
     constructor(
         PollFactory _pollFactory,
         SignUpGatekeeper _signUpGatekeeper,
-        InitialVoiceCreditProxy _initialVoiceCreditProxy
+        InitialVoiceCreditProxy _initialVoiceCreditProxy,
+        uint8 _stateTreeDepth
     ) {
         // Deploy the state AccQueue
         stateAq = new AccQueueQuinaryBlankSl(STATE_TREE_SUBDEPTH);
@@ -137,6 +139,7 @@ contract MACI is IMACI, DomainObjs, Params, SnarkCommon, Ownable {
         pollFactory = _pollFactory;
         signUpGatekeeper = _signUpGatekeeper;
         initialVoiceCreditProxy = _initialVoiceCreditProxy;
+        stateTreeDepth = _stateTreeDepth;
 
         signUpTimestamp = block.timestamp;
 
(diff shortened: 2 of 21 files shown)


contracts/contracts/MACI.sol
```diff
@@ -129,6 +131,7 @@ contract MACI is IMACI, DomainObjs, Params, SnarkCommon, Ownable {
         pollFactory = _pollFactory;
         signUpGatekeeper = _signUpGatekeeper;
         initialVoiceCreditProxy = _initialVoiceCreditProxy;
+        stateTreeDepth = _stateTreeDepth;
```
samajammin, line 134: Is there any danger in this value now being an arbitrary size? i.e. should there be a check against some max value?

contracts/ts/__tests__/MACI.test.ts
```diff
@@ -113,12 +112,15 @@ describe('MACI', () => {
...
     let pollId: number
+    let user: ethers.Signer
 
     describe('Deployment', () => {
         beforeAll(async () => {
             signer = await getDefaultSigner()
             const r = await deployTestContracts(
                 initialVoiceCreditBalance,
+                STATE_TREE_DEPTH,
+                true
```
samajammin, line 123: Why pass `true` here?

ctrlc03: oh basically those utils functions like deployTestContract or deployContract have an optional parameter called "quiet", passing true so it doesn't print unnecessary logs while running the tests

contracts/contracts/MACI.sol
ctrlc03: Yes and no. If they put a value that does not match the circuit, they would basically need to start over (there might be some workarounds though, still not ideal).

This is a uint8 so max value would be 255, having said that, a tree depth of 255 would be incredibly big, considering 10 allows for a little bit more than 1M total signups. 
My view here is the following:
1. if we put a max value, what should this be?
2. In the past, we asked the users to change a constant variable rather than passing it as a constructor parameter:
   - I consider these options kinda the same, we never really validated that this constant variable was > than a max value
   - people running MACI should be (and are to some extent) informed that this parameter should match the one in the circuits, hence it is their responsibility of setting it up correctly.
3. even if this value was < than a max value we decide on, it could still be different than what they put in the circuit and in the other components of MACI. 

To summarize, I think the best we can do here in terms of "validation" is to properly inform that this value should match whichever value is in the Circom circuits.

baumstern:
To confirm, we don't have to follow the steps described in [this doc](https://github.com/privacy-scaling-explorations/maci/wiki/Follow-ups-when-changing-%60stateTreeDepth%60 ) anymore?

ctrlc03:
> To confirm, we don't have to follow the steps described in [this doc](https://github.com/privacy-scaling-explorations/maci/wiki/Follow-ups-when-changing-%60stateTreeDepth%60) anymore?

correct, I would say we probably need to improve documentation on this change - for now I updated the scripts in the contracts folder, as well as the README. What do you think, should we updated the cli docs too to ensure the new flag ("-s") is more evident? would be curious to hear your opinion after you review this PR and how it worked for you testing it

circuits/ts/__tests__/utils.ts
```diff
@@ -26,7 +26,4 @@ async function getSignal(circuit: CircuitT, witness: bigint[], name: string): Pr
     return BigInt(witness[indexInWitness]);
 }
 
-export {
-    str2BigInt,
-    getSignal
-}
+export const STATE_TREE_DEPTH = 10
```
baumstern, line 29: Is there a specific reason this is here? Perhaps it could fit well in `constants.ts`

crypto/ts/__tests__/AccQueue.test.ts
```diff
@@ -346,7 +346,6 @@ describe('AccQueue', () => {
                 })
 
                 it('Should fail if there are 0 leaves', () => {
-                    expect.assertions(1)
```
baumstern, line 349: Just curious, why was this removed?

crypto/ts/__tests__/Crypto.test.ts
```diff
@@ -111,7 +111,6 @@ describe('Cryptographic operations', () => {
         })
 
         it('The ciphertext should differ from the plaintext', () => {
-            expect.assertions(plaintext.length)
```
baumstern, line 114: Just curious, why was this removed?

baumstern:
> > To confirm, we don't have to follow the steps described in [this doc](https://github.com/privacy-scaling-explorations/maci/wiki/Follow-ups-when-changing-%60stateTreeDepth%60) anymore?
> 
> correct, I would say we probably need to improve documentation on this change - for now I updated the scripts in the contracts folder, as well as the README. What do you think, should we updated the cli docs too to ensure the new flag ("-s") is more evident? would be curious to hear your opinion after you review this PR and how it worked for you testing it

It’s an essential value, so it should be as visible as possible. However, people often don’t read the documentation thoroughly..so we might need to seek a somewhat better way to highlight it. Although a good idea doesn’t come to mind immediately.

circuits/ts/__tests__/utils.ts
ctrlc03: totally

crypto/ts/__tests__/AccQueue.test.ts
ctrlc03: I'm not really sure these are needed tbf. You can use expect.assertions when there might be cases of certain assertions are not called, and you want to constraint it to actually check those run. In this case it seems random to me.. For instance:

```ts 
   expect.assertions(plaintext.length)
   for (let i = 0; i < plaintext.length; i++) {
                expect(plaintext[i] !== ciphertext[i + 1]).toBeTruthy()
    }
```
we are looping for plaintext.length iterations, so it's kinda obvious that it will call expect for plaintext.length times

On the other hand, if you had this (very simple and dumb example): 

```ts
 const data = fetchData();
  if (data) {
        expect(data).toBe('expected data');
 }
```
here it would make sense to add expect.assertions(1) -> if you are running the test cases assuming that `data` should always be defined

crypto/ts/__tests__/Crypto.test.ts
ctrlc03: read above please

crypto/ts/__tests__/AccQueue.test.ts
baumstern: thanks for the explaination! Well as you said there is no meaning of using `expect.assertion()` in this code block.

baumstern: Anyway we don't use jest anymore so it doesn't matter at all

baumstern:
Just a reminder that there is a merge conflict that needs to be addressed. Once that's resolved, it looks like we're all set to merge this PR into the dev branch. Great work!

ctrlc03:
> Just a reminder that there is a merge conflict that needs to be addressed. Once that's resolved, it looks like we're all set to merge this PR into the dev branch. Great work!

merge conflicts sorted!  thanks for the review

baumstern (approved):
LGTM

Issue #803, opened 2023-11-02
Set up nightly build for proving key


baumstern:
Establish a nightly build to validate proving key generation amidst potential external dependencies changes. This practice aims to catch common issues such as breaking changes.

samajammin:
@baumstern mind explaining what this is? Not sure what this means.

baumstern:
> @baumstern mind explaining what this is? Not sure what this means.

Updated the description.

PR #775, opened 2023-10-23
fix(ci) - remove unsupported node versions from workflows and speed things up

ctrlc03:
- [x] generate small zkeys for this branch to use in the e2e and integration tests
- [x] reduce run time by not compiling zkeys every time
- [x] fix integration tests

PR #804, opened 2023-11-02
ci: nightly proving key generation workflow

baumstern:
This PR introduces a nightly proving key generation workflow. It will help us maintain the ability to generate the proving key and ensure that it remains functional even if no new changes are made to the circuits. It provides an additional layer of assurance for the codebase and allows others to build their own proving keys without any issues.

### DIFF
--- a/.github/workflows/circuit-build.yml
+++ b/.github/workflows/circuit-build.yml
@@ -2,8 +2,9 @@ name: Circuit
 
 on:
   push:
-    branches: [ master, audit, dev ]
+    branches: [ master, dev ]
   pull_request:
+
 
 concurrency: 
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
@@ -12,18 +13,15 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-22.04
-    strategy:
-      fail-fast: false
-      matrix:
-        node-version: [18]
+
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js 20
         uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 20
           cache: 'npm'
 
       - name: Update npm to latest version
--- a/.github/workflows/circuit-build.yml
+++ b/.github/workflows/circuit-build.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Use Node.js 20
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: 20
           cache: 'npm'
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -6,7 +6,7 @@ on:
 
 
 jobs:
-  generate-proving-key:
+  generate-proving-keys:
     runs-on: ubuntu-22.04
 
     steps:
@@ -28,17 +28,6 @@ jobs:
             libsodium-dev \
             nasm \
             nlohmann-json3-dev
-
-      # To prevent `npm install` failure of circuit package, 
-      # it has to checkout manually because version of `circomlib` has pinned to a specific commit
-      - name: Checkout circomlib
-        run: |
-          git init circuits/node_modules/circomlib
-          cd circuits/node_modules/circomlib
-          git remote add origin https://github.com/weijiekoh/circomlib
-          git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin ac85e82c1914d47789e2032fb11ceb2cfdd38a2b
-          git checkout --progress --force ac85e82c1914d47789e2032fb11ceb2cfdd38a2b
-          rm -rf ./.git
 
       - name: Initialize Project
         run: |
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,125 @@
+name: Nightly
+
+on:
+  schedule:
+    - cron: 0 0 * * *
+
+
+jobs:
+  generate-proving-key:
+    runs-on: ubuntu-22.04
+
+    steps:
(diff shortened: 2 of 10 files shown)


samajammin:
@baumstern I assume this is aimed to solve https://github.com/privacy-scaling-explorations/maci/issues/803 ?

Moving forward, please provide PR descriptions & reference the relevant issue

samajammin:
FYI @baumstern looks like this PR includes a bunch of miscellaneous commits. Why? Perhaps you didn't branch of the latest `dev`?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,172 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: master
+
+      - name: Use Node.js 16
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
samajammin, line 21: I believe we should use Node 18 now, correct? See https://github.com/privacy-scaling-explorations/maci/issues/775

baumstern: Yes for unit tests. But e2e test has problem with Node 18. See https://github.com/privacy-scaling-explorations/maci/pull/775#discussion_r1374742715

baumstern:
> FYI @baumstern looks like this PR includes a bunch of miscellaneous commits. Why? Perhaps you didn't branch of the latest `dev`?

The reason for this is that the PR was created from the previous default branch, which was the `master`. [GitHub Actions schedule triggers only work on the default branch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule). But the default branch has now been changed to the `dev` so things are sorted out well now.

.github/workflows/circuit-build.yml
```diff
@@ -2,28 +2,26 @@ name: Circuit
 
 on:
   push:
-    branches: [ master, audit, dev ]
+    branches: [ master, dev ]
```
ctrlc03, line 5: good catch there

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,125 @@
+name: Nightly
+
+on:
+  schedule:
+    - cron: 0 0 * * *
+
+
+jobs:
+  generate-proving-key:
```
ctrlc03, line 9: small nit: would call it "generate-proving-keys" as it's > 1 key

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,125 @@
...
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes \
+            build-essential \
+            libgmp-dev \
+            libsodium-dev \
+            nasm \
+            nlohmann-json3-dev
```
ctrlc03, line 30: nlohmann-json3-dev is not needed anymore

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,125 @@
...
+
+      # To prevent `npm install` failure of circuit package, 
+      # it has to checkout manually because version of `circomlib` has pinned to a specific commit
+      - name: Checkout circomlib
```
ctrlc03, line 34: how come this is not something we do while working locally but it's needed here?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,125 @@
...
+          sudo apt-get update
+          sudo apt-get install --yes \
+            build-essential \
+            libgmp-dev \
+            libsodium-dev \
+            nasm
+
+      # # To prevent `npm install` failure of circuit package, 
+      # # it has to checkout manually because version of `circomlib` has pinned to a specific commit
+      # - name: Checkout circomlib
```
samajammin, line 34: Why is this now commented out?

If not needed, can we delete this?

.github/workflows/nightly.yml
baumstern: It appears that removing nlohmann-json3-dev leads to a compilation failure. Here's the error we encounter:

```
main.cpp:9:10: fatal error: nlohmann/json.hpp: No such file or directory
    9 | #include <nlohmann/json.hpp>
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

.github/workflows/nightly.yml
baumstern: he step to manually checkout circomlib was introduced in April 2022, as part of a pull request (see [PR #403](https://github.com/privacy-scaling-explorations/maci/pull/403/files)). The reason for this step was due to a specific requirement at the time. maci-circuits imports circomlib and pins it to a specific commit (as indicated [here in the package.json](https://github.com/privacy-scaling-explorations/maci/blob/dev/circuits/package.json#L25)). Back then, the direct pulling of this dependency with a pinned commit like https://github.com/weijiekoh/circomlib#ac85e82c1914d47789e2032fb11ceb2cfdd38a2b failed.

This issue could have been related to the behavior of npm.js or changes in GitHub's security policy, as outlined in this [GitHub blog post](https://github.blog/2021-09-01-improving-git-protocol-security-github/) or specific to the GitHub Actions Runner environment.

However, I've re-triggered the CI run without this manual checkout step and it worked smoothly, indicating that this step may no longer be necessary. It seems that the underlying issue has been resolved, either through updates to npm.js or GitHub's handling of dependencies.

.github/workflows/nightly.yml
ctrlc03: mm I want to do some checks on a fresh VM actually. Looks like rapidnsark does not need this, but zkey-manager does, however where does it use it? is it here: https://github.com/privacy-scaling-explorations/zkey-manager/blob/master/ts/compile.ts#L102?

baumstern: It's needed when computing witness with c++: https://docs.circom.io/getting-started/computing-the-witness/

.github/workflows/nightly.yml
baumstern: reflected in 20676ac36afb18c8d763b05e6d680aec52f0c518

.github/workflows/nightly.yml
baumstern: Removed in 20676ac36afb18c8d763b05e6d680aec52f0c518

PR #817, opened 2023-11-07
feat(cross-platform-maci) - Allow MACI to run e2e on any architecture

### LINKED ISSUE #814: cross-platform MACI
Currently, it is not possible to fully use MACI (e2e process) on a non-intel machine. This is due to the use of rapidsnark which uses Intel assembly for faster zk-SNARK proof generation. 

To solve this, we could check on which architecture we are running , if running on ARM, then use Snarkjs with the wasm file to generate a proof.

Steps:

- [x] add fullProve declaration to snarkjs.d.ts
- [x] change genProof function on circuits ts code
- [x] add function to check for architecture
- [x] change genProofs to accept new arguments and new validation + new genProof function signature
- [x] change test scripts to check the system we are running on and pass the correct args
- [x] test it all

ctrlc03:
modify proof generation code and related tests to ensure MACI can be run on any architecture

fix #814 

- [x] add fullProve declaration to snarkjs.d.ts
- [x] change genProof function on circuits ts code
- [x] add function to check for architecture
- [x] change genProofs to accept new arguments and new validation + new genProof function signature
- [x] change test scripts to check the system we are running on and pass the correct args
- [x] test it all

### DIFF
--- a/circuits/scripts/downloadTestCircuits.sh
+++ b/circuits/scripts/downloadTestCircuits.sh
@@ -7,6 +7,6 @@ mkdir -p build/test/
 cd build/test
 
 wget --quiet -O circuits.tar.gz \
-    https://macitestcircuits.blob.core.windows.net/circom2/macitestcircuits.tar.gz
+    https://maci-develop-fra.s3.eu-central-1.amazonaws.com/v1.1.2/maci_keys_10-2-1-2_dev.tar.gz
 tar xf circuits.tar.gz
 rm circuits.tar.gz
--- a/circuits/ts/index.ts
+++ b/circuits/ts/index.ts
@@ -2,6 +2,7 @@ import * as fs from 'fs'
 import * as path from 'path'
 import * as shelljs from 'shelljs'
 import * as tmp from 'tmp'
+import * as os from 'os'
 import { zKey, groth16 } from 'snarkjs'
 
 import { stringifyBigInts } from 'maci-crypto'
@@ -23,13 +24,20 @@ const cleanThreads = async () => {
     return promises
 }
 
-const genProof = (
+const genProof = async (
     inputs: string[],
-    rapidsnarkExePath: string,
-    witnessExePath: string,
     zkeyPath: string,
+    rapidsnarkExePath?: string,
+    witnessExePath?: string,
+    wasmPath?: string,
     silent = true,
-): any => {
+): Promise<any> => {
+    // if we are running on an arm chip we can use snarkjs directly
+    if (isArm()) {
+        const { proof, publicSignals } = await groth16.fullProve(inputs, wasmPath, zkeyPath)
+        return { proof, publicSignals }
+    }
+    // intel chip flow (use rapidnsark)
     // Create tmp directory
     const tmpObj = tmp.dirSync()
     const tmpDirPath = tmpObj.name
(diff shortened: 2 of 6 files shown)


circuits/scripts/downloadTestCircuits.sh
```diff
@@ -7,6 +7,6 @@ mkdir -p build/test/
 cd build/test
 
 wget --quiet -O circuits.tar.gz \
-    https://macitestcircuits.blob.core.windows.net/circom2/macitestcircuits.tar.gz
+    https://maci-develop-fra.s3.eu-central-1.amazonaws.com/v1.1.2/maci_keys_10-2-1-2_dev.tar.gz
```
samajammin, line 10: Why do test circuits change?

circuits/ts/index.ts
```diff
@@ -23,13 +24,20 @@ const cleanThreads = async () => {
...
     inputs: string[],
-    rapidsnarkExePath: string,
-    witnessExePath: string,
     zkeyPath: string,
+    rapidsnarkExePath?: string,
+    witnessExePath?: string,
+    wasmPath?: string,
     silent = true,
-): any => {
+): Promise<any> => {
```
samajammin, line 34: Why do we return `any` here? Can we not be sure of the interface of the return values?

samajammin:
@ctrlc03 could you explain what tests/scripts/etc. I should run to test that this all works? Or should a simple `npm run test` from within the /circuits/ directory cover all this? I'm not seeing any test that actually executes `genProof` 🧐

circuits/scripts/downloadTestCircuits.sh
ctrlc03: tbf I think we can remove this altogether as we have the same script in integrationTests which is what we actually use. This script is not used and the link was outdated (pre v1.1.1)

circuits/ts/index.ts
ctrlc03: yes I think we can sort all that out when doing the actual refactor

ctrlc03:
> @ctrlc03 could you explain what tests/scripts/etc. I should run to test that this all works? Or should a simple `npm run test` from within the /circuits/ directory cover all this? I'm not seeing any test that actually executes `genProof` 🧐
> 
> I was able to successfully:
> 
> * `npm run build`
> * `npm run test`

for now, tests are the e2e tests written in bash. You will have to navigate to the contracts folder, run:

```bash
npm run compileSol
npm run hardhat &
cd ../cli
bash tests/vanilla/test1.sh
```

samajammin:
Cool, running
```
bash tests/vanilla/test1.sh
```
Passes!

samajammin (approved):

circuits/ts/index.ts
```diff
@@ -95,4 +103,12 @@ const extractVk = async (zkeyPath: string) => {
     return vk
 }
 
+/**
+ * Check if we are running on an arm chip
+ * @returns whether we are running on an arm chip
+ */
+const isArm = (): boolean => {
```
baumstern, line 110: nit: Since WASM is platform-independent, there is no need to specifically check for the arm architecture. It's worth considering that there are other platform architectures, such as RISC-V (riscv64)

PR #819, opened 2023-11-08
deps(testing) - consolidate testing frameworks across the project

ctrlc03:
This PR removes usage of JEST in favour of Mocha + Chai -> given we have multiple packages in this monorepo and both circuits and contracts tests use mocha (circom_tester and hardhat) 

- [x] remove jest @types/jest 
- [x] install mocha ts-mocha @types/mocha chai @types/chai
- [x] fix all tests
- [x] change test scripts
- [x] bonus: run all unit tests in one command from the root of the monorepo: npm run test from monorepo root will run all unit tests


package.json
```diff
@@ -8,9 +8,10 @@
     "bootstrap": "lerna bootstrap",
     "build": "lerna run build",
     "lint": "eslint './**/**/*.ts'",
-    "lint-fix": "eslint './**/**/*.ts' --fix"
+    "lint-fix": "eslint './**/**/*.ts' --fix",
+    "test": "lerna run test --ignore maci-common --ignore maci-server --ignore maci-integrationtests --ignore maci-cli"
```
samajammin, line 12: Why ignore `maci-cli`? Is that because you'll plan to refactor that as part of #794?

samajammin: I also wonder how intuitive it is to ignore integration tests 🤔 but I think it makes sense

.github/workflows/reusable-e2e.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         id: check_changes
         run: |
           CHANGED_FILES=${{ steps.get-changed-files.outputs.all }}
-          if echo "$CHANGED_FILES" | grep -q "circuit" && ! echo "$CHANGED_FILES" | grep -q ".github/workflows/circuit-build.yml"; then
+          if echo "$CHANGED_FILES" | grep -q "\.circom"; then
```
samajammin, line 30: What's this change about?

package.json
ctrlc03: Basically here I'm trying to run only in unit tests, hence why cli and integration tests are ignored

.github/workflows/reusable-e2e.yml
ctrlc03: Oh yea sorry for not mentioning in the description! Basically this change is so that the e2e test only run when the actual circom files change, not any files in the circuits folder as it was until this commit. I was planning to introduce this with another PR but got carried away by not wanting to wait on circuits compilation for this PR

README.md
```diff
@@ -108,17 +108,26 @@ own unit tests.
...
+```
+
+Our you can run unit tests within each submodule. for example to run the `crypto` tests:
 
 ```bash
 cd crypto
 npm run test
 ```
 
+<!-- TODO: confirm if true -->
```
samajammin, line 130: @ctrlc03 is this necessary? Seems I'm able to run the full suite e.g. of `contracts` tests without nonce errors...

README.md
```diff
@@ -108,17 +108,26 @@ own unit tests.
 
 ### Unit tests
 
+Unit tests within the project are built using [Mocha](https://mochajs.org/) and [Chai](https://www.chaijs.com/). Mocha is a test framework that provides the environment to write and run JavaScript tests, while Chai is an assertion library that allows us to write assertions in a more expressive and readable way.
+
+
 The following submodules contain unit tests: `core`, `crypto`, `circuits`,
 `contracts`, and `domainobjs`.
 
-Except for the `contracts` submodule, run unit tests as such (the following
```
samajammin, line 114: > Except for the `contracts` submodule

This can be removed right @ctrlc03?

README.md
samajammin: > First, start a Hardhat instance in a separate terminal:
> ```bash
> cd contracts
> npm run hardhat
> ```

Is this needed? I'm able to just run `npm run test` within `contracts` & it seems to automatically start up Hardhat

circuits/jest.config.js
```diff
@@ -1,29 +0,0 @@
-module.exports = {
```
samajammin, line 1: So no Mocha/Chai config files needed?

samajammin (changes requested):
Looks great! A couple small comments/questions. I added a small update to the README, could please check my work? Feel free to expand on that testing section of the README to cover these changes / update anything that's outdated.

README.md
ctrlc03: yea that's how it's supposed to work (hardhat does start itself) so yea feel free to remove. In general most of these README are outdated (wondering if it might better to address all in one PR?)

circuits/jest.config.js
ctrlc03: nope it just works (for now) :)

README.md
ctrlc03: correct!

README.md
samajammin: > In general most of these README are outdated (wondering if it might better to address all in one PR?)

I'm _alright_ with that, though generally I think best to update documentation at the same time as the code is updated (ideally in the same PR), because:
- that tends to make it easier for PR creator, given the specific info & context is fresh in their mind
- it also makes it easier for the reviewer, so they can easily understand what's changing
- it avoids the issue of having this big chore for someone to pick up at the end of this refactoring effort. Not to mention if the chore drops as a priority, then everything is a mess 😄

package.json
samajammin: > Running them in this command would involve first spawning a local hardhat node

But doesn't running contract tests also spin up a hardhat node?

samajammin (approved):
🚀

package.json
ctrlc03: It does, however because we run them using hardhat directly (hardhat test x.test.ts) there is no need to pre spawn a node, it just does it automatically and kills it once done

README.md
ctrlc03: Noted! I'll make sure from now on all doc updates are included in the PR 🙂

baumstern:
Great work!

PR #821, opened 2023-11-10
Add zkeys from 'download-zkeys' script to gitignore

samajammin:
Noticed after running:
```
cd integrationTest
npm run download-zkeys
```

That the artifacts are picked up in git, which I assume we don't want/need:

```
~/code/pse/maci on  dev [$?] via ⬢ v18.18.2
➜ gs
On branch dev
Your branch is up to date with 'origin/dev'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
    cli/maci_keys.tar.gz

nothing added to commit but untracked files present (use "git add" to track)
```


ctrlc03:
> cli/maci_keys.tar.gz

would it also work if we directly delete the archive file after it's decompressed (to also help them save disk space)? https://github.com/privacy-scaling-explorations/maci/blob/dev/integrationTests/scripts/download_zkeys.sh -> i.e. add `rm cli/maci_keys.tar.gz`

baumstern (approved):

ctrlc03 (approved):
Posted a comment (https://github.com/privacy-scaling-explorations/maci/pull/821#issuecomment-1814453708) otherwise LGTM

PR #824, opened 2023-11-11
Add clean script

samajammin:
Simple script to flush out /build & /node_modules for all submodules. I've found this useful with the various refactoring work we're doing to get clean installs of dependencies


samajammin:
Noting this does not remove the /node_modules at the root of the project, so you'd still want to run `rm -rf node_modules` from root to fully clean out all /node_modules. Not sure if we should just add that as part of this command?

baumstern (approved):
LGTM

PR #837, opened 2023-11-25
docs: add 1.0 spec

baumstern:
This PR add MACI 1.0 spec to website

### DIFF
--- a/website/versioned_docs/version-v1.x/spec.md
+++ b/website/versioned_docs/version-v1.x/spec.md
@@ -1,15 +1,19 @@
 ---
-id: spec
-title: MACI 1.0 Specification
+title: MACI v1.0 Specification
 description: A detailed specification meant to assist auditors in reviewing MACI version 1.0
 sidebar_label: Specification
 sidebar_position: 14
-
 ---
 
-# MACI 1.0 Specification (for audit)
+# MACI v1.0 Specification
 
-This is a detailed specification meant to assist auditors in reviewing MACI version 1.0.
+:::info
+This document is a copy of the [MACI 1.0 Specification (for audit) document](https://hackmd.io/AP6zPSgtThWxx6pjXY7R8A), created in 2021 for one of [MACI's formal audits](/docs/audit).
+
+This historical document is stored here for informational purposes. We do not intend to edit it. As a result, some of the information within this document may be outdated.
+:::
+
+This is a detailed specification meant to assist auditors in reviewing [MACI version 1.0](https://github.com/privacy-scaling-explorations/maci/releases/tag/v1.0.0).
 
 We thank the Zkopru team for their [protocol specification](https://github.com/zkopru-network/protocol-specification/), which this document adopts.
 
@@ -17,12 +21,13 @@ We thank the Zkopru team for their [protocol specification](https://github.com/z
 
 The commit hashes relevant to this audit are the following:
 
-| Name | Commit |
-|-|-|
-| `appliedzkp/maci` (`v1` branch) | [`2db5f625b67a6b810bd851950d7a42c26189088b`](https://github.com/appliedzkp/maci/tree/2db5f625b67a6b810bd851950d7a42c26189088b) | 
-| `weijiekoh/circomlib` (`feat/poseidon-encryption` branch) | [`0e78dde8c813b95f4585b0613927e9c4269de500`](https://github.com/weijiekoh/circomlib/tree/0e78dde8c813b95f4585b0613927e9c4269de500) |  
+| Name                                                      | Commit                                                                                                                             |
+| --------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| `appliedzkp/maci` (`v1` branch)                           | [`2db5f625b67a6b810bd851950d7a42c26189088b`](https://github.com/appliedzkp/maci/tree/2db5f625b67a6b810bd851950d7a42c26189088b)     |
+| `weijiekoh/circomlib` (`feat/poseidon-encryption` branch) | [`0e78dde8c813b95f4585b0613927e9c4269de500`](https://github.com/weijiekoh/circomlib/tree/0e78dde8c813b95f4585b0613927e9c4269de500) |
 
 The scope of the audit with regards to the `circomlib` library covers:
+
 - all the JS files that MACI references, **excluding** those which are not referenced by MACI's TS files
 - all circuit files **excluding** whose which are not referenced by MACI's circuit files
 
@@ -31,20 +36,19 @@ The scope of the audit with regards to the `circomlib` library covers:
 Through this audit, we wish to challenge the following statements:
 
 1. MACI exhibits collusion resistance
-    - No one except a trusted coordinator should be certain of the validity of a vote, reducing the effectiveness of bribery
+   - No one except a trusted coordinator should be certain of the validity of a vote, reducing the effectiveness of bribery
 2. MACI exhibits receipt-freeness
-    - No voter may prove (besides to the coordinator) which way they voted
+   - No voter may prove (besides to the coordinator) which way they voted
 3. MACI provides privacy
-    - No one except a trusted coordinator should be able to decrypt a vote
+   - No one except a trusted coordinator should be able to decrypt a vote
 4. MACI is uncensorable:
-    - No one (not even the trusted coordinator) should be able to censor a vote
+   - No one (not even the trusted coordinator) should be able to censor a vote
 5. MACI provides unforgeability
-    - Only the owner of a user's private key may cast a vote tied to its corresponding public key
+   - Only the owner of a user's private key may cast a vote tied to its corresponding public key
 6. MACI provides non-repudiation
-    - No one may modify or delete a vote after it is cast, although a user may cast another vote to nullify it
+   - No one may modify or delete a vote after it is cast, although a user may cast another vote to nullify it
 7. Correct execution
-    - No one (not even the trusted coordinator) should be able to produce a false tally of votes
-
+   - No one (not even the trusted coordinator) should be able to produce a false tally of votes
 
 [TOC]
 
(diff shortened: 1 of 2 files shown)


website/docusaurus.config.ts
```diff
@@ -104,6 +104,12 @@ const config: Config = {
                     label: "Blog",
                     position: "left",
                 },
+                {
```
ctrlc03, line 107: I would probably remove this and let it be indexed as another doc (inside the versioned_docs/verion-v1.x) folder

website/versioned_docs/version-v1.x/spec.md
```diff
@@ -0,0 +1,944 @@
+---
+id: spec
+title: Spec
+
+---
+
+# MACI 1.0 Specification (for audit)
```
ctrlc03, line 7: A lot of this content was already spread between the different markdown files (version-v1.x), for instance you will find all of the concepts around keypairs, messages and hash functions inside the primitives.md file. 
As not everything was added, I believe it might be more beneficial to actually extract the content that we are missing that is useful, and place it in the appropriate doc file (contracts, primitives, circuits, etc.).

ctrlc03 (changes requested):

website/versioned_docs/version-v1.x/spec.md
baumstern: MACI 1.0 Specification is a comprehensive document created specifically for auditors and those who need an in-depth understanding of MACI's technical aspects. While it shares topics like keypairs and hash functions with existing documentation, the specification delves into greater depth and provides an interconnected context that the current documents do not fully capture. This comprehensive approach is similar to how the [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) and the [Go language specification](https://go.dev/ref/spec) are provided as standalone documents, serving as in-depth references for their respective technologies.

Given these points, I believe keeping the MACI 1.0 Specification as a standalone document on our site would greatly benefit users needing detailed technical info.

Looking forward to hearing your thoughts on this.

website/docusaurus.config.ts
baumstern: I've considered your suggestion on integrating the MACI 1.0 Spec into our versioned documentation. However, I think maintaining it as a standalone link in the navigation could be more beneficial for users. This approach, often adopted in well-established projects, enhances the spec's visibility and accessibility, particularly for those seeking detailed technical insights. It simplifies finding and referencing this key document.

What are your thoughts on this?

website/versioned_docs/version-v1.x/spec.md
ctrlc03: I thought that the documentation itself would "greatly benefit users needing detailed technical info." and having everything divided into each component (which if you read you'll find contain extracts from this spec) might be easier to navigate than one standalone document.
Anyways please remove references to the audit (like the first section) if you want to use this as protocol spec

website/docusaurus.config.ts
samajammin: > I think maintaining it as a standalone link in the navigation could be more beneficial for users

I don't see this as necessary. It's a highly technical document & it's specific to MACI v1.0. I agree with @ctrlc03 that it makes sense within the v1 docs.

website/versioned_docs/version-v1.x/spec.md
```diff
@@ -0,0 +1,944 @@
+---
+id: spec
+title: Spec
+
```
samajammin, line 4: Should have `description`, `sidebar_label` & `sidebar_position` fields as well

website/versioned_docs/version-v1.x/spec.md
```diff
@@ -0,0 +1,944 @@
+---
+id: spec
+title: Spec
```
samajammin, line 3:
```suggestion
title: MACI 1.0 Specification
sidebar_label: Specification
sidebar_position: 14
```

website/versioned_docs/version-v1.x/spec.md
samajammin: I tend to agree with @ctrlc03. I fear this duplicates some content, which risks future mistakes of us updating content in one place but not the other. But if the goal is to serve as a static specification (like the Ethereum Yellow Paper), I'm ok with having it.

samajammin (changes requested):
A couple comments.

Also noticed the builds are failing. Looks like a broken link:
```
  [cause]: Error: Docusaurus found broken links!
  
  Please check the pages of your site in the list below, and make sure you don't reference any path that does not exist.
  Note: it's possible to ignore broken links with the 'onBrokenLinks' Docusaurus configuration, and let the build pass.
  
  Exhaustive list of all broken links found:
  
  - On source page path = /docs/spec:
     -> linking to %5Bhttps://%5D(https://github.com/iden3/circomlib/blob/master/src/eddsa.js) (resolved as: /docs/%5Bhttps://%5D(https://github.com/iden3/circomlib/blob/master/src/eddsa.js))
  ```

File: website/versioned_docs/version-v1.x/spec.md (hunk 2 above)
Comment (ctrlc03) [reply website/versioned_docs/version-v1.x/spec.md:7]: If we want to serve it as a static specification, how do we host it in a way that avoids confusing the user? as mentioned, some information in this document is outdated, do we go and change it (and basically let it be a plain copy of what we already have in all separate docs, which already shared quite a lot of content with), or leave it as it is (minus the references/wording of the audit) and specify clearly (somewhere/somehow) that it is outdated? Maybe we host it as a static file and just add a link to it in the introduction.md doc (same way as we reference to the original etherresearch post)?

File: website/docusaurus.config.ts (hunk 1 above)
Comment (baumstern) [reply website/docusaurus.config.ts:107]: Fair enough

File: website/versioned_docs/version-v1.x/spec.md (hunk 2 above)
Comment (baumstern) [reply website/versioned_docs/version-v1.x/spec.md:7]: @ctrlc03 I'd like to challenge that this documentation is outdated. Have there been any breaking changes since the last audit? For example, have there been changes to how the circuit works, or to the cryptographic primitives like the Baby Jubjub curve or Poseidon? Have there been modifications in the way we use the accumulator queue, or in the structure of the verifying key? Until such core components are changed, this is the only complete reference for understanding the technical details (!= just implementation details) of MACI

Comment (baumstern) [reply website/versioned_docs/version-v1.x/spec.md:7]: I don't think there is a duplication; rather, our working documentation is currently incomplete. For example, the [primitives](https://maci.pse.dev/docs/primitives) page lacks some important details that I attached below. width="1325" alt="Screenshot 2023-11-30 at 07 46 09" src="https://github.com/privacy-scaling-explorations/maci/assets/1610146/962f3dc1-1c85-40a4-a2d4-73d06a1ae1e6">

**Comment by baumstern** [comment]
> A couple comments.
> 
> Also noticed the builds are failing. Looks like a broken link:
> 
> ```
>   [cause]: Error: Docusaurus found broken links!
>   
>   Please check the pages of your site in the list below, and make sure you don't reference any path that does not exist.
>   Note: it's possible to ignore broken links with the 'onBrokenLinks' Docusaurus configuration, and let the build pass.
>   
>   Exhaustive list of all broken links found:
>   
>   - On source page path = /docs/spec:
>      -> linking to %5Bhttps://%5D(https://github.com/iden3/circomlib/blob/master/src/eddsa.js) (resolved as: /docs/%5Bhttps://%5D(https://github.com/iden3/circomlib/blob/master/src/eddsa.js))
> ```

oh Docusaurus is so smart 🤯

File: website/versioned_docs/version-v1.x/spec.md [hunk 5]
```diff
@@ -0,0 +1,944 @@
...
+A public key is a point on the Baby Jubjub curve. It is determistically derived from a private key $s$, the procedure to do so is almost identical to [RFC8032](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5).
+
+1. Format the private key  [1.3]
+2. Multiply $G$ by 8 and multiply the resulting point by the formatted private key to derive the public key $A$:
+    $B = 8 \cdot G$
+    $A = B \cdot h_4$
+
+#### 1.5. Digital signature generation
+
+We use the Edwards-curve Digital Signature Algorithm (EdDSA) to sign messages. The code which implements signature generation and verification is iden3's [implementation]([https://](https://github.com/iden3/circomlib/blob/master/src/eddsa.js)) in the circomlib library.
```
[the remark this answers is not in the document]
baumstern: Reflected in 2d1e3f2

website/docusaurus.config.ts
baumstern: Reflected in 2d1e3f2

website/versioned_docs/version-v1.x/spec.md
ctrlc03: Oh yea no one said all of this is outdated, hence why I used the word "some" (some information in this document is outdated). 
Also: "I believe it might be more beneficial to actually extract the content that we are missing that is useful, and place it in the appropriate doc file (contracts, primitives, circuits, etc.)." -> as you also highlighted as an example, the primitives file is missing some information present on this, it would have been very neat to add it in the respective sections 🙂 
Looks like we agreed to keep it as its own entry within the v1 docs, my only request is that you remove this section: "## Audit scope" and change the wording of the "### Statements that we wish to challenge" section -> as this is supposed to be used as a spec, it should be worded as such and not as a document addressed to an audit firm

baumstern: @ctrlc03 thank you for the clarification :) However, I'd like to preserve the original form since it's a primary resource (fyi [this link](https://www.wgu.edu/blog/what-difference-between-primary-secondary-source2304.html#openSubscriberModal) explains the difference between primary and secondary sources). Also, I believe in maintaining the original context as much as possible to prevent any loss of understanding about the data's background and origin due to middleman intervention. Separately, the "Audit scope" section allows readers to know precisely exactly which commit the audit was based on (it's not stated any other places), indicating that the document reflects that particular time point. Through this, discerning readers can infer that, although the current codebase may have changes since the audited version, they can still expect to find relevant information on the core design, as both the document and the current codebase pertain to version v1.

samajammin (approved):
Added an info banner to disclaim that this is a historic doc that won't be updated:

Image 2023-12-04 at 13 31 49

@baumstern @ctrlc03 hoping this resolves most of the differences of opinion you folks expressed. LMK if you think this is a reasonable compromise that we can move forward with.

ctrlc03 (approved):

Issue #792, opened 2023-10-30
Core package refactoring


ctrlc03:
The core package should be refactored to improve extensibility, documentation and ensure correctness. 

- [x] split from monolith file (https://github.com/privacy-scaling-explorations/maci/pull/846)
- [x] better tests (https://github.com/privacy-scaling-explorations/maci/pull/910 #954 #924)
- [x] better comments and typedoc (https://github.com/privacy-scaling-explorations/maci/pull/883, https://github.com/privacy-scaling-explorations/maci/pull/883)
- [x] fix types madness (BigInt -> bigint, etc.)
- [x] publish types (https://github.com/privacy-scaling-explorations/maci/pull/879)
- [x] ensure correctness (dae & ctrlc03)

added 12th Dec:
- [x] remove dead code
- [x] remove todo 

To be merged before working on this:
* https://github.com/privacy-scaling-explorations/maci/pull/767 
* https://github.com/privacy-scaling-explorations/maci/pull/758

PR #846, opened 2023-11-29
Restructure `maci-core`  for enhanced discoverability

baumstern:
## Overview
Initiating the core package refactoring as outlined in issue #792, this PR reorganize the single, extensive 1800-line `MaciState.ts` file.  The goal here is to improve the visibility of core components, mainly the `MaciState` and `Poll` classes, aligning with our README's outlined structure.

## Scope

The focus of this PR is to improve the directory structure for easier identification and access to the core classes:

- **Direct Access to Core Classes**: `MaciState` and `Poll` are now positioned prominently in the 'core/ts' directory.
- **Consolidated Supplementary Elements**: utility functions are grouped in a 'lib' subdirectory, streamlining the primary directory and highlighting the core classes.

## Next Step

The next step will involve extracting key functions and data structures as detailed in the README. This will further declutter the codebase and emphasize the primary functionalities.

## Out of Scope

This PR does not include:

- Documentation updates or additional commenting.
- Internal code logic refactoring or quality enhancements.
- Any functional changes or optimizations.

These aspects will be addressed in later stages, if necessary

## Intent

This restructuring serves as a foundational step towards a more intuitive and navigable codebase. By making core components more visible and accessible, we aim to facilitate quicker orientation for developers, both new and existing.

## Notes

Feedback on this structural shift is highly appreciated. We're particularly interested in how these changes affect ease of navigation and overall developer experience within the codebase.


baumstern:
The MaciState.ts is now concise, under 200 lines. I'm looking at ways to split the 1500-line Poll class into smaller parts.

core/ts/lib/cryptoUtils.ts
```diff
@@ -0,0 +1,143 @@
+import * as assert from "assert";
+
+interface TreeDepths {
```
ctrlc03, line 3: maybe worth separating interfaces to their own file too (interfaces.ts or types.ts could be a good name for that)

core/ts/lib/cryptoUtils.ts
```diff
@@ -0,0 +1,143 @@
+import * as assert from "assert";
```
ctrlc03, line 1: In general, other folders have "utils" saved inside an util folder or placed directly inside the ts folder, but if you feel strongly about calling it "lib", we should make sure to maintain consistency across all packages. I personally have no preference, just would be great to see all follow the same standard

baumstern:
> Given you should be diving into the code for deeper understanding of the protocol (and ensuring it's working as expected)

Technically, this is not within the scope of the refactoring. I suggest merging this PR if there are no blockers. By doing so, users will find it easier to navigate our codebase immediately. If we delay merging this PR for a week or a month, people will continue to struggle with the unclear structure in the meantime.

core/ts/lib/cryptoUtils.ts
baumstern: I would like to address it in the following PR. The main goal of this PR is to improve the visibility of the core components of the `MaciState` and `Poll` classes in our codebase.

core/ts/lib/cryptoUtils.ts
baumstern: Naming is always interesting topic but I'll just follow the current convetion for now

ctrlc03:
> > Given you should be diving into the code for deeper understanding of the protocol (and ensuring it's working as expected)
> 
> Technically, this is not within the scope of the refactoring. I suggest merging this PR if there are no blockers. By doing so, users will find it easier to navigate our codebase immediately. If we delay merging this PR for a week or a month, people will continue to struggle with the unclear structure in the meantime.

right so you meant it's not within the scope of this PR?

core/ts/lib/cryptoUtils.ts
ctrlc03: As long as you address it as some point 👍

core/ts/MaciState.ts
```diff
@@ -1430,30 +181,5 @@ class MaciState {
...
-
-const genSubsidyVkSig = (_stateTreeDepth: number, _intStateTreeDepth: number, _voteOptionTreeDepth: number): bigint => {
-  return (
-    (BigInt(_stateTreeDepth) << BigInt(128)) + (BigInt(_intStateTreeDepth) << BigInt(64)) + BigInt(_voteOptionTreeDepth)
-  );
-};
-
-export { MaxValues, TreeDepths, MaciState, Poll, genProcessVkSig, genTallyVkSig, genSubsidyVkSig };
+// todo: remove re-export `IncrementalQuinTree` it's originally exported by `maci-crypto`
+export { IncrementalQuinTree, MaciState };
```
ctrlc03, line 185: I'm not sure I understand why the export of IncrementalQuinTree was added? also given you added a comment saying that it should be removed

ctrlc03 (approved):
LGTM, just left one comment but nothing that would prevent this from being merged.

core/ts/MaciState.ts
baumstern: somehow after this refactoring MaciState.d.ts file doesn't export IncrementalQuinTree automatically so had to explicitly re-exported

PR #853, opened 2023-11-29
feat: update features on homepage

samajammin:
Working with the design team on building illustrations for these features, so wanted to quickly update these

## Before

Image 2023-11-29 at 13 49 33

## After

Image 2023-11-29 at 13 46 29

### DIFF
--- a/website/src/components/HomepageFeatures/index.tsx
+++ b/website/src/components/HomepageFeatures/index.tsx
@@ -10,22 +10,32 @@ type FeatureItem = {
 
 const FeatureList: FeatureItem[] = [
   {
-    title: "Easy to Use",
+    title: "Highly secure",
     Svg: require("@site/static/img/undraw_docusaurus_mountain.svg").default,
-    description: <>MACI was designed to be an easy to use voting solution.</>,
+    description: (
+      <>
+        MACI was designed to be a secure voting system. It is built on top of the Ethereum blockchain which ensures
+        votes cannot faked, censored, or tampered with.
+      </>
+    ),
   },
   {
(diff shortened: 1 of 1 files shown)


samajammin:
@ctrlc03 tagging you for review - I believe you originally created these features.

My main reasoning is that MACI optimizes for security > ease of use. Obviously we want both but when it comes to tradeoffs, I think it's clear the system has been designed to ensure security. I suspect there's many other voting systems which are easier to use 😄

ctrlc03 (changes requested):
After the first read of the first section, it feels like we are saying that these security properties come solely from the project using the Ethereum blockchain, whereas certain properties listed here do come from the protocol itself. Not sure if we can reword this? for instance, while votes are recorded on chain, the processing is done offchain, so technically the non-censurability comes from other parts of the code, not the blockchain itself.

The rest looks great, thanks for improving this section

samajammin:
> After the first read of the first section, it feels like we are saying that these security properties come solely from the project using the Ethereum blockchain, whereas certain properties listed here do come from the protocol itself. Not sure if we can reword this? for instance, while votes are recorded on chain, the processing is done offchain, so technically the non-censurability comes from other parts of the code, not the blockchain itself.
> 
> The rest looks great, thanks for improving this section

@ctrlc03 updated slightly to mention the smart contracts - how's that looks for now? Didn't mention the ZK stuff because that's highlighted in the other feature.

ctrlc03 (approved):
Thanks for improving this section, LGTM

ctrlc03 (dismissed):

ctrlc03 (approved):

PR #868, opened 2023-12-04
chore(cz): update commitizen (cz) and add scripts around it

ctrlc03:
This is in reference to us establishing our GH processes (#847). Users can easily commit using convential messages in an interactive way by running:

`npm run commit`

Furthermore, pre-commit hooks will ensure that the commit message follows the guideline.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ coverage
 build
 typechain-types
 .eslintrc.js
+commitlint.config.js
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+    extends: ['@commitlint/config-conventional'],
+  };
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,3 +1,3 @@
 module.exports = {
-    extends: ['@commitlint/config-conventional'],
-  };
+  extends: ["@commitlint/config-conventional"],
+};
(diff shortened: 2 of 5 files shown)


commitlint.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+    extends: ['@commitlint/config-conventional'],
+  };
```
0xmad, line 3: Please add new line at the end

package.json
```diff
@@ -42,5 +45,12 @@
     "commitizen": {
       "path": "./node_modules/cz-conventional-changelog"
     }
+  },
+  "husky": {
+    "hooks": {
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
+      "pre-commit": "lint-staged"
+    }
   }
```
0xmad, line 54: I don't think this is needed here. We have `.husky` folder with all the githooks

0xmad (changes requested):
@ctrlc03 thanks, just few comments.

.eslintignore
```diff
@@ -4,3 +4,4 @@ coverage
 build
 typechain-types
 .eslintrc.js
+commitlint.config.js
```
0xmad, line 7: And new line at the end here

ctrlc03: 😤

0xmad (approved):
@ctrlc03 thanks!

Issue #843, opened 2023-11-28
Smart contracts package refactoring


ctrlc03:
Currently, the smart contract package could use a refactor, especially the typescript files, and all of the build utilities/scripts.

- [x] rework deployment scripts (split into multiple files, etc. )
- [x] comment smart contract code better and add proper natspec comments 
- [x] remove dead code
- [x] remove todo 
- [x] unify solidity files
- [x] cleanup tests

ctrlc03:
fixed by #919 #913 #909 #898 #897

PR #870, opened 2023-12-04
docs(contracts): setup solidity docgen and configure it to work on docusaurus

### LINKED ISSUE #770: Better documentation in smart contracts and solidity docs
- [ ] Add https://github.com/OpenZeppelin/solidity-docgen
- [ ] ensure [NatSpecs](https://docs.soliditylang.org/en/latest/natspec-format.html) comments are formatted correctly
- [ ] Add generated docgen pages to website

ctrlc03:
fix #770 

This PR adds support for Solidity [Docgen](https://github.com/OpenZeppelin/solidity-docgen) as well as fixes syntax errors in current NatSpec comments (note it does not enhance the comments as that is to be worked on #843). Support for hosting the generated content on Docusaurus is also added


contracts/contracts/MACI.sol
```diff
@@ -15,40 +15,39 @@ import {SnarkConstants} from "./crypto/SnarkConstants.sol";
...
-/*
+/**
  * Minimum Anti-Collusion Infrastructure
  * Version 1
  */
 contract MACI is IMACI, DomainObjs, Params, SnarkCommon, Ownable {
-    // The state tree depth is fixed. As such it should be as large as feasible
-    // so that there can be as many users as possible.  i.e. 5 ** 10 = 9765625
-    // this should also match the parameter of the circom circuits.
+    /// @notice The state tree depth is fixed. As such it should be as large as feasible
```
samajammin, line 23: Why the triple slash? I'm guessing this was not intended.

contracts/contracts/MACI.sol
```diff
@@ -15,40 +15,39 @@ import {SnarkConstants} from "./crypto/SnarkConstants.sol";
...
+    /// so that there can be as many users as possible.  i.e. 5 ** 10 = 9765625
+    /// this should also match the parameter of the circom circuits.
     uint8 public immutable stateTreeDepth;
 
-    // IMPORTANT: remember to change the ballot tree depth
-    // in contracts/ts/genEmptyBallotRootsContract.ts file
-    // if we change the state tree depth!
-
+    /// @notice IMPORTANT: remember to change the ballot tree depth
```
samajammin, line 28: Same here (and in comments below)

contracts/contracts/MACI.sol
```diff
@@ -99,17 +98,17 @@ contract MACI is IMACI, DomainObjs, Params, SnarkCommon, Ownable {
     event MergeStateAqSubRoots(uint256 _pollId, uint256 _numSrQueueOps);
     event MergeStateAq(uint256 _pollId);
 
-    /*
-    * Ensure certain functions only run after the contract has been initialized
-    */
+    /**
+     * @notice Ensure certain functions only run after the contract has been initialized
```
samajammin, line 102: Prob makes sense to use a consistent commenting format? Could this be added to the linter @0xmad?

contracts/contracts/MessageProcessor.sol
```diff
@@ -29,15 +29,15 @@ contract MessageProcessor is Ownable, SnarkCommon, CommonUtilities, Hasher {
     error CurrentMessageBatchIndexTooLarge();
     error BatchEndIndexTooLarge();
 
-    // Whether there are unprocessed messages left
+    /// @notice Whether there are unprocessed messages left
```
samajammin, line 32: More triple slashes

contracts/contracts/Tally.sol
```diff
@@ -28,18 +28,18 @@ contract Tally is
...
-    //
-    // hash3(
-    //   hashLeftRight(merkle root of current results, salt0)
-    //   hashLeftRight(number of spent voice credits, salt1),
-    //   hashLeftRight(merkle root of the no. of spent voice credits per vote option, salt2)
-    // )
-    //
-    // Where each salt is unique and the merkle roots are of arrays of leaves
-    // TREE_ARITY ** voteOptionTreeDepth long.
+    /// @notice The commitment to the tally results. Its initial value is 0, but after
```
samajammin, line 31: Again 😄

contracts/package.json
```diff
@@ -36,7 +37,8 @@
     "maci-core": "^1.1.2",
     "maci-crypto": "^1.1.2",
     "maci-domainobjs": "^1.1.2",
-    "module-alias": "^2.2.2"
+    "module-alias": "^2.2.2",
+    "solidity-docgen": "^0.6.0-beta.36"
```
samajammin, line 41: Think this beta version is ok to use? I see they haven't released a stable v0.6 yet:
https://github.com/OpenZeppelin/solidity-docgen/releases

Overall it's just doc generation... so not too much risk...

contracts/contracts/MACI.sol
ctrlc03: https://docs.soliditylang.org/en/v0.8.16/natspec-format.html - you can define single line natspec comments with /// and multiline with /// or / ** */

contracts/package.json
ctrlc03: seems like the  last stable version was 0.5.7 and came out last year: https://www.npmjs.com/package/solidity-docgen?activeTab=versions
since then many beta versions.. 
I'd say it's ok as it's just docs but happy to go with what the team feels more comfortable with

contracts/contracts/MACI.sol
ctrlc03: let's go with /// for both single and multiline comments? https://docs.soliditylang.org/en/v0.8.20/natspec-format.html in their examples they use ///

website/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "setup-typedoc": "ts-node setupTypedoc.ts",
+    "setup-typedoc": "ts-node ./src/scripts/setupTypedoc.ts",
+    "setup-soliditydocs": "ts-node ./src/scripts/setupSolidityDocs.ts",
```
samajammin, line 7: Noticed trying to run `npm run build` without first running this cause my build to fail:

```
~/code/pse/maci/website on  docs/contracts [$!?] via ⬢ v18.18.2
➜ npm run build

> website@0.0.0 build
> docusaurus build

[INFO] [en] Creating an optimized production build...
Warn: `docsDir` doesn't exist: "/Users/samrichards/code/pse/maci/website/docs".



✔ Client


✖ Server
  Compiled with some errors in 16.84s

Error: Docusaurus server-side rendering could not render static page with path /solidity-docs/TopupCredit because of error: IERC20 is not defined
    at serverEntry_render (main:72233:358)
    at runNextTicks (node:internal/process/task_queues:60:5)
    at process.processImmediate (node:internal/timers:447:9)
    at async /Users/samrichards/code/pse/maci/website/node_modules/p-map/index.js:57:22 {
  [cause]: ReferenceError: IERC20 is not defined
      at _createMdxContent (main:3906:98)
      at MDXContent (main:3936:8)
      at Uc (main:120769:44)
      at Xc (main:120771:253)
      at Z (main:120777:89)
      at Xc (main:120775:231)
      at Z (main:120777:89)
      at Xc (main:120771:481)
      at Z (main:120777:89)
      at Xc (main:120771:481)
```

I wonder, should we just run these two scripts as part of the `build` script? e.g.
```
"build": "npm run setup-typedoc && npm run setup-soliditydocs && docusaurus build",
```
Unsure if that's a good best practice, but it would ensure we avoid that issue.

ctrlc03: great catch, I think running all of it in the build script is fine!

website/src/scripts/setupSolidityDocs.ts
```diff
@@ -0,0 +1,38 @@
+const fileS = require("fs");
+const pathX = require("path");
+
+const solidityDocDir = pathX.join(__dirname, "../pages/solidity-docs");
```
samajammin, line 4: Hmmmm not sure how I feel about dropping these directly in the /pages/ directory of our source code. Is there not a way to drop these in a /static/ directory, like we do the for TypeDoc files? Or somehow just have them in the final build vs. within /src/?

samajammin: If we must do this, we should at least add this directory to .gitignore. Otherwise all these files show up as untracked files after running `npm run build`

samajammin: It also feels like a weird pattern that running a script in the `contracts` dumps a bunch of files in `website`... maybe it'd be cleaner to add a step?
1. `npm run docs` from `contracts` generates the static docgen output somewhere in the package
2. `npm run setup-soliditydocs` from `website` copies over that static directory?

website/src/scripts/setupSolidityDocs.ts
```diff
@@ -0,0 +1,38 @@
+const fileS = require("fs");
```
samajammin, line 1: Let's document these scripts somewhere, either in the README or as comment sdirectly in these files, that explain we're pulling generated documentation from `solidity-docgen`/TypeDoc into the website.

website/src/scripts/setupSolidityDocs.ts
ctrlc03: > It also feels like a weird pattern that running a script in the `contracts` dumps a bunch of files in `website`... maybe it'd be cleaner to add a step?
> 
> 1. `npm run docs` from `contracts` generates the static docgen output somewhere in the package
> 2. `npm run setup-soliditydocs` from `website` copies over that static directory?

I like this, good suggestion!

contracts/contracts/MACI.sol
samajammin: Ah, ok, thanks for the explanation 👌

contracts/contracts/MACI.sol
samajammin: I'm down. Seems like it'd make it easier for changing a comment from single to multi-line? Or avoids the need for 3 lines when making a single-line comment.

website/src/scripts/setupSolidityDocs.ts
ctrlc03: > Hmmmm not sure how I feel about dropping these directly in the /pages/ directory of our source code. Is there not a way to drop these in a /static/ directory, like we do the for TypeDoc files? Or somehow just have them in the final build vs. within /src/?

I'll see if we can host them from within the artifacts, read that for markdown files the best way was to host within the pages folder - if that becomes too convoluted, could we go with the .gitignore trick?

website/src/scripts/setupSolidityDocs.ts
```diff
@@ -0,0 +1,38 @@
...
+      fileS.writeFileSync(absolutePath, fileContent, "utf8");
+
+      content += `- [${contractName}](./${fileRelativePathWithoutExtension})\n`;
+    }
+  }
+
+  return content;
+}
+
+const content = "# Solidity NatSpec Documentation\n\n" + generateMarkdownLinks(solidityDocDir, "solidity-docs");
```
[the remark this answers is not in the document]
ctrlc03: We could have an index.md file already and just happen this, for sure

contracts/package.json
samajammin: I'm ok since it's just doc generation

samajammin (changes requested):
Nice work! Looks good overall but added some change requests.

website/src/scripts/setupSolidityDocs.ts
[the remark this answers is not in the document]
ctrlc03: have you got any content in mind you'd like to add there? to which we can append the table of contents

website/src/scripts/setupSolidityDocs.ts
ctrlc03: Looks like markdown files inside the static folders are not rendered, so it would just be the raw markdown. Gonna resort to adding the folder to .gitignore for now.

samajammin: > I'll see if we can host them from within the artifacts, read that for markdown files the best way was to host within the pages folder - if that becomes too convoluted, could we go with the .gitignore trick?

Ah, I see, and TypeDocs is in /static/ because they're HTML files?

website/src/scripts/setupSolidityDocs.ts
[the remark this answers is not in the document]
samajammin:
```
# Solidity NatSpec Documentation

This directory provides autogenerated documentation for the [MACI v1.x smart contracts](/docs/contracts/) and interfaces based on the [NatSpec Format](https://docs.soliditylang.org/en/latest/natspec-format.html) for Solidity contracts.

## MACI contracts and interfaces

[TOC of contracts, linking to pages]
```

website/src/scripts/setupSolidityDocs.ts
ctrlc03: Correct! one option could be to put them in static and host them via some sort of React markdown library but think that would just make it slower/less responsive

.github/workflows/gh-pages.yml
```diff
@@ -22,17 +22,30 @@ jobs:
...
+
+      - name: Install website dependencies
         run: |
           npm install
+        working-directory: website
+
+      - name: Fix typedoc and solidity docgen links
+        run: |
           npm run setup-typedoc
+          npm run setup-soliditydocs
```
baumstern, line 48: Could we introduce a script to simplify the website generation process? Running a single command like `./build-website.sh` would make the build steps much more efficient.

baumstern (changes requested):

.github/workflows/gh-pages.yml
ctrlc03: Done

website/README.md
```diff
@@ -12,32 +12,45 @@ $ npm
 
 ### Local Development
 
+First, ensure you have setup MACI:
```
samajammin, line 15:
```suggestion
First, ensure you have setup MACI.

From the root of the project (not within this /website/ directory, run:
```
Clarifying the instruction a bit

website/docusaurus.config.ts
```diff
@@ -108,6 +108,11 @@ const config: Config = {
           label: "Typedoc",
```
samajammin, line 108: General thought (not something we need to update now) - should this just be "TypeScript Docs"? Is TypeDoc a well known thing? Saying TS might be clearer.

Might be more in line with the "Solidity Docs" label (e.g. we're not calling it "Docgen")

.github/scripts/website.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+# install and build project
+npm install && npm run bootstrap && npm run build
```
samajammin, line 6: Kinda unfortunate this all this is necessary now just to deploy the website. This is just because we need to run docgen from the contracts folder?

.github/workflows/gh-pages.yml
```diff
@@ -22,23 +22,9 @@ jobs:
...
-        run: |
-          npm install
-          npm run setup-typedoc
-        working-directory: website
-
-      - name: Build 🔧
-        run: |
-          npm run build
-        working-directory: website
+          ./.github/scripts/website.sh
```
samajammin, line 27: Is extracting this into a separate script really necessary? Feels like we're just creating more files for ourselves to look through.

Or is the `.github/scripts/website.sh` script used in multiple places? Or needed for Netlify or something?

samajammin: I suppose I'm a bit confused here. We have this `gh-pages` workflow, but we're also deploying the website to Netlify? Why do we need both?

ctrlc03: I added it cause it was needed for Netlify, no other reasons

samajammin: So why is `gh-pages` workflow needed if we deploy from Netlify?

ctrlc03: Netlify is only for preview! not sure how it'd work just deploying there @baumstern there might be costs involved/and/or an account to setup?

website/docusaurus.config.ts
ctrlc03: sure, I like that idea

samajammin (approved):
Approving with a few comments/suggestions, but nothing that needs to block merging.

Keeping this open incase @kittybest has comments/questions

@ctrlc03 Great work! One last suggestion for future reference, I think this PR could've ideally been split up into multiple PRs (1. adding docgen, 2. adding Solidity comments) for the sake of simpler reviews.

.github/workflows/gh-pages.yml
samajammin: Ahhhh, I gotcha. Ok that seems silly lol, but not a change for this PR. Let's resolve that separately.

ctrlc03:
> ocgen, 2. adding Solidity comments) for the

Sure, that's fair feedback and thanks! will fix conflicts and comments tomorrow

ctrlc03:
> Approving with a few comments/suggestions, but nothing that needs to block merging.
> 
> Keeping this open incase @kittybest has comments/questions
> 
> @ctrlc03 Great work! One last suggestion for future reference, I think this PR could've ideally been split up into multiple PRs (1. adding docgen, 2. adding Solidity comments) for the sake of simpler reviews.

I'm afraid some comments on functions were needed to make codegen run (it complained a lot)! but yes in general it's better to split it and there will be separate work to add full NatSpec

.github/workflows/gh-pages.yml
baumstern: Having a build script offers two benefits: it simplifies the gh-pages.yml and makes the build step more universally accessible to the community, regardless of knowledge of GitHub Actions' syntax

baumstern: > Netlify is only for preview! not sure how it'd work just deploying there @baumstern there might be costs involved/and/or an account to setup?

It's a free tier :)

baumstern: > Having a build script offers two benefits: it simplifies the gh-pages.yml and makes the build step more universally accessible to the community, regardless of their knowledge of GitHub Actions syntax

oh wow now we can building website just using `npm run build` command thanks to this PR: https://github.com/privacy-scaling-explorations/maci/pull/888

ctrlc03: so well now to get it working we have to build twice the website (check the setup-website.sh script) - first time we build with lerna, then have to build individually 🤔 
any suggestions @0xmad @samajammin @baumstern ?
Also I had to make the `setupSolidityDocs` and `setupTypedoc` silently complete without making any changes if either typedoc or the solidity output dirs are not present.. which might result in a user locally skipping some steps (run typedoc, and run solidity docgen) thought the website would still work, just it would miss content on typedoc and solidity docs.. curious about what you think here

baumstern: I think it's okay because the build time takes about 2 to 3 minutes

ctrlc03: not sure I follow you! how does that make it ok?

kittybest:
I have some thoughts for the typedocs and soliditydocs, but not really related to this PR. Try to put some thoughts here though, if it's not a proper place, then we could move to open a new discussion thread, or somewhere else.

First of all, I wanna ask that if there's version difference between different versions? For example, the `circuits.md` file is different in v0.x and v1.x. Will type docs and solidity docs also have this difference? Such as changing type from Int to BigInt for some arguments in different versions. If so, I would suggest to have this kind of document structure:
截圖 2023-12-11 下午5 09 13
Also it's clearer for me that both typedoc and soliditydocs are belong to Documents, instead of separating them on the navbar.

Second, I also suggest to split PR into docgen related one and NatSpec one.

Last is I think the netlify is not displaying correct website?

ctrlc03:
> I have some thoughts for the typedocs and soliditydocs, but not really related to this PR. Try to put some thoughts here though, if it's not a proper place, then we could move to open a new discussion thread, or somewhere else.
> 
> First of all, I wanna ask that if there's version difference between different versions? For example, the `circuits.md` file is different in v0.x and v1.x. Will type docs and solidity docs also have this difference? Such as changing type from Int to BigInt for some arguments in different versions. If so, I would suggest to have this kind of document structure: 截圖 2023-12-11 下午5 09 13 Also it's clearer for me that both typedoc and soliditydocs are belong to Documents, instead of separating them on the navbar.
> 
> Second, I also suggest to split PR into docgen related one and NatSpec one.
> 
> Last is I think the netlify is not displaying correct website?

For v0.x vs v1.x - we will continue to host v0.x docs (legacy) but do not intend to update them or touch them at all. These are just a reference of how MACI v.0x works (though we do not support it anymore). Moving forward, all new content will only be inside v.1x or newer versions. TLDR no typedoc nor natspec docs for v.0x

Sure we can put them in the documents folder as highlighted by that image, though we will not be making typedoc related changes on this PR, but please feel free to open one yourself to move them.

While I agree that the natspec could have been moved to a separate one (which is also coming today), some of the tags were broken and were preventing docgen to complete. Would it be ok to keep it like this for now, just change the folder where the docs are hosted?

Not sure what you mean by not displaying the correct website? what is wrong about it?
Currently: Netlify: preview only - GH pages with custom domain: deploys on pushes to dev

samajammin:
@ctrlc03 FYI looks like builds are failing in Netlify:

```
6:09:24 AM: $ ./.github/scripts/website.sh
6:09:24 AM: bash: ./.github/scripts/website.sh: No such file or directory
6:09:24 AM: ​
6:09:24 AM: "build.command" failed                                        
6:09:24 AM: ────────────────────────────────────────────────────────────────
6:09:24 AM: ​
6:09:24 AM:   Error message
6:09:24 AM:   Command failed with exit code 127: ./.github/scripts/website.sh (https://ntl.fyi/exit-code-127)
6:09:24 AM: ​
6:09:24 AM:   Error location
6:09:24 AM:   In Build command from Netlify app:
6:09:24 AM:   ./.github/scripts/website.sh
6:09:24 AM: ​
6:09:24 AM:   Resolved config
6:09:24 AM:   build:
6:09:24 AM:     command: ./.github/scripts/website.sh
6:09:24 AM:     commandOrigin: ui
6:09:24 AM:     environment:
6:09:24 AM:       - REVIEW_ID
6:09:24 AM:     publish: /opt/build/repo/website/build
6:09:24 AM:     publishOrigin: ui
6:09:25 AM: Failed during stage "building site": Build script returned non-zero exit code: 2
6:09:25 AM: Build failed due to a user error: Build script returned non-zero exit code: 2
6:09:25 AM: Failing build: Failed to build site
6:09:25 AM: Finished processing build request in 19.439s
```

Is Netlify able to access that script? 🤔

ctrlc03:
> @ctrlc03 FYI looks like builds are failing in Netlify:
> 
> ```
> 6:09:24 AM: $ ./.github/scripts/website.sh
> 6:09:24 AM: bash: ./.github/scripts/website.sh: No such file or directory
> 6:09:24 AM: ​
> 6:09:24 AM: "build.command" failed                                        
> 6:09:24 AM: ────────────────────────────────────────────────────────────────
> 6:09:24 AM: ​
> 6:09:24 AM:   Error message
> 6:09:24 AM:   Command failed with exit code 127: ./.github/scripts/website.sh (https://ntl.fyi/exit-code-127)
> 6:09:24 AM: ​
> 6:09:24 AM:   Error location
> 6:09:24 AM:   In Build command from Netlify app:
> 6:09:24 AM:   ./.github/scripts/website.sh
> 6:09:24 AM: ​
> 6:09:24 AM:   Resolved config
> 6:09:24 AM:   build:
> 6:09:24 AM:     command: ./.github/scripts/website.sh
> 6:09:24 AM:     commandOrigin: ui
> 6:09:24 AM:     environment:
> 6:09:24 AM:       - REVIEW_ID
> 6:09:24 AM:     publish: /opt/build/repo/website/build
> 6:09:24 AM:     publishOrigin: ui
> 6:09:25 AM: Failed during stage "building site": Build script returned non-zero exit code: 2
> 6:09:25 AM: Build failed due to a user error: Build script returned non-zero exit code: 2
> 6:09:25 AM: Failing build: Failed to build site
> 6:09:25 AM: Finished processing build request in 19.439s
> ```
> 
> Is Netlify able to access that script? 🤔

there was a name mismatch, fixed now on our side.

website/src/scripts/setupSolidityDocs.ts
```diff
@@ -0,0 +1,71 @@
...
+    const isDirectory = fs.statSync(absolutePath).isDirectory();
+
+    if (isDirectory) {
+      updateHeadings(absolutePath);
+    } else {
+      // Read the content of the file
+      let fileContent = fs.readFileSync(absolutePath, "utf8");
+
+      // Remove the "# Solidity API" line
+      fileContent = fileContent.replace("# Solidity API\n", "");
```
samajammin, line 54: ❤️

integrationTests/tsconfig.json
```diff
@@ -1,6 +1,9 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "target": "ES2020",
```
samajammin, line 4: Is this relevant?

website/src/scripts/setupSolidityDocs.ts
```diff
@@ -0,0 +1,71 @@
+import fs from "fs";
+import path from "path";
+
+const solidityDocDir = path.resolve(__dirname, "../../versioned_docs/version-v1.x/natspec-docs");
```
samajammin, line 4: I'd suggest maybe `solidity-api` or `contracts-api` instead of `natspec-docs`. Subjective take but I suspect many folks don't know what NatSpec is & may get confused.

website/src/scripts/setupSolidityDocs.ts
```diff
@@ -0,0 +1,71 @@
...
+const sourceDir = path.resolve(__dirname, "../../../contracts/docs");
```
samajammin, line 5: Add comment that we pull generated docs from the contracts package?

samajammin (approved):
LGTM! A couple comments but no blockers.

I didn't scrutinize the contract comments closely but website looks great & all pages are searchable! 🎉

website/src/scripts/setupSolidityDocs.ts
samajammin: I wonder if there's also a way to specify the display name of the directory (e.g. to "Solidity docs" or "Solidity API") 🤔 

Image 2023-12-11 at 21 44 46

website/src/scripts/setupSolidityDocs.ts
ctrlc03: sure

integrationTests/tsconfig.json
ctrlc03: that was a rebase conflicts not properly solved (duplicate values) - should disappear now from the changed files

website/src/scripts/setupSolidityDocs.ts
ctrlc03: I think we can accomplish that with a custom sidebar, can do that in a later PR if you want - we can directly specify order and name there

baumstern (approved):
Great work, thanks!

website/src/scripts/setupSolidityDocs.ts
samajammin: > I think we can accomplish that with a custom sidebar, can do that in a later PR if you want - we can directly specify order and name there

Sweet, love that idea!

PR #881, opened 2023-12-06
chore: update blog slugs to kebab case

samajammin:
This is very much a subjective nitpick (I'm a bit OCD from my SEO days), so open to a discussion here...

I'd just like to adopt a simple convention for URLs. I suggest we use kebab case, so pretty much:
- Use lowercase letters
- Separate words with hyphens (don't use spaces)
- Avoid other characters (periods, underscores,etc)

I just personally hate seeing encoded spaces, e.g. 
https://maci.pse.dev/blog/MACI%201.0%20Technical%20Introduction

Thoughts?

### DIFF
--- a/website/blog/2021-10-12-maci-v1.md
+++ b/website/blog/2021-10-12-maci-v1.md
@@ -1,5 +1,5 @@
 ---
-slug: MACI 1.0
+slug: maci-1-0-release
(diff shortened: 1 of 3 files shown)


ctrlc03 (approved):
Have no preferences myself so happy with either!

Issue #845, opened 2023-11-28
Make linter stricter and unified


0xmad:
As a developer, I'd like to have linter configuration for all the packages with predefined rules.

### Acceptance criteria:
- [x] Add linter config package
- [x] Get rid of `public` keyword for classes
- [x] Apply for `circuits`
- [x] Apply for `cli`
- [x] Apply for `common`
- [x] Apply for `contracts`
- [x] Apply for `core`
- [x] Apply for `crypto`
- [x] Apply for `domainobjs`
- [x] Apply for `integrationTests`
- [x] Apply for `server`
- [x] Apply for `website`
- [x] Fix solidity lint errors
- [ ] Add readme docs for linter

samajammin:
Request - as part of this work, please update our README/docs with how the linting setup & configuration works

PR #887, opened 2023-12-07
chore(intergationTests): apply linter for integration tests package

0xmad:
Blocked by #882
Related to #845


integrationTests/ts/__tests__/integration.test.ts
```diff
@@ -29,19 +38,19 @@ import {
...
-import { homedir } from "os";
+import { ITestSuite, Subsidy } from "./utils/interfaces";
 import { expectSubsidy, expectTally, genTestUserCommands, isArm } from "./utils/utils";
-import { genPubKey, genRandomSalt } from "maci-crypto";
-import { existsSync, readFileSync, readdir, unlinkSync } from "fs";
-import { expect } from "chai";
+
+chai.use(chaiAsPromised);
+
+const { expect } = chai;
```
ctrlc03, line 46: can we keep it in one line and do `import chai, { expect } from "chai";` ?

ctrlc03 (approved):
Only one comment, great stuff thanks

integrationTests/ts/__tests__/integration.test.ts
0xmad: Unfortunately, no. Before it I need to promisify chai to make it work with typedoc.

ctrlc03: Fair enough!

PR #888, opened 2023-12-07
chore(website): apply linter for website and remove unused packages 

0xmad:
- [x] Remove maci-common
- [x] Remove maci-server
- [x] Linter fixes
- [x] Add linter checks for tsx files
- [x] Add commitlint workflow

Blocked by #887
Related to #845


website/setupTypedoc.ts
```diff
@@ -1,22 +1,23 @@
...
+import path from "path";
 
 // Define the directory where the Typedoc HTML files are located
-const typedocDir = path.join(__dirname, "static/typedoc_output");
+const typedocDir = path.resolve(__dirname, "static/typedoc_output");
 
 // Define a recursive function to find all HTML files in a directory
 function findHtmlFiles(dir: string, fileList: string[] = []) {
   const files = fs.readdirSync(dir);
+  let list: string[] = [];
```
ctrlc03, line 10: Looks like this script is now only returning the html files inside the last directory it recurses into. Please could you amend this? one possible fix would be:

```ts
// Define a recursive function to find all HTML files in a directory
function findHtmlFiles(dir: string) {
  const files = fs.readdirSync(dir);
  let list: string[] = [];

  files.forEach((file: string) => {
    if (fs.statSync(path.join(dir, file)).isDirectory()) {
      list.concat(findHtmlFiles(path.join(dir, file)));
    } else if (file.endsWith(".html")) {
      list.push(path.join(dir, file));
    }
  });

  return list;
}
```

ctrlc03 (changes requested):
Just one change request, other than that great stuff!

baumstern (approved):

PR #895, opened 2023-12-11
chore(circuits): add test file routes to circuit test commands

kittybest:
Most of the commands testing single test are not pointing to the exact test file path.
Add the path in the command.
e.g.
`"test-hasher": "ts-mocha --exit Hasher.test.ts"` ➡️ `"test-hasher": "ts-mocha --exit ts/__tests__/Hasher.test.ts"`


ctrlc03 (approved):
Good catch, thank you!

PR #896, opened 2023-12-11
docs(typedoc): Typedoc Indexing

kittybest:
**Description**
Try the package [docusaurus-plugin-typedoc](https://www.npmjs.com/package/docusaurus-plugin-typedoc) at the beginning, it could automatically generate markdown files for typedocs, and put them inside `doc/` folder, which is default for the **Next** version document in Docusaurus. Also the sidebar will also be auto-generated.
However, we will need to do something before we actually start or build the docusaurus:
1. replace the package folder link in `README.md` (which should be the index page of typedoc) to github links. (currently we are doing it in the `setupTypedoc.ts`, by replacing the route in html file ➡️ https://github.com/privacy-scaling-explorations/maci/blob/88bfae966ec602592691b3d3ae8ad53e4a4c9a6c/website/src/scripts/setupTypedoc.ts#L38
2. `docs/` is not the actual folder we put our documents, so will also need to move or copy the whole `typedoc/` folder to `versioned_docs/v1.x/`.

It's hard to split the command in [docusaurus-plugin-typedoc](https://www.npmjs.com/package/docusaurus-plugin-typedoc), so I instead, tried to use only [typedoc-plugin-markdown](https://www.npmjs.com/package/typedoc-plugin-markdown). I put the scripts doing the above 2 steps in the `setupTypedoc.ts` file.

But there's still a problem, that the sidebar name is not setup as **Typedoc**. I will fix it.

Also wanna ask that the if we wanna auto-generate the index page of typedoc from the `README.md`? or we wanna write our own?

Last is that should we put the generated markdown files in the repo instead of ignoring them?

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "bootstrap": "lerna bootstrap",
-    "build": "lerna run build",
+    "build": "lerna run build --ignore=website",
     "clean": "lerna exec -- rm -rf node_modules build && rm -rf node_modules",
     "commit": "git cz",
     "prettier": "prettier -c .",
@@ -18,7 +18,7 @@
     "test:cli": "lerna run test --scope \"maci-cli\"",
     "test:integration": "lerna run test --scope \"maci-integrationtests\"",
     "test": "lerna run test --ignore maci-integrationtests --ignore maci-cli",
-    "typedoc": "typedoc --options typedoc.json",
+    "typedoc": "typedoc --plugin typedoc-plugin-markdown --options typedoc.json",
     "prepare": "is-ci || husky install"
   },
   "author": "PSE",
@@ -48,6 +48,7 @@
     "prettier-plugin-solidity": "^1.2.0",
     "solhint": "^4.0.0",
     "typedoc": "^0.25.4",
+    "typedoc-plugin-markdown": "^3.17.1",
     "typescript": "^5.3.2"
   },
   "config": {
--- a/website/src/scripts/setupSolidityDocs.ts
+++ b/website/src/scripts/setupSolidityDocs.ts
@@ -1,38 +1,12 @@
 import fs from "fs";
 import path from "path";
 
+import { copyDirectory } from "./utils";
+
 // where to move the solidity doc files over
 const solidityDocDir = path.resolve(__dirname, "../../versioned_docs/version-v1.x/solidity-docs");
 // the origin folder (from the contracts package)
 const sourceDir = path.resolve(__dirname, "../../../contracts/docs");
-
-/**
- * Allow to copy a directory from source to target
- * @param source - the source directory
- * @param target - the target directory
- */
-function copyDirectory(source: string, target: string) {
-  if (!fs.existsSync(target)) {
-    fs.mkdirSync(target, { recursive: true });
-  }
-
-  if (!fs.existsSync(source)) {
-    return;
-  }
-
-  const files = fs.readdirSync(source);
-
-  files.forEach((file: string) => {
-    const sourcePath = path.join(source, file);
-    const targetPath = path.join(target, file);
-
-    if (fs.lstatSync(sourcePath).isDirectory()) {
-      copyDirectory(sourcePath, targetPath);
-    } else {
-      fs.copyFileSync(sourcePath, targetPath);
-    }
-  });
-}
 
 /**
  * Currently, Solidity docgen generates the same heading for all files.
--- a/website/src/scripts/setupTypedoc.ts
+++ b/website/src/scripts/setupTypedoc.ts
@@ -5,6 +5,10 @@ import { copyDirectory } from "./utils";
 
 const TYPEDOC_DIR = path.resolve(__dirname, "../../typedoc");
 
+/**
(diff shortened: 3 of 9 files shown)


ctrlc03:
> **Description** Try the package [docusaurus-plugin-typedoc](https://www.npmjs.com/package/docusaurus-plugin-typedoc) at the beginning, it could automatically generate markdown files for typedocs, and put them inside `doc/` folder, which is default for the **Next** version document in Docusaurus. Also the sidebar will also be auto-generated. However, we will need to do something before we actually start or build the docusaurus:
> 
> 1. replace the package folder link in `README.md` (which should be the index page of typedoc) to github links. (currently we are doing it in the `setupTypedoc.ts`, by replacing the route in html file ➡️ https://github.com/privacy-scaling-explorations/maci/blob/88bfae966ec602592691b3d3ae8ad53e4a4c9a6c/website/src/scripts/setupTypedoc.ts#L38
> 2. `docs/` is not the actual folder we put our documents, so will also need to move or copy the whole `typedoc/` folder to `versioned_docs/v1.x/`.
> 
> It's hard to split the command in [docusaurus-plugin-typedoc](https://www.npmjs.com/package/docusaurus-plugin-typedoc), so I instead, tried to use only [typedoc-plugin-markdown](https://www.npmjs.com/package/typedoc-plugin-markdown). I put the scripts doing the above 2 steps in the `setupTypedoc.ts` file.
> 
> But there's still a problem, that the sidebar name is not setup as **Typedoc**. I will fix it.
> 
> Also wanna ask that the if we wanna auto-generate the index page of typedoc from the `README.md`? or we wanna write our own?
> 
> Last is that should we put the generated markdown files in the repo instead of ignoring them?

Last is that should we put the generated markdown files in the repo instead of ignoring them? I would suggest to .gitignore them - anyways they will be generated on the gh-pages branch upon deployment

samajammin:
> docs/ is not the actual folder we put our documents, so will also need to move or copy the whole typedoc/ folder to versioned_docs/v1.x/.

Hmmm so is the takeaway that [`docusaurus-plugin-typedoc`](https://www.npmjs.com/package/docusaurus-plugin-typedoc) doesn't support versioned docs? That seems surprising... feels like a pretty common use of Docusarus 🤔 In any case, sounds like you found a solution to copy them over to the write location.

> But there's still a problem, that the sidebar name is not setup as Typedoc. I will fix it.

What's the current name - is it the same as the directory slug? If so, I noticed the same issue in (https://github.com/privacy-scaling-explorations/maci/pull/870#discussion_r1423431026)

> Also wanna ask that the if we wanna auto-generate the index page of typedoc from the README.md? or we wanna write our own?

Do we need an index page? Happy to write our own if so - I can add a quick explainer. Better yet, why don't you take a quick swing at a description for it (e.g. these API docs are generated using TypeDoc for xyz packages of the repo).

> Last is that should we put the generated markdown files in the repo instead of ignoring them?

I agree with @ctrlc03 I think we can gitignore these since they're autogenerated.

kittybest:
> Do we need an index page? Happy to write our own if so - I can add a quick explainer. Better yet, why don't you take a quick swing at a description for it (e.g. these API docs are generated using TypeDoc for xyz packages of the repo).

Currently use the modules.md page as the index page, if anyone wants to add any description on the page, just leave a comment :) or it could be an independent issue~~

> What's the current name - is it the same as the directory slug? If so, I noticed the same issue in (https://github.com/privacy-scaling-explorations/maci/pull/870#discussion_r1423431026)

Yes it's actually the same question. Now it's automatically generated as just `index.md` (since this file is its homepage).
Since they are the same question, I think we could open an issue to deal with it, not modifying it on this branch, that would be clearer.

website/src/scripts/setupSolidityDocs.ts
```diff
@@ -1,39 +1,13 @@
 import fs from "fs";
 import path from "path";
 
+import { copyDirectory } from "./utils";
```
ctrlc03, line 4: neat

website/src/scripts/setupTypedoc.ts
```diff
@@ -1,42 +1,56 @@
...
+updateMentionFiles("interfaces");
+updateMentionFiles("modules");
+
+// find the target moving directory
+const versionFile = path.resolve(__dirname, "../../versions.json");
+let versionDir = "";
+try {
+  const versionContent = fs.readFileSync(versionFile, "utf8");
+  if (versionContent) {
+    const versionContentJson: string[] = JSON.parse(versionContent) as string[];
```
ctrlc03, line 46: you can remove `: string[]` as it's redundant (given you cast using `as string[]`)

samajammin:
> I think we could open an issue to deal with it, not modifying it on this branch, that would be clearer.

@kittybest agreed - do you mind creating an issue to capture this?

Nice work on the PR!

package.json
```diff
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "bootstrap": "lerna bootstrap",
-    "build": "lerna run build",
+    "build": "lerna run build --ignore=website",
```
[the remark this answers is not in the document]
kittybest: Because we could only run `npm run typedoc` in root folder only after all the packages are built (instead of website). The `npm run build` command in `website/` would do things to Typedocs and put them in the `versioned_docs/v1.x/` folder. Which means the runnable order of the script should be 
1. build all packages
2. build typedoc and soliditydocs
3. build website

Of course we could ignore the `setup-typedoc` and `setup-soliditydocs` command for the first build while running `lerna run build` command, and build typedocs and soliditydocs, and build the website again; but I don't think we should build the website twice. That's why I ignore the website package for the first build.

PR #908, opened 2023-12-12
fix(cli): check that the subsidy vk is set

### LINKED ISSUE #446: Verify local subsidy `.zkey` file with on-chain verifying key
https://github.com/privacy-scaling-explorations/maci/blob/v1/cli/ts/checkVerifyingKey.ts

ctrlc03:
add a check that the subsidy verifying key was set correctly on chain inside the cli command `checkVerifyingKeys`

fix #446


cli/ts/commands/setVerifyingKeys.ts
```diff
@@ -120,16 +120,17 @@ export const setVerifyingKeys = async (
...
     const ssIntStateTreeDepth = Number(ssMatch[2]);
     const ssVoteOptionTreeDepth = Number(ssMatch[3]);
+
     if (
-      stateTreeDepth !== ssStateTreeDepth ||
-      intStateTreeDepth !== ssIntStateTreeDepth ||
-      voteOptionTreeDepth !== ssVoteOptionTreeDepth
+      Number(stateTreeDepth) !== ssStateTreeDepth ||
+      Number(intStateTreeDepth) !== ssIntStateTreeDepth ||
+      Number(voteOptionTreeDepth) !== ssVoteOptionTreeDepth
```
kittybest, line 127: Why we need the `Number` here? also same question for L84:89, since I think the input already define them as `number`?

ctrlc03: @kittybest fixed now

kittybest (approved):

PR #913, opened 2023-12-13
refactor(contracts): refactor contracts code 

ctrlc03:
# Description

Refactor by removing dead code, and various optimizations.

- [x] Remove dead code 
- [x] Remove VkRegistry from ExtContracts and pass it directly to Mp, Tally and Subsidy - Poll does not need it and we save on external calls on these contracts
- [x] Remove onlyOwner from pollFactory to save on gas - It's fine if anyone deploys a Poll using the PollFactory - only Polls that have access to privileged MACI's functions are Polls deployed via MACI's `deployPoll` function
- [x] Remove init function from MACI's contract as this is not needed: If PollFactory is open to all, there is no need to check that ownership is assigned to the MACI contract, so we can remove the init function and modifiers to save gas, and just pass the top up credit address in the constructor.

## Related issue(s)

Re #843 

## Confirmation


baumstern (approved):
decentralizing PollFactory +1

Issue #415, opened 2022-04-29
Problem of vote overriding and key switching


chnejohnson:
## TL;DR

The problem is that messages are processed in reverse order of being published. This brings about that any message with nonce 1 can override the previous message regardless of key switching or not.

## Collusion resistance argument

According to the scenario in [A Technical Introduction to MACI 1.0](https://medium.com/privacy-scaling-explorations/a-technical-introduction-to-maci-1-0-db95c3a9439a), the article illustrates the collusion resistance argument in section **2. System Overview - b. Vote Overriding and Public Key Switching**.

The core concept is that:
> A voter can tell a briber that they are voting for option A, but in reality they voted for option B.

And the example is as follows:
1. Bob signs up with public key 1
2. Bob then creates a command that contains — a vote for option A and public key 2
3. Bob signs this command with public key 1, the key he used to sign up
4. Bob encrypts this command into a message and submits it to the MACI smart contracts
5. The coordinator decrypts this message, and checks to ensure that the command is signed by Bob’s previous key — public key 1. This message is valid.
6. The coordinator then records Bob’s vote for option A and updates his public key to public key 2

> At this point, Bob has successfully voted for option A, and in order to override this vote must send in a new vote with a signature from public key 2. At this point, a briber now tries to get Bob to vote for option B:

1. Bob creates a command that contains — a vote for option B and public key 1
2. Bob signs this command with public key 1, encrypts the message and submits it to the MACI smart contracts
3. Bob shows the briber the decrypted message as proof of his vote for option B
4. The coordinator decrypts Bob’s message and sees that the signature does not match up with public key 2 — Bob’s previous key added in his previous message. Therefore this message is invalid and this vote is not counted in the final tally.
5. The briber has no way of knowing whether the vote was valid or invalid, and so is not incentivized to offer bribes to other users.

## Implementation
Let's implement the above scenario with MACI. Assume that Bob's initial keypair are **pk** (public key) and **sk** (private key), and Bob's new key are denoted as **npk** and **nsk**.

Now Bob published two messages to vote for option 1 with vote weight equal to 3, and vote for option 2 with vote weight equal to 2 as follows:

```
1. pk+sk => 1:3:2 (vote-option=1, vote-weight=3, nonce=2)
2. pk+sk => 2:2:1 (vote-option=2, vote-weight=2, nonce=1)
```

[...]

PR #914, opened 2023-12-13
refactor(core): refactor the process message functions and general cleanup

ctrlc03:
# Description

- [x] Remove unnecessary code
- [x]  fix types
- [x] remove unused variables or duplicate ones
- [x] refactor the process messages functions to ensure they can be tested individually
- [x] add custom error messages for message processing debugging
- [x]  ensure code is well commented and explained

## Related issue(s)

Related to #792 

## Confirmation

### DIFF
--- a/core/ts/MaciState.ts
+++ b/core/ts/MaciState.ts
@@ -9,15 +9,19 @@ import { IJsonMaciState, IJsonPoll, IMaciState, MaxValues, TreeDepths } from "./
  * A representation of the MACI contract.
  */
 export class MaciState implements IMaciState {
+  // a MaciState can hold multiple polls
   public polls: Poll[] = [];
 
+  // in this quinary tree we hold all signups (hash of a state leaf)
   public stateTree: IncrementalQuinTree;
+  // the leaves of the state tree
   public stateLeaves: StateLeaf[] = [];
+  // how deep the state tree is
+  public stateTreeDepth: number;
 
   public numSignUps = 0;
 
-  public stateTreeDepth: number;
-
+  // to keep track if a poll is currently being processed
   public pollBeingProcessed: boolean;
   public currentPollBeingProcessed: number;
 
@@ -29,6 +33,7 @@ export class MaciState implements IMaciState {
     this.stateTreeDepth = stateTreeDepth;
     this.stateTree = new IncrementalQuinTree(this.stateTreeDepth, blankStateLeafHash, STATE_TREE_ARITY, hash5);
 
+    // we put a blank state leaf to prevent a DoS attack
     this.stateLeaves.push(blankStateLeaf);
     this.stateTree.insert(blankStateLeafHash);
   }
@@ -51,7 +56,6 @@ export class MaciState implements IMaciState {
 
   /**
    * Deploy a new poll with the given parameters.
-   * @param duration - The duration of the poll in seconds.
    * @param pollEndTimestamp - The Unix timestamp at which the poll ends.
    * @param maxValues - The maximum number of values for each vote option.
    * @param treeDepths - The depths of the tree.
@@ -60,16 +64,13 @@ export class MaciState implements IMaciState {
    * @returns The index of the newly deployed poll.
    */
   public deployPoll(
-    duration: number,
     pollEndTimestamp: bigint,
     maxValues: MaxValues,
     treeDepths: TreeDepths,
     messageBatchSize: number,
     coordinatorKeypair: Keypair,
   ): number {
-    // TODO: fix the order of the arguments
     const poll: Poll = new Poll(
-      duration,
       pollEndTimestamp,
       coordinatorKeypair,
       treeDepths,
@@ -80,7 +81,6 @@ export class MaciState implements IMaciState {
       },
       maxValues,
       this,
-      this.stateTreeDepth,
     );
 
     this.polls.push(poll);
--- a/core/ts/Poll.ts
+++ b/core/ts/Poll.ts
@@ -22,7 +22,6 @@ import {
   Ballot,
   PrivKey,
   blankStateLeaf,
-  blankStateLeafHash,
   IMessageContractParams,
   IJsonCommand,
   IJsonPCommand,
@@ -40,13 +39,14 @@ import {
   IJsonPoll,
   IProcessMessagesOutput,
 } from "./utils/types";
+import { ProcessMessageErrors, ProcessMessageError } from "./utils/errors";
 import { STATE_TREE_ARITY, MESSAGE_TREE_ARITY, VOTE_OPTION_TREE_ARITY } from "./utils/constants";
+import { PathElements } from "optimisedmt";
 
 /**
  * A representation of the Poll contract.
  */
 export class Poll implements IPoll {
-  public duration: number;
   // Note that we only store the PubKey on-chain while this class stores the
   // Keypair for the sake of convenience
   public coordinatorKeypair: Keypair;
@@ -56,8 +56,6 @@ export class Poll implements IPoll {
 
   // the depth of the state tree
   public stateTreeDepth: number;
-
-  public numSignUps: number;
 
   public pollEndTimestamp: bigint;
 
@@ -100,28 +98,28 @@ export class Poll implements IPoll {
   public MM = 50; // adjustable parameter
   public WW = 4; // number of digits for float representation
 
+  // an empty ballot and its hash to be used as zero value in the
+  // ballot tree
+  public emptyBallot: Ballot;
+  public emptyBallotHash: bigint;
+
   /**
    * Constructs a new Poll object.
-   * @param duration - The duration of the poll in seconds.
    * @param pollEndTimestamp - The Unix timestamp at which the poll ends.
    * @param coordinatorKeypair - The keypair of the coordinator.
    * @param treeDepths - The depths of the trees used in the poll.
    * @param batchSizes - The sizes of the batches used in the poll.
    * @param maxValues - The maximum values the MACI circuits can accept.
    * @param maciStateRef - The reference to the MACI state.
-   * @param stateTreeDepth - The depth of the state tree.
    */
   constructor(
-    duration: number,
     pollEndTimestamp: bigint,
     coordinatorKeypair: Keypair,
     treeDepths: TreeDepths,
     batchSizes: BatchSizes,
     maxValues: MaxValues,
     maciStateRef: MaciState,
-    stateTreeDepth: number,
   ) {
-    this.duration = duration;
     this.pollEndTimestamp = pollEndTimestamp;
     this.coordinatorKeypair = coordinatorKeypair;
     this.treeDepths = treeDepths;
@@ -129,10 +127,8 @@ export class Poll implements IPoll {
     this.maxValues = maxValues;
     this.maciStateRef = maciStateRef;
     this.pollId = maciStateRef.polls.length;
-    this.numSignUps = Number(maciStateRef.numSignUps.toString());
-    this.stateTreeDepth = stateTreeDepth;
+    this.stateTreeDepth = maciStateRef.stateTreeDepth;
 
-    this.stateTree = new IncrementalQuinTree(this.stateTreeDepth, blankStateLeafHash, STATE_TREE_ARITY, hash5);
     this.messageTree = new IncrementalQuinTree(
       this.treeDepths.messageTreeDepth,
       NOTHING_UP_MY_SLEEVE,
@@ -140,14 +136,13 @@ export class Poll implements IPoll {
       hash5,
     );
 
-    for (let i = 0; i < this.maxValues.maxVoteOptions; i++) {
-      this.tallyResult.push(BigInt(0));
-      this.perVOSpentVoiceCredits.push(BigInt(0));
-      this.subsidy.push(BigInt(0));
-    }
+    this.tallyResult = new Array(this.maxValues.maxVoteOptions).fill(BigInt(0));
+    this.perVOSpentVoiceCredits = new Array(this.maxValues.maxVoteOptions).fill(BigInt(0));
+    this.subsidy = new Array(this.maxValues.maxVoteOptions).fill(BigInt(0));
 
-    const blankBallot = Ballot.genBlankBallot(this.maxValues.maxVoteOptions, treeDepths.voteOptionTreeDepth);
-    this.ballots.push(blankBallot);
+    // we put a blank state leaf to prevent a DoS attack
+    this.emptyBallot = Ballot.genBlankBallot(this.maxValues.maxVoteOptions, treeDepths.voteOptionTreeDepth);
+    this.ballots.push(this.emptyBallot);
   }
 
   /**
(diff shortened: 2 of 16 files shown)


core/ts/Poll.ts
```diff
@@ -600,44 +647,60 @@ export class Poll implements IPoll {
...
+    // then take the ones part of this batch
     encPubKeys = encPubKeys.slice(index, index + messageBatchSize);
 
     const msgRoot = this.messageTree.root;
     const currentStateRoot = this.stateTree.root;
     const currentBallotRoot = this.ballotTree.root;
+    // calculate the current state and ballot root
+    // commitment
+    // which is the hash  of the state tree root,
+    // the ballot tree root and a salt
```
kittybest, line 703: Thinking about putting the only the `commitment` in a single line would be weird 🤔
Also there's an extra space between `hash` and `of`

core/ts/Poll.ts
```diff
@@ -600,44 +647,60 @@ export class Poll implements IPoll {
...
     if (batchEndIndex > this.messages.length) {
       batchEndIndex = this.messages.length;
     }
 
+    // copy the public keys, pad the array with the last keys if needed
     let encPubKeys = this.encPubKeys.map((x) => x.copy());
     while (encPubKeys.length % messageBatchSize > 0) {
       encPubKeys.push(encPubKeys[encPubKeys.length - 1]);
     }
+    // then take the ones part of this batch
```
kittybest, line 694: Not quite understand this description? take some of the `encPubKeys` part of this batch?

core/ts/Poll.ts
```diff
@@ -121,33 +122,29 @@ export class Poll implements IPoll {
...
-      this.perVOSpentVoiceCredits.push(BigInt(0));
-      this.subsidy.push(BigInt(0));
-    }
+    this.tallyResult = new Array(this.maxValues.maxVoteOptions).fill(BigInt(0));
+    this.perVOSpentVoiceCredits = new Array(this.maxValues.maxVoteOptions).fill(BigInt(0));
+    this.subsidy = new Array(this.maxValues.maxVoteOptions).fill(BigInt(0));
 
-    const blankBallot = Ballot.genBlankBallot(this.maxValues.maxVoteOptions, treeDepths.voteOptionTreeDepth);
-    this.ballots.push(blankBallot);
+    // we put a blank state leaf to prevent a DoS attack
```
samajammin, line 145: I don't understand how this prevents a DoS attack 🤔

ctrlc03: https://maci.pse.dev/docs/spec -> section 6.1 

It's explained in more details there :)

core/ts/Poll.ts
ctrlc03: yup so each message has its corresponding encPubKey - we take the messages and keys part of the batch we are currently processing

core/ts/Poll.ts
ctrlc03: yup will fix

core/ts/Poll.ts
ctrlc03: let's add it in the code comments too?

samajammin: Nice, thanks

core/ts/Poll.ts
```diff
@@ -121,33 +122,29 @@ export class Poll implements IPoll {
...
-    this.duration = duration;
     this.pollEndTimestamp = pollEndTimestamp;
     this.coordinatorKeypair = coordinatorKeypair;
     this.treeDepths = treeDepths;
     this.batchSizes = batchSizes;
     this.maxValues = maxValues;
     this.maciStateRef = maciStateRef;
     this.pollId = maciStateRef.polls.length;
-    this.numSignUps = Number(maciStateRef.numSignUps.toString());
     this.stateTreeDepth = stateTreeDepth;
```
samajammin, line 132: I wonder why `stateTreeDepth` is needed in `Poll`?

Seems we could remove like you removed `numSignUps`, given we can just grab that from `maciStateRef` when needed?

ctrlc03: good shout

core/ts/Poll.ts
```diff
@@ -161,18 +158,16 @@ export class Poll implements IPoll {
     this.stateTree = this.maciStateRef.stateTree.copy();
 
     // Create as many ballots as state leaves
-    const emptyBallot = new Ballot(this.maxValues.maxVoteOptions, this.treeDepths.voteOptionTreeDepth);
-    const emptyBallotHash = emptyBallot.hash();
-    this.ballotTree = new IncrementalQuinTree(this.stateTreeDepth, emptyBallot.hash(), STATE_TREE_ARITY, hash5);
-    this.ballotTree.insert(emptyBallotHash);
+    this.emptyBallotHash = this.emptyBallot.hash();
+    this.ballotTree = new IncrementalQuinTree(this.stateTreeDepth, this.emptyBallotHash, STATE_TREE_ARITY, hash5);
```
samajammin, line 162: Ok dumb question - why the need for a `ballotTree`? I thought there was just the state tree & message tree?

ctrlc03: The ballot tree, which will have the same amount of leaves as the state tree (1 Ballot <> 1 Signup), holds the computed ballot for a given MACI key. When we go and process messages, we update a voter's ballot depending on the content of the message processed, we hash it and include (replace) as leaf in the ballot tree.

ctrlc03: a public key which never casted a vote will have a empty ballot. All these ballots are then used in the tally process to tally the results

samajammin: Ok so a user's ballot is essentially their "final" vote? So it's their final valid message?

samajammin: There's no `ballotTree` in the contract state, correct? This is just for the off-chain processing/tallying of votes?

ctrlc03: Almost - each vote updates the ballot, so the ballot could be the result of 2 valid messages: a message is something like:

```json
{
  nonce: 1,
  voteOption: 1,
  voteWeight: 1,
  pubKey: PubKey1 
}
```

A ballot (at a high level) would be:

```json
[0, 0, 0, 0, 0, 0, 5]
```

Vote zero for option 1 to 6 and vote to option 7

core/ts/Poll.ts
```diff
@@ -181,33 +176,22 @@ export class Poll implements IPoll {
    * @param index - The index of the message to process.
    * @returns A number of variables which will be used in the zk-SNARK circuit.
    */
-  private processMessage = (index: number): IProcessMessagesOutput => {
-    //TODO: throw custom errors for no-ops
-
+  public processMessage = (message: Message, encPubKey: PubKey): IProcessMessagesOutput => {
```
samajammin, line 179: Perhaps doesn't matter but why change this to a public method?

ctrlc03: So we can unit test it

core/ts/Poll.ts
ctrlc03: > There's no `ballotTree` in the contract state, correct? This is just for the off-chain processing/tallying of votes?

we only store a commitment to it (which is the hash of the root of the state treee, the root of the ballot tree and a salt)

ctrlc03: there could not be the same tree on chain because it is the result of the processing of the decrypted messages

core/ts/Poll.ts
samajammin: Gotcha. Is that the same reason you're now passing `encPubKey` into the function?

ctrlc03: Not only, we already extracted the message in the line before, so thought we might as well pass the message we already extracted (to check the msgType), so I added the key as param too:

https://github.com/privacy-scaling-explorations/maci/blob/dev/core/ts/Poll.ts#L451-L455

core/ts/Poll.ts
ctrlc03: updated

core/ts/Poll.ts
ctrlc03: updated

core/ts/MaciState.ts
```diff
@@ -60,16 +64,13 @@ export class MaciState implements IMaciState {
    * @returns The index of the newly deployed poll.
    */
   public deployPoll(
-    duration: number,
```
samajammin, line 63: Why do we no longer need duration? Infer it from `pollEndTimestamp`?

samajammin: Seems like given the `Poll` contract accepts a `duration`, it makes sense to have?

ctrlc03: I removed it cause we never used it anywhere, we only ever used the end timestamp

core/ts/Poll.ts
```diff
@@ -181,33 +175,22 @@ export class Poll implements IPoll {
    * @param index - The index of the message to process.
```
samajammin, line 175: Should update the param comments now that we pass `message` & `encPubKey`

core/ts/Poll.ts
samajammin: Awesome, appreciate the explanations 🙏 😄 

Ah, ok, because each command/message can only include one vote, right?

So e.g. in a QF/QV round, if I submit a ballot casting votes for 8 projects, I'm actually submitting 8 messages?

ctrlc03: Correct!

core/ts/Poll.ts
ctrlc03: done

core/ts/Poll.ts
```diff
@@ -112,42 +114,36 @@ export class Poll implements IPoll {
    * @param stateTreeDepth - The depth of the state tree.
```
kittybest, line 114: The `stateTreeDepth` is no longer a parameter of constructor now, so it should be removed

core/ts/Poll.ts
```diff
@@ -161,53 +157,41 @@ export class Poll implements IPoll {
...
+      this.ballots.push(this.emptyBallot);
     }
 
-    this.numSignUps = Number(this.maciStateRef.numSignUps.toString());
-
     this.stateCopied = true;
   };
 
   /**
    * Process one message at index `index`.
```
kittybest, line 174: It accept `message` and `encPubKey` as parameters now, so the description here should be modified.

core/ts/Poll.ts
```diff
@@ -295,20 +291,24 @@ export class Poll implements IPoll {
...
   public topupMessage = (message: Message): void => {
-    assert(message.msgType == BigInt(2));
+    assert(message.msgType == BigInt(2), "A Topup message must have msgType 2");
+
     for (const d of message.data) {
-      assert((d as bigint) < SNARK_FIELD_SIZE);
+      assert(d < SNARK_FIELD_SIZE, "The message data is not in the correct range");
     }
-    this.messages.push(message);
     const padKey = new PubKey([
```
kittybest, line 299: I'm curious about why we  need `padKey` here and how are they chosen?

core/ts/Poll.ts
```diff
@@ -399,7 +404,7 @@ export class Poll implements IPoll {
 
     // Only allow one poll to be processed at a time
     if (this.maciStateRef.pollBeingProcessed) {
-      assert(this.maciStateRef.currentPollBeingProcessed === pollId);
+      assert(this.maciStateRef.currentPollBeingProcessed === pollId, "Another poll is currently being processed");
     }
 
     if (this.numBatchesProcessed === 0) {
```
kittybest, line 410: Thinking about if `r===0` --> `this.messages.length` is multiple of `batchSize` --> then doing `Math.floor(this.messages.length / batchSize) * batchSize` actually equal to `this.messages.length`?

core/ts/Poll.ts
```diff
@@ -423,159 +428,201 @@ export class Poll implements IPoll {
...
 
+    // we want to store the vote weights at this point in time
+    // and the path elements of the vote weight tree
     const currentVoteWeights: bigint[] = [];
-    const currentVoteWeightsPathElements: any[] = [];
+    const currentVoteWeightsPathElements: PathElements[] = [];
 
+    // loop through the batch of messages
     for (let i = 0; i < batchSize; i++) {
+      // we process the messages in reverse order
```
kittybest, line 462: Just curious about why we process messages in reverse order?

core/ts/Poll.ts
```diff
@@ -423,159 +428,201 @@ export class Poll implements IPoll {
...
+              this.stateLeaves[index] = r.newStateLeaf.copy();
+              // we also update the state tree with the hash of the new state leaf
+              this.stateTree.update(index, r.newStateLeaf.hash());
+              // store the new ballot
+              this.ballots[index] = r.newBallot;
+              // update the ballot tree
+              this.ballotTree.update(index, r.newBallot.hash());
+            } catch (e) {
+              // if the error is not a ProcessMessageError we throw it and exit here
+              // otherwise we continue processing but add the deafult blank data instead of
```
kittybest, line 497: typo: default

core/ts/Poll.ts
```diff
@@ -423,159 +428,201 @@ export class Poll implements IPoll {
...
     circuitInputs.currentBallots = currentBallots.map((x) => x.asCircuitInputs());
     circuitInputs.currentBallotsPathElements = currentBallotsPathElements;
     circuitInputs.currentVoteWeights = currentVoteWeights;
     circuitInputs.currentVoteWeightsPathElements = currentVoteWeightsPathElements;
 
+    // record that we processed one batch
     this.numBatchesProcessed++;
 
     if (this.currentMessageBatchIndex > 0) {
       this.currentMessageBatchIndex -= batchSize;
```
kittybest, line 603: Do we need to prevent `this.currentMessageBatchIndex` from being negative? or it won't happen?

core/ts/Poll.ts
```diff
@@ -969,12 +1028,12 @@ export class Poll implements IPoll {
 
     //debugger
```
kittybest, line 1029: should we remove this line?

core/ts/Poll.ts
ctrlc03: https://github.com/privacy-scaling-explorations/maci/blob/dev/contracts/contracts/crypto/SnarkConstants.sol#L12-L23

here is explained how it's chosen. I'll have to dig a bit deeper as to why it's needed, my immediate guess is that we use it as we need to pass in a public key when publishing a topup message (which is not encrypted), and same for when we push a zero leaf in the message tree on chain

core/ts/Poll.ts
ctrlc03: answered on discord but if someone else is curious here are some places where to look for this answer: 

* https://maci.pse.dev/docs/key-change
* https://github.com/privacy-scaling-explorations/maci/issues/415
* https://maci.pse.dev/docs/spec#61-message-processing-circuit

core/ts/Poll.ts
ctrlc03: that's a great catch, think we can just remove that altogether

core/ts/Poll.ts
ctrlc03: we have this check before: `assert(this.currentMessageBatchIndex % batchSize === 0` -> so it means that it's either the same as `batchSize` or a multiple of it, thus when we do `this.currentMessageBatchIndex -= batchSize;` it can never be < 0

kittybest (approved):

PR #919, opened 2023-12-15
refactor(contracts): smart contracts optimizations

ctrlc03:
# Description

- [x] use calldata vs memory for immutable function params
- [x] add payable to constructors to reduce deployment cost
- [x] unify TREE ARITY as we use quinary trees everywhere
- [x] ensure we do not accept max messages + 1 in top up 

## Related issue(s)

#843 

## Confirmation

### DIFF
--- a/contracts/contracts/MACI.sol
+++ b/contracts/contracts/MACI.sol
@@ -18,7 +18,7 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 
 /// @title MACI - Minimum Anti-Collusion Infrastructure Version 1
 /// @notice A contract which allows users to sign up, and deploy new polls
-contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
+contract MACI is IMACI, Params, Utilities, Ownable {
   /// @notice The state tree depth is fixed. As such it should be as large as feasible
   /// so that there can be as many users as possible.  i.e. 5 ** 10 = 9765625
   /// this should also match the parameter of the circom circuits.
@@ -28,8 +28,7 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
   /// in contracts/ts/genEmptyBallotRootsContract.ts file
   /// if we change the state tree depth!
   uint8 internal constant STATE_TREE_SUBDEPTH = 2;
-  uint8 internal constant STATE_TREE_ARITY = 5;
-  uint8 internal constant MESSAGE_TREE_ARITY = 5;
+  uint8 internal constant TREE_ARITY = 5;
 
   /// @notice The hash of a blank state leaf
   uint256 internal constant BLANK_STATE_LEAF_HASH =
@@ -78,6 +77,7 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
     _;
   }
 
+  /// @notice custom errors
   error CallerMustBePoll(address _caller);
   error PoseidonHashLibrariesNotLinked();
   error TooManySignups();
@@ -96,7 +96,7 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
     InitialVoiceCreditProxy _initialVoiceCreditProxy,
     TopupCredit _topupCredit,
     uint8 _stateTreeDepth
-  ) {
+  ) payable {
     // Deploy the state AccQueue
     stateAq = new AccQueueQuinaryBlankSl(STATE_TREE_SUBDEPTH);
     stateAq.enqueue(BLANK_STATE_LEAF_HASH);
@@ -137,14 +137,15 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
     bytes memory _initialVoiceCreditProxyData
   ) public {
     // ensure we do not have more signups than what the circuits support
-    if (numSignUps == uint256(STATE_TREE_ARITY) ** uint256(stateTreeDepth)) revert TooManySignups();
+    if (numSignUps == uint256(TREE_ARITY) ** uint256(stateTreeDepth)) revert TooManySignups();
 
     if (_pubKey.x >= SNARK_SCALAR_FIELD || _pubKey.y >= SNARK_SCALAR_FIELD) {
       revert MaciPubKeyLargerThanSnarkFieldSize();
     }
 
     // Increment the number of signups
-    // cannot overflow as numSignUps < 5 ** 10 -1
+    // cannot overflow with realistic stateTreeDepth
+    // values as numSignUps < 5 ** stateTreeDepth -1
     unchecked {
       numSignUps++;
     }
@@ -176,6 +177,7 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
     TreeDepths memory _treeDepths,
     PubKey memory _coordinatorPubKey
   ) public onlyOwner returns (address pollAddr) {
+    // cache the poll to a local variable so we can increment it
     uint256 pollId = nextPollId;
 
     // Increment the poll ID for the next poll
@@ -190,9 +192,9 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
 
     // The message batch size and the tally batch size
     BatchSizes memory batchSizes = BatchSizes(
-      uint24(MESSAGE_TREE_ARITY) ** _treeDepths.messageTreeSubDepth,
-      uint24(STATE_TREE_ARITY) ** _treeDepths.intStateTreeDepth,
-      uint24(STATE_TREE_ARITY) ** _treeDepths.intStateTreeDepth
+      uint24(TREE_ARITY) ** _treeDepths.messageTreeSubDepth,
+      uint24(TREE_ARITY) ** _treeDepths.intStateTreeDepth,
+      uint24(TREE_ARITY) ** _treeDepths.intStateTreeDepth
     );
 
     Poll p = pollFactory.deploy(
--- a/contracts/contracts/Poll.sol
+++ b/contracts/contracts/Poll.sol
@@ -49,20 +49,11 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable, EmptyBallotRoots {
 
   uint256 public numMessages;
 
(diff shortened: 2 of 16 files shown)


contracts/contracts/MACI.sol
```diff
@@ -133,14 +133,15 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
     bytes memory _initialVoiceCreditProxyData
   ) public {
     // ensure we do not have more signups than what the circuits support
-    if (numSignUps == uint256(STATE_TREE_ARITY) ** uint256(stateTreeDepth)) revert TooManySignups();
+    if (numSignUps == uint256(TREE_ARITY) ** uint256(stateTreeDepth)) revert TooManySignups();
```
kittybest, line 136: Should we do `>=` here?

contracts/contracts/MACI.sol
```diff
@@ -98,7 +98,7 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
     InitialVoiceCreditProxy _initialVoiceCreditProxy,
     TopupCredit _topupCredit,
     uint8 _stateTreeDepth
-  ) {
+  ) payable {
```
kittybest, line 101: Sorry, not quite sure why we need payable for the constructor?

contracts/contracts/Poll.sol
```diff
@@ -151,7 +142,8 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable, EmptyBallotRoots {
   /// @param stateIndex The index of user in the state queue
   /// @param amount The amount of credits to topup
   function topup(uint256 stateIndex, uint256 amount) public isWithinVotingDeadline {
-    if (numMessages > maxValues.maxMessages) revert TooManyMessages();
+    // we check that we do not exceed the max number of messages
+    if (numMessages == maxValues.maxMessages) revert TooManyMessages();
```
kittybest, line 146: same here, should it be `>=` here?

contracts/contracts/Poll.sol
```diff
@@ -173,9 +165,11 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable, EmptyBallotRoots {
   /// @param _encPubKey An epheremal public key which can be combined with the
   /// coordinator's private key to generate an ECDH shared key with which
   /// to encrypt the message.
-  function publishMessage(Message memory _message, PubKey memory _encPubKey) public isWithinVotingDeadline {
+  function publishMessage(Message memory _message, PubKey calldata _encPubKey) public isWithinVotingDeadline {
+    // we check that we do not exceed the max number of messages
     if (numMessages == maxValues.maxMessages) revert TooManyMessages();
```
kittybest, line 170: same here, `>=`?

contracts/contracts/Subsidy.sol
```diff
@@ -16,13 +16,16 @@ import { VkRegistry } from "./VkRegistry.sol";
...
+  uint256 public rbi;
+  // column batch index
+  uint256 public cbi;
+
   // The final commitment to the state and ballot roots
   uint256 public sbCommitment;
   uint256 public subsidyCommitment;
 
-  uint8 public constant treeArity = 5;
+  uint8 internal constant TREE_ARITY = 5;
```
kittybest, line 28: Thinking about that `TREE_ARITY` is called so many times in different contracts, and they should keep the same. Should we put it in somewhere and import it in these contracts?

contracts/contracts/gatekeepers/FreeForAllSignUpGatekeeper.sol
```diff
@@ -7,6 +7,9 @@ import { MACI } from "../MACI.sol";
 /// @title FreeForAllGatekeeper
 /// @notice A SignUpGatekeeper which allows anyone to sign up.
 contract FreeForAllGatekeeper is SignUpGatekeeper {
+  /// @notice Create a new instance of FreeForAllGatekeeper
+  constructor() payable {}
```
kittybest, line 11: Why we need to specify the payable with empty constructor here?

contracts/contracts/MACI.sol
ctrlc03: Allows to save a tiny bit of gas while deploying a contract: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3059

contracts/contracts/MACI.sol
ctrlc03: Would you mind expanding why? 
I have been thinking `numSignups` nor `numMessages` can never be more than what we support in the circuits (`uint256(TREE_ARITY) ** uint256(stateTreeDepth)` and `maxValues.maxMessages`) so thought that `==`as condition would be enough, but maybe I'm missing something?

kittybest: I agree that it couldn't be more than the max values, and if they surpass the max value, they must be equal to the max values first (since they are added 1 once a time). But we also don't want it to become over the max value? To prevent maybe user accidentally modify it somewhere else to a very big value, that's why I suggest to change it to `>=`. I'm also curious why we only use `==`? Is it because of the gas fee? 
Maybe I'm thinking too much. If the condition I mentioned above will never happen, just ignore this recommendation.

contracts/contracts/MACI.sol
kittybest: This is a cool feature to me, since I searched and also asked ChapGPT if we could make constructor payable and find nothing 🤔 Thanks for the information~~

contracts/contracts/MACI.sol
ctrlc03: There are no instances where these variables are increased without first checking that we have not reached the limit, aside from the `init` function of Poll. In terms of gas costs these would increase slightly, but this is a very small value, don't think this was done because of gas costs (`eq` opcode vs `LT` and `ISZERO` so 3 gas vs 6 gas)

Having said that, I just saw that we have a problem [here](https://github.com/privacy-scaling-explorations/maci/blob/dev/contracts/contracts/MACI.sol#L136) we can have 1 signup more than what we support in the circuits, because in the constructor we add the [`BLANK_STATE_LEAF_HASH`](https://github.com/privacy-scaling-explorations/maci/blob/dev/contracts/contracts/MACI.sol#L104)

What you suggest here, we increase `numSignups` in constructor, or add a -1 to the check `numSignUps - 1 == uint256(STATE_TREE_ARITY) ** uint256(stateTreeDepth)`?

contracts/contracts/Subsidy.sol
ctrlc03: mm maybe we could put it inside another contract shared by all of them though I think that would bloat things up a little bit 🤔 was looking and couldn't find a contract that was already in use by all of them (MACI, Poll, MP, Tally, etc) where we could add this constant..

contracts/contracts/MACI.sol
ctrlc03: no worries 😄

contracts/contracts/MACI.sol
kittybest: Actually I just simply suggest to check `numSignUps >= uint256(TREE_ARITY) ** uint256(stateTreeDepth)`, just in case there's any condition that the numSignUps would be larger than the max value.

contracts/contracts/Subsidy.sol
kittybest: Is that a good idea to have a contract holding the constants? or we just keep status quo?

ctrlc03: mm we have `SnarkConstants` holding some constants, though I'm not sure if it's a widely used pattern.. to avoid bloating contracts even more I would just keep this one constant inside each contract, at least for now

contracts/contracts/MACI.sol
ctrlc03: Even if based on our current code it cannot happen? 

ps. I was also referring to our  MACI constructor where we enqueue one item inside the `stateAq` but do not increase `numSignups` - I have added a line to add +1 to `numSignups` after the first enqueue to avoid the possibility of `numSignups` being off by one

kittybest: In my opinion, yes, to avoid any accidence... but if you think this is unnecessary, we could keep it as it is now.

contracts/contracts/Subsidy.sol
kittybest: Then when it comes to deployment, if a user wants to alter the TREE_ARITY, they currently need to navigate through every contract file to modify the value. What if we convert it into an environmental variable for the sake of convenience during deployment?

ctrlc03: Fair point, though they would also need to change the type of `AccQueue` in use, which currently is hardcoded to a quinary tree everywhere, mm 🤔

ctrlc03: useful discussion here, there's definitely some more simplifying we can do given we always use quinary trees

contracts/contracts/MACI.sol
ctrlc03: I currently believe it is unnecessary, given I see no way to actually go > the limit, but if you see a way that this could go over the limit, then 100% we should add it, and prevent it from ever going over in the first place. What do you mean by accident here?

contracts/contracts/Subsidy.sol
ctrlc03: by converting to env variable, you mean that it should be a value passed in the constructor when deploying the contracts?

contracts/contracts/MACI.sol
kittybest: I don't know, just in case? but if you think it's unnecessary, then just ignore it~~

contracts/contracts/Subsidy.sol
kittybest: yes... but not sure if this is a good idea, just wanna reduce the risk that if some users deploy it and miss to change one of them?

ctrlc03: let's write this down and discuss together at the next team meeting? from next week we'll have a smart contract developer to ask all SC questions to 😈

kittybest (approved):

PR #925, opened 2023-12-15
feat(proofs): add preferred witness type in circuit's genProof function

ctrlc03:
# Description

Remove `isArm` check in circuits and instead accept a parameter to decide which witness type to use. Made appropriate changes in tests and cli. 

This allows to use wasm witness in an intel chip as well should this be preferred.

## Confirmation

### DIFF
--- a/circuits/ts/index.ts
+++ b/circuits/ts/index.ts
@@ -1,2 +1,2 @@
 export { genProof, verifyProof, extractVk } from "./proofs";
-export { isArm, cleanThreads } from "./utils";
+export { cleanThreads } from "./utils";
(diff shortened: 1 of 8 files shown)


kittybest:
The code lgtm.
I believe the objective of this pull request is to enable wasm support on Intel chips, if users prefer. However, considering that most users might simply want the code to be executable, I propose retaining the `isArm()` check in `genProof`. Additionally, we can introduce `useWasm` as a parameter for those interested in experimenting with options beyond the default configuration.
If I misunderstood it, please tell me~~

ctrlc03:
> The code lgtm. I believe the objective of this pull request is to enable wasm support on Intel chips, if users prefer. However, considering that most users might simply want the code to be executable, I propose retaining the `isArm()` check in `genProof`. Additionally, we can introduce `useWasm` as a parameter for those interested in experimenting with options beyond the default configuration. If I misunderstood it, please tell me~~

mm I like this thinking, it would definitely make it easier for users to use. We could make the flag `useWasm` optional, and then add some more checks, I'll do that now

kittybest (approved):

PR #926, opened 2023-12-15
test(crypto): add a test to verify that the keypair generation does not fail

### LINKED ISSUE #239: Generate keypair: "TypeError: Invalid hex string"
Sometimes when I call `new Keypair()` it fails with `TypeError: Invalid hex string` error. It started to happen after upgrading to 0.7.1.
The error occurs at this line: https://github.com/appliedzkp/maci/blob/0a4f7cc0e355a191cc69f7b9fedb4440370a2abb/crypto/ts/index.ts#L56 

I tried to debug it and noticed that `Buffer.from` throws type error when `i.toString(16)` is a 63-character string, for example:

```
18582ee20a67742c1a065b14b057dd78f79e9f885518a347e7956a7e6a442e1
```

but it doesn't throw error if `i.toString(16)` consists of 64 characters:

```
177df0660a150cdeb60d19f9383dfeb664bdd681ff2a1f990d4d84efb8347792
```

The error occurs only in browser environment, I can't reproduce it with node.js.

ctrlc03:
# Description

Add a test case which attemps to generate a new MACI keypair 500 times in a row to check whether issue #239 still applies to the current codebase.

TODO: 

- [ ] test in a browser app

## Related issue(s)

fix #239 

## Confirmation

### DIFF
--- a/crypto/ts/__tests__/Crypto.test.ts
+++ b/crypto/ts/__tests__/Crypto.test.ts
@@ -195,4 +195,15 @@ describe("Cryptographic operations", function test() {
       expect(commitment).to.eq(commitment2);
     });
   });
+
+  describe("genKeypair", () => {
+    it("should consistently generate a valid keypair (500 runs)", () => {
+      for (let i = 0; i < 500; i += 1) {
+        const key = genKeypair();
+        expect(key.privKey < SNARK_FIELD_SIZE).to.eq(true);
+        expect(key.pubKey[0] < SNARK_FIELD_SIZE).to.eq(true);
+        expect(key.pubKey[1] < SNARK_FIELD_SIZE).to.eq(true);
+      }
+    });
+  });
 });


baumstern:
hmm I've noticed that the error occurred recently in the browser environment. However, it was fine when using the Node.js runtime🤔

ctrlc03:
> hmm I've noticed that the error occurred recently in the browser environment. However, it was fine when using the Node.js runtime🤔

ohhhh I see, thanks for the context. Will try it in a react app

crypto/ts/__tests__/Crypto.test.ts
```diff
@@ -195,4 +195,15 @@ describe("Cryptographic operations", function test() {
...
+  describe("genKeypair", () => {
+    it("should consistently generate a valid keypair (500 runs)", () => {
+      for (let i = 0; i < 500; i += 1) {
+        const key = genKeypair();
+        expect(key.privKey < SNARK_FIELD_SIZE).to.eq(true);
+        expect(key.pubKey[0] < SNARK_FIELD_SIZE).to.eq(true);
+        expect(key.pubKey[1] < SNARK_FIELD_SIZE).to.eq(true);
+      }
+    });
+  });
```
0xmad, line 208: I suggest using here https://github.com/dubzzz/fast-check. Property based testing approach will be a good fit.

ctrlc03: This looks cool, would definitely be useful for other parts of our suites!!!

ctrlc03: I think I might remove this test case here though, as the issue seems to be with environments !== node 🤔

PR #932, opened 2023-12-18
test(coverage): setup coverage for TS packages

### LINKED ISSUE #931: Setup code coverage (TS)
As developers we should ensure that code is properly tested. In order to determine the percentage of code tested, we can use a code coverage tool such as [nyc[(https://istanbul.js.org/) which is often used with mocha.

Code coverage works by instrumenting the code to add tracing calls and then monitoring which parts of the code are executed when the test suite runs. The results are usually reported as a percentage, with 100% meaning that all lines of code are covered by the tests.

ctrlc03:
# Description

setup code coverage for TS packages using [istanbul (nyc)](https://istanbul.js.org/)

## Related issue(s)

fix #931 

## Confirmation

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -27,6 +27,7 @@
         "@types/node": "^18.11.9",
         "chai": "^4.3.10",
         "mocha": "^10.2.0",
+        "nyc": "^15.1.0",
         "ts-mocha": "^10.0.0",
         "typescript": "^5.3.2"
       }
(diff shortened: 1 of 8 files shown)


kittybest:
Will we also add `nyc` to `circuits` package? I think we're also using `mocha` to run tests in that package.
Also will we use another tool to test the code coverage for smart contracts? I think there are some packages such as **solidity-coverage**.

ctrlc03:
> Will we also add `nyc` to `circuits` package? I think we're also using `mocha` to run tests in that package.
> 
> Also will we use another tool to test the code coverage for smart contracts? I think there are some packages such as **solidity-coverage**.

For the circuits package, it depends if we want to add unit tests for the ts code that is there (might be worth doing). If it's just for the circuits, then I'm afraid it wont really work with circom code, and I don't really know of any coverage tool for circom code unfortunately, we should do some research there.

Yes for contract there's that one, tho I tried to set it up and was having some issues with contract size due to the tool bloating contracts for collecting coverage, so I thought it would be best if we were to tackle that in a separate PR

kittybest (approved):

PR #938, opened 2023-12-20
test(core): remove msgTree

samajammin:
# Description

Remove msgTree variable

## Confirmation

### DIFF
--- a/core/ts/__tests__/e2e.test.ts
+++ b/core/ts/__tests__/e2e.test.ts
@@ -356,7 +356,6 @@ describe("MaciState/Poll e2e", function test() {
     let maciState: MaciState;
     let pollId: bigint;
     let poll: Poll;
-    let msgTree: IncrementalQuinTree;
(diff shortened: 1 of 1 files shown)


samajammin:
@ctrlc03 this is mostly for my understanding...

Why are we creating this `msgTree` object? Don't we want to be comparing the state of the `poll` object in this test?

ctrlc03:
> Member

@samajammin Great catch, you are absolutely right 🙂 ps. I have been thinking that we could probably remove the `AccQueue` instances throughout the core package tests and keep the `IncrementalQuinTree` instead (as that's what we use inside the Poll contract for storing message hashes) - what you think? if you like the idea maybe this PR could be extended to do that

daodesigner:
That was meant to be an integration test between maci-domainobjs and maci-crypto 🙃 careful making changes to things you don't understand!

ctrlc03 (approved):

Issue #772, opened 2023-10-23
Poseidon encryption/decryption support


ctrlc03:
Currently, we are using a [fork](https://github.com/weijiekoh/circomlib) of circomlib which offers the utilities for poseidon encryption and decryption. This is a core component of MACI, however this results in us using generally outdated circomlib(js) code. 

Unfortunately this [PR](https://github.com/iden3/circomlib/pull/60) was never merged (and should actually be moved to circomlibjs), thus we need to think of a more future-proof way of using this code while having access to the most up to date circomlib/circomlibjs code. 

Further issue: seems like when changing from BigInt to bigint in packages importing this fork, certain actions fail with this error: "Cannot find module 'big-integer' from '../domainobjs/node_modules/circomlibjs/src/smt_hashes_mimc.js'"

Options:

* add encryption/decryption code to MACI repo
* fork circomlibjs and submit PR again (or ask the original author as that's their code)
* keep using this fork only for poseidon encryption and use instead the latest version of circomlibjs for all other needs (definitely simpler option)

PR #943, opened 2023-12-30
refactor(zk-kit-poseidon): remove old circomlib dependencies in crypto

ctrlc03:
# Description

This PR removes an old import (ciromlib) from the crypto library and instead imports the same functions from [zk-kit](https://github.com/privacy-scaling-explorations/zk-kit). zk-kit provides lightweight TypeScript libraries for zk developers, allowing MACI's packages to be slimmer and more up to date.

## Additional Notes

This is the first step into removing an old [dependency](https://github.com/weijiekoh/circomlib), currently we still use it inside the circom circuits, and will be tackled in another PR.

@cedoor MACI 🤝 zk-kit 

## Related issue(s)

Re: #772 

## Additional Info

Let's merge #944 first 

## Confirmation


crypto/ts/types.ts
```diff
@@ -1,10 +1,24 @@
-export type SnarkBigInt = bigint;
-export type PrivKey = bigint;
-export type PubKey = bigint[];
-export type Point = bigint[];
-export type EcdhSharedKey = bigint[];
-export type Plaintext = bigint[];
-export type Ciphertext = bigint[];
+// we define a bignumber as either a bigint or a string
+// which is what we use the most in MACI
+export type BigNumber = bigint | string;
```
0xmad, line 3: I think this type can have conflicts if it's using with BigNumber from ethers of from bignumber.js. Maybe it makes sense to reuse BigNumberish type from ethers?

0xmad (approved):
@ctrlc03 thanks, just one comment/question.

crypto/ts/types.ts
ctrlc03: Good shout, I can use that instead

ctrlc03: What do you think about simply changing the type name? so we don't import from ethers and make further changes to the code

0xmad: I think it works as well.

Issue #793, opened 2023-10-30
Dependency refresh


ctrlc03:
Before release perform the following:

- [x] npm audit 
- [x] upgrade as much as possible to newest version of the same library 
- [x] remove any unnecessary dependency to keep packages slim

baumstern:
fyi: NPM audit has addressed in https://github.com/privacy-scaling-explorations/maci/pull/806

ctrlc03:
Closing as #965 #972 #963 #944 #806 and all other refactor work resolved this.

PR #944, opened 2023-12-30
refactor(crypto): remove ffjavascript and implement utils functions locally

### LINKED ISSUE #458: Enforce maci private key length to specific size
Both following private keys has different length to each other and are valid key:
macisk.2044549fd073daf95bfed48ab0583a6cc5cea0f45e11885a2670e712409d739 (63 character long)
macisk.18843ce600faa1656c6d1fed25b08b2c39a9aa3081d3b12eccc162e59d576ba4 (64 character long)

ctrlc03:
# Description

Remove the large dependency `ffjavascript` which was used only for a couple of util functions. Also ensure that we safely convert a bigint to a buffer and thus result in serialized maci private keys being the same length.

## Related issue(s)

Related to #793 and fix #458

## Confirmation


crypto/ts/__tests__/Utils.test.ts
```diff
@@ -0,0 +1,158 @@
...
+import { bigInt2Buffer, stringifyBigInts, unstringifyBigInts } from "../bigIntUtils";
+import { SNARK_FIELD_SIZE } from "../constants";
+import { genTreeCommitment } from "../utils";
+
+describe("Utils", () => {
+  describe("stringifyBigInts", () => {
+    it("Should work on a BigInt input", () => {
+      expect(stringifyBigInts(BigInt(1))).to.eq("1");
+    });
+    it("Should work on a BigInt[] input", () => {
```
0xmad, line 12: Let's add new lines between tests.

crypto/ts/bigIntUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+      return null;
+    }
+    const res: Record<string, bigint> = {};
+    const keys = Object.keys(input);
+    keys.forEach((k) => {
+      const value = unstringifyBigInts(input[k]);
+      res[k] = value as bigint;
+    });
+    return res;
+  }
```
0xmad, line 33:
```suggestion
  if (input === null) {
    return null;
  }

  if (typeof input === "object") {
    return Object.entries(input).reduce<Record<string, bigint>>((acc, [key, value]) => {
      acc[key] = unstringifyBigInts(value) as bigint;
      return acc;
   }, {}); 
  }
```

crypto/ts/bigIntUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+      return null;
+    }
+    const res: Record<string, StringifiedBigInts> = {};
+    const keys = Object.keys(input);
+    keys.forEach((k) => {
+      const value = stringifyBigInts(input[k]);
+      res[k] = value;
+    });
+    return res;
+  }
```
0xmad, line 104:
```suggestion
  if (input === null) {
    return null;
  }
 
  if (typeof input === "object") {
    return Object.entries(input).reduce<Record<string, StringifiedBigInts>>((acc, [key, value]) => {
      acc[key] = stringifyBigInts(value);
      return acc;
   }, {}); 
  }
```

0xmad (changes requested):
@ctrlc03 thanks, just few changes to improve

PR #946, opened 2024-01-02
chore(core): integrate linter

0xmad:
# Description

- [x] Apply linter and type fixes
- [x] Dependency management

## Additional Notes

N/A

## Related issue(s)

Related to #845 

## Confirmation


core/ts/Poll.ts
```diff
@@ -28,8 +27,13 @@ import {
   IJsonTCommand,
 } from "maci-domainobjs";
 
-import { MaciState } from "./MaciState";
-import { packTallyVotesSmallVals, packSubsidySmallVals } from "./utils/utils";
+import assert from "assert";
+
+import type { MaciState } from "./MaciState";
+import type { PathElements } from "optimisedmt";
```
ctrlc03, line 33: I can see this is the only import from `optimisedmt`, given we already import `IncrementalQuinTree` from `crypto` (which is exported from `optimisedmt`) should we maybe add one extra export in `crypto` so we don't have to add `optimisedmt` as dependency here in `core`? if you notice all over the codebase we always import `IncrementalQuinTree` directly from `crypto` and not `optimisedmt`

ctrlc03 (changes requested):
Great stuff thanks, left one comment

ctrlc03 (approved):
Thank you!! LGTM

PR #948, opened 2024-01-03
fix(signups): ensure we account for the blank state leaf

### LINKED ISSUE #947: A blank state leaf is not counted as 1 signup
Currently, we do not count the `BLANK_STATE_LEAF` added to the `stateAq` upon MACI's contract deployment as one signup. This could cause the tree to be filled with 1 extra leaf than supported by the circuits. 

Furthermore, we wrongly calculate the number of batches in the [`proveOnChain`](https://github.com/privacy-scaling-explorations/maci/blob/dev/cli/ts/commands/proveOnChain.ts#L376) cli command -> if we have 10 signups and the batch size is 5 -> we end up checking proofs for 3 batches instead of 2

ctrlc03:
# Description

currently the blank state leaf is not counted as one signup, and thus we could go over the supported amount of signups (based on the circuits parameters) by 1, potentially preventing one voters' votes from being tallied
## Related issue(s)

fix #947

## Confirmation

### DIFF
--- a/contracts/contracts/MACI.sol
+++ b/contracts/contracts/MACI.sol
@@ -101,6 +101,12 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
     stateAq = new AccQueueQuinaryBlankSl(STATE_TREE_SUBDEPTH);
     stateAq.enqueue(BLANK_STATE_LEAF_HASH);
 
+    // because we add a blank leaf we need to count one signup
+    // so we don't allow max + 1
+    unchecked {
+      numSignUps++;
+    }
+
     pollFactory = _pollFactory;
     topupCredit = _topupCredit;
     signUpGatekeeper = _signUpGatekeeper;
--- a/cli/ts/commands/proveOnChain.ts
+++ b/cli/ts/commands/proveOnChain.ts
@@ -303,12 +303,11 @@ export const proveOnChain = async (
   if (Object.keys(data.subsidyProofs).length !== 0) {
     let rbi = Number(await subsidyContract.rbi());
     let cbi = Number(await subsidyContract.cbi());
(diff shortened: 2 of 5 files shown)


cli/tests/e2e.test.ts
```diff
@@ -156,7 +156,7 @@ describe("e2e tests", function () {
     });
   });
 
-  describe("test2", () => {
+  describe.only("test2", () => {
```
0xmad, line 159: Let's remove this

ctrlc03: oh yea my bad I forgot that

0xmad (approved):
@ctrlc03 thanks!

contracts/contracts/MACI.sol
```diff
@@ -101,6 +101,12 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable {
     stateAq = new AccQueueQuinaryBlankSl(STATE_TREE_SUBDEPTH);
     stateAq.enqueue(BLANK_STATE_LEAF_HASH);
 
+    // because we add a blank leaf we need to count one signup
+    // so we don't allow max + 1
+    unchecked {
+      numSignUps++;
+    }
```
baumstern, line 108: Shouldn't users expect numSignUps to reflect the actual number of signups without having to adjust for the BLANK_STATE_LEAF by subtracting one manually? This could lead to misleading data representation

samajammin: Fair point, although I think it depends on what you mean by "users" & options of how to represent that data to users. For instance, a frontend integration could easily account for this by adjusting what the data they display in the application

An alternative might be to not count the blank leaf as a signup, but to restrict the total number of signups by 1

PR #961, opened 2024-01-08
chore: bundle size optimizations

### LINKED ISSUE #781: Treeshaking not working with maci library packages
We noticed that by importing just 1 function (genTallyResultCommitment) from the maci-core library, our web app size would double.  Duplicating that function (as opposed to importing from maci-core) would bring the web app size back down.

Here are the sizes of our web app:

(import from maci-core)
7,567.22 kB │ gzip: 3,538.62 kB

(duplicate function locally) 
3,669.37 kB │ gzip: 1,726.62 kB

I tried different rollup treeshaking options on our vite.config file, without success.
I tried adding "sideEffects: false" manually in the maci-core package in the local node_modules folder and still not working.

While working on using the snarkjs API, I stumbled onto this issue (https://github.com/iden3/snarkjs/issues/152) which shows that ffjavascript library that maci-crypto depends on has side effects that might be causing treeshaking to not work. 

For optimization, ffjavascript uses a global (globalThis variable) to pass back a handle for the consumer to terminate the threads.

Note that maci-crypto only uses the utils functions in (ffjavascript)[https://github.com/privacy-scaling-explorations/maci/blob/dev/crypto/ts/index.ts#L9), may be duplicating those functions instead of `require('ffjavascript`) could reduce the bundle size of app?

Also, moving the `genTallyResultCommitment` from maci-core to maci-crypto may help as we only need to import maci-crypto.

We use all the hash and encrypt/decrypt functions in maci-crypto to generate key and encrypt/decrypt messages on our web app.

 Appreciate any suggestion on what to do to enable treeshaking and reduce our web app size.

0xmad:
# Description

- [x] Remove snarkjs from contracts
- [x] Use type prefix for type imports
- [x] Update ethers

## Additional Notes

N/A

## Related issue(s)

Related #781

## Confirmation

### DIFF
--- a/crypto/package.json
+++ b/crypto/package.json
@@ -16,11 +16,10 @@
     "@zk-kit/baby-jubjub": "^0.1.1",
     "@zk-kit/eddsa-poseidon": "^0.5.1",
     "@zk-kit/poseidon-cipher": "^0.1.1",
-    "ethers": "^5.4.7",
+    "ethers": "^6.9.2",
     "optimisedmt": "^0.0.9"
   },
   "devDependencies": {
-    "@babel/eslint-parser": "^7.23.3",
     "@types/chai": "^4.3.11",
     "@types/mocha": "^10.0.6",
     "@types/node": "^18.11.9",
--- a/circuits/ts/proofs.ts
+++ b/circuits/ts/proofs.ts
@@ -1,12 +1,20 @@
 import { stringifyBigInts } from "maci-crypto";
-import { zKey, groth16, FullProveResult, PublicSignals, Groth16Proof, ISnarkJSVerificationKey } from "snarkjs";
+import {
+  zKey,
+  groth16,
+  type FullProveResult,
+  type PublicSignals,
+  type Groth16Proof,
+  type ISnarkJSVerificationKey,
+} from "snarkjs";
 
 import { execSync } from "child_process";
 import fs from "fs";
 import { tmpdir } from "os";
 import path from "path";
 
-import { IGenProofOptions } from "./types";
+import type { IGenProofOptions } from "./types";
+
 import { cleanThreads, isArm } from "./utils";
 
 /**
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -13,7 +13,7 @@
         "circomlib": "^2.0.5",
         "commander": "^11.1.0",
         "dotenv": "^16.3.1",
-        "ethers": "^6.9.0",
+        "ethers": "^6.9.2",
         "hardhat": "^2.19.2",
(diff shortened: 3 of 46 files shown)


0xmad:
@yuetloo please check if current branch has smaller size.
This branch also contains fixes from @ctrlc03 (crypto package refactoring).

ctrlc03 (approved):
LGTM! thank you

yuetloo:
> @yuetloo please check if current branch has smaller size. This branch also contains fixes from @ctrlc03 (crypto package refactoring).

thanks, will take a look

crypto/package.json
```diff
@@ -16,11 +16,10 @@
     "@zk-kit/baby-jubjub": "^0.1.1",
     "@zk-kit/eddsa-poseidon": "^0.5.1",
     "@zk-kit/poseidon-cipher": "^0.1.1",
-    "ethers": "^5.4.7",
+    "ethers": "^6.9.2",
     "optimisedmt": "^0.0.9"
```
yuetloo, line 20: This PR also included ethers v5 to v6 upgrade?

0xmad: Yes, after that it will be the same version across the whole project. Only crypto package didn't have v6.

yuetloo:
Is this branch published on npm as beta or alpha package? What about the develop branch, any npm package?On Jan 9, 2024, at 10:14 AM, Anton ***@***.***> wrote:﻿
@0xmad commented on this pull request.



In crypto/package.json:
> @@ -16,11 +16,10 @@
     ***@***.***/baby-jubjub": "^0.1.1",
     ***@***.***/eddsa-poseidon": "^0.5.1",
     ***@***.***/poseidon-cipher": "^0.1.1",
-    "ethers": "^5.4.7",
+    "ethers": "^6.9.2",
     "optimisedmt": "^0.0.9"

Yes, after that it will be the same version across the whole project. Only crypto package didn't have v6.

0xmad:
@yuetloo no, it's not published but if you clone it and use it as local dependency it should work. Or @ctrlc03 if it's possible to publish this temp version to npm it will be better (I don't have an access for publishing).

ctrlc03:
> @yuetloo no, it's not published but if you clone it and use it as local dependency it should work. Or @ctrlc03 if it's possible to publish this temp version to npm it will be better (I don't have an access for publishing).

we have these which are published at every commit to the dev branch: https://www.npmjs.com/package/maci-crypto?activeTab=versions -> `npm install maci-crypto@ci`

0xmad:
@ctrlc03 @yuetloo ok, let's merge this PR and if there are some comments just mention them in #781

Issue #977, opened 2024-01-10
Nightly run failed


baumstern:
It looks like test runtime can not find a proving key ` ./zkeys/ProcessMessages_10-2-1-2_test.0.zkey`, since [Generate zkeys step has failed](https://github.com/privacy-scaling-explorations/maci/actions/runs/7468746866/job/20324740337): 
```
Written successfully: ./zkeys/SubsidyPerBatch_10-1-2_test_js/SubsidyPerBatch_10-1-2_test.wat
Written successfully: ./zkeys/SubsidyPerBatch_10-1-2_test_js/SubsidyPerBatch_10-1-2_test.wasm
Everything went okay, circom safe
g++ -c main.cpp -std=c++11 -O3 -I.
g++ -c calcwit.cpp -std=c++11 -O3 -I.
g++ -c fr.cpp -std=c++11 -O3 -I.
nasm -felf64 fr.asm -o fr_asm.o
g++ -c SubsidyPerBatch_10-1-2_test.cpp -std=c++11 -O3 -I.
g++ -o SubsidyPerBatch_10-1-2_test *.o -lgmp 
TypeError: Cannot read properties of null (reading '1')
    at getPtauFromConfig (/home/runner/work/maci/maci/cli/node_modules/zkey-manager/build/utils.js:54:29)
    at /home/runner/work/maci/maci/cli/node_modules/zkey-manager/build/genZkeys.js:87:46
    at step (/home/runner/work/maci/maci/cli/node_modules/zkey-manager/build/genZkeys.js:52:23)
    at Object.next (/home/runner/work/maci/maci/cli/node_modules/zkey-manager/build/genZkeys.js:33:53)
    at /home/runner/work/maci/maci/cli/node_modules/zkey-manager/build/genZkeys.js:27:71
    at new Promise (<anonymous>)
    at __awaiter (/home/runner/work/maci/maci/cli/node_modules/zkey-manager/build/genZkeys.js:23:12)
    at genZkeys (/home/runner/work/maci/maci/cli/node_modules/zkey-manager/build/genZkeys.js:79:54)
    at /home/runner/work/maci/maci/cli/node_modules/zkey-manager/build/index.js:107:62
    at step (/home/runner/work/maci/maci/cli/node_modules/zkey-manager/build/index.js:53:23)
```

PR #978, opened 2024-01-10
fix(nightly): add back snarkjs to cli as dev dependency and fix poll duration

ctrlc03:
# Description

There have been some issues in the nightly workflows, which compile circuits and generate zkeys from scratch to ensure the whole process works as expected.

This PR fixes two issues:

* having removed snarkjs from the cli package we cannot use zkey-manager anymore to generate new zkeys - this was added back as a dev dependency
* the nightly workflow with the ceremony params has a very small poll duration set causing it to fail

## Related issue(s)

re #977 

## Confirmation


0xmad (approved):
@ctrlc03 thanks!

PR #1008, opened 2024-01-11
refactor(contracts): optimize code

ctrlc03:
# Description

Optimize code by using immutable variables are removing redundant code.

## Confirmation


website/versioned_docs/version-v1.x/contracts.md
```diff
@@ -31,8 +31,6 @@ constructor(
     signUpGatekeeper = _signUpGatekeeper;
     initialVoiceCreditProxy = _initialVoiceCreditProxy;
 
-    signUpTimestamp = block.timestamp;
```
0xmad, line 34: Is it supposed to be removed?

0xmad (approved):
@ctrlc03 thanks, just one question. Do we need to support the same state for deployed smart contracts? As far as I know, if someone tries to redeploy these contracts using proxies, it won't work because state if not the same.
In case if we don't care about it, I'm good with these changes.

website/versioned_docs/version-v1.x/contracts.md
ctrlc03: yes I removed it from the code as well as it wasn't used anywhere. Let me know if you think we should have kept it in the code

ctrlc03:
> @ctrlc03 thanks, just one question. Do we need to support the same state for deployed smart contracts? As far as I know, if someone tries to redeploy these contracts using proxies, it won't work because state if not the same. In case if we don't care about it, I'm good with these changes.

Could you please expand on that? do you mean because of immutable parameters?

0xmad:
> > @ctrlc03 thanks, just one question. Do we need to support the same state for deployed smart contracts? As far as I know, if someone tries to redeploy these contracts using proxies, it won't work because state if not the same. In case if we don't care about it, I'm good with these changes.
> 
> Could you please expand on that? do you mean because of immutable parameters? if so, there's libraries for proxy clones with immutable args: https://github.com/wighawag/clones-with-immutable-args

I had some issues after adding immutable which causes storage layout change. But if it's not a case, let's leave it.

ctrlc03:
> > > @ctrlc03 thanks, just one question. Do we need to support the same state for deployed smart contracts? As far as I know, if someone tries to redeploy these contracts using proxies, it won't work because state if not the same. In case if we don't care about it, I'm good with these changes.
> > 
> > 
> > Could you please expand on that? do you mean because of immutable parameters? if so, there's libraries for proxy clones with immutable args: https://github.com/wighawag/clones-with-immutable-args
> 
> I had some issues after adding immutable which causes storage layout change. But if it's not a case, let's leave it.

issues when upgrading?

0xmad:
> > > > @ctrlc03 thanks, just one question. Do we need to support the same state for deployed smart contracts? As far as I know, if someone tries to redeploy these contracts using proxies, it won't work because state if not the same. In case if we don't care about it, I'm good with these changes.
> > > 
> > > 
> > > Could you please expand on that? do you mean because of immutable parameters? if so, there's libraries for proxy clones with immutable args: https://github.com/wighawag/clones-with-immutable-args
> > 
> > 
> > I had some issues after adding immutable which causes storage layout change. But if it's not a case, let's leave it.
> 
> issues when upgrading? I've not tried these contracts with any proxy pattern before tbf

Yes, it's related to upgrades

ctrlc03:
> > > > > @ctrlc03 thanks, just one question. Do we need to support the same state for deployed smart contracts? As far as I know, if someone tries to redeploy these contracts using proxies, it won't work because state if not the same. In case if we don't care about it, I'm good with these changes.
> > > > 
> > > > 
> > > > Could you please expand on that? do you mean because of immutable parameters? if so, there's libraries for proxy clones with immutable args: https://github.com/wighawag/clones-with-immutable-args
> > > 
> > > 
> > > I had some issues after adding immutable which causes storage layout change. But if it's not a case, let's leave it.
> > 
> > 
> > issues when upgrading? I've not tried these contracts with any proxy pattern before tbf
> 
> Yes, it's related to upgrades

oh that makes sense then! I don't think MACI as a protocol should use upgradable proxies, to enhance transparency. On the other hand, at some point we should definitely consider using clone proxies (or another efficient method) to deploy instances. What do you think?

0xmad:
@ctrlc03 I think yes, at least we need to know community about that because someone else can use it with proxies and it can be a problem after upgrade with storage layout.
From our side, I think it will be useful to have something which helps to reduce gas for deployment or even better user can just use already deployed contracts.

ctrlc03:
> @ctrlc03 I think yes, at least we need to know community about that because someone else can use it with proxies and it can be a problem after upgrade with storage layout. From our side, I think it will be useful to have something which helps to reduce gas for deployment or even better user can just use already deployed contracts.

Definitely agree! Yes being able to use already deployed contracts and just point a proxy to it would be fantastic. Let's merge this then?

0xmad:
@ctrlc03 sure, thanks!

contracts/contracts/MACI.sol
```diff
@@ -38,31 +38,27 @@ contract MACI is IMACI, Params, Utilities, Ownable {
...
   /// @notice The number of signups
-  uint256 public override numSignUps;
+  uint256 public numSignUps;
 
   /// @notice A mapping of block timestamps to the number of state leaves
   mapping(uint256 => uint256) public numStateLeaves;
 
   // ERC20 contract that hold topup credits
-  TopupCredit public topupCredit;
+  TopupCredit public immutable topupCredit;
```
kittybest, line 47: image

Does this matter? or just leave it?

ctrlc03: I think we can can change at some other point

samajammin: Should we create an issue for this @kittybest @ctrlc03 @0xmad?

samajammin: Created on here: https://github.com/privacy-scaling-explorations/maci/issues/1012

PR #1010, opened 2024-01-12
feat(circuits): remove circomlib fork in favour of latest and zk-kit

### LINKED ISSUE #772: Poseidon encryption/decryption support

ctrlc03:
# Description

Remove circomlib fork for poseidon decryption circuit, and use zk-kit instead. Also update circomlib to the latest version and use circomkit for circuits management and testing

## Related issue(s)

fix #772

## Confirmation


circuits/ts/__tests__/utils/utils.ts
```diff
@@ -1,4 +1,16 @@
-import type { WasmTester } from "circom_tester";
+import { Circomkit, type WitnessTester, type CircomkitConfig } from "circomkit";
+
+import fs from "fs";
+import path from "path";
+
+const configFilePath = path.resolve(__dirname, "..", "..", "..", "circomkit.json");
+const config = JSON.parse(fs.readFileSync(configFilePath, "utf-8")) as CircomkitConfig;
+
+// eslint-disable-next-line import/prefer-default-export
```
0xmad, line 9: Why do you need this rule? It should be disabled in linter.

0xmad (changes requested):
@ctrlc03 thanks, just few comments.
Regarding input types: I mentioned it only for one case by it's related to all the files in the PR.

circuits/circom/poseidon/poseidon.circom
```diff
@@ -0,0 +1,16 @@
+pragma circom 2.0.0;
+
+// https://github.com/weijiekoh/circomlib/blob/feat/poseidon-encryption/circuits/poseidon.circom
+include "./poseidon-cipher.circom";
+
+template Poseidon_OLD(nInputs) {
```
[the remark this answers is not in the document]
ctrlc03: it would conflict with circomlib's one if I didn't change the name (and yes technically is the old poseidon implementation which is what we have been using until now and won't change because we would need different zkeys)

circuits/ts/__tests__/utils/utils.ts
ctrlc03: oh yea that was copy paste from another repo, will remove

circuits/ts/__tests__/CeremonyParams.test.ts
```diff
@@ -150,11 +182,34 @@ describe("Ceremony param tests", () => {
...
         const currentBallotRoot = ballotTree.root;
 
-        const generatedInputs = poll.processMessages(pollId);
+        const inputs = poll.processMessages(pollId);
 
         // Calculate the witness
-        const witness = await circuit.calculateWitness(generatedInputs, true);
-        await circuit.checkConstraints(witness);
+        const correctedInputs = {
+          inputHash: inputs.inputHash as unknown as bigint,
```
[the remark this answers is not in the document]
ctrlc03: fixed

circuits/ts/__tests__/CeremonyParams.test.ts
```diff
@@ -172,30 +206,56 @@ describe("Ceremony param tests", () => {
...
-          pollEndTimestamp: generatedInputs.pollEndTimestamp,
-        });
-
-        const hasherWitness = await hasherCircuit.calculateWitness(hasherCircuitInputs, true);
-        await hasherCircuit.checkConstraints(hasherWitness);
+          coordPubKey: inputs.coordPubKey as unknown as [bigint, bigint],
+          msgRoot: inputs.msgRoot as unknown as bigint,
+          currentSbCommitment: inputs.currentSbCommitment as unknown as bigint,
+          newSbCommitment: inputs.newSbCommitment as unknown as bigint,
+          pollEndTimestamp: inputs.pollEndTimestamp as unknown as bigint,
```
0xmad, line 215: Here you don't need casts anymore.

circuits/ts/__tests__/ProcessMessages.test.ts
```diff
@@ -148,19 +182,19 @@ describe("ProcessMessage circuit", function test() {
...
-        pollEndTimestamp: generatedInputs.pollEndTimestamp,
-      });
-
-      const hasherWitness = await hasherCircuit.calculateWitness(hasherCircuitInputs, true);
-      await circuit.checkConstraints(witness);
+        coordPubKey: inputs.coordPubKey as unknown as [bigint, bigint],
+        msgRoot: inputs.msgRoot as unknown as bigint,
+        currentSbCommitment: inputs.currentSbCommitment as unknown as bigint,
+        newSbCommitment: inputs.newSbCommitment as unknown as bigint,
+        pollEndTimestamp: inputs.pollEndTimestamp as unknown as bigint,
```
0xmad, line 191: Here you can also remove casts

circuits/ts/__tests__/CeremonyParams.test.ts
ctrlc03: great catch thank you

0xmad (approved):
@ctrlc03 thanks!

circuits/scripts/build_intel.sh
0xmad, line 1: Let's add new line at the end

circuits/circom/poseidon/poseidon.circom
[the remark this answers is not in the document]
samajammin: I see, thanks for the explanation @ctrlc03. Should we create an issue to upgrade the implementation we're using (i.e. to use circomlib's)?

circuits/package.json
```diff
@@ -27,11 +27,13 @@
...
   "dependencies": {
-    "circomlib": "https://github.com/weijiekoh/circomlib#ac85e82c1914d47789e2032fb11ceb2cfdd38a2b",
+    "@zk-kit/circuits": "^0.3.0",
+    "circomlib": "^2.0.5",
     "maci-core": "^1.1.2",
     "maci-crypto": "^1.1.2",
     "maci-domainobjs": "^1.1.2",
-    "snarkjs": "^0.7.2"
+    "snarkjs": "^0.7.2",
+    "circomkit": "^0.0.19"
```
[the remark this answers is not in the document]
samajammin: Seems like more moving pieces than is necessary, but maybe there was some requirement to do this I'm missing

cli/testScript.sh
```diff
@@ -47,7 +47,7 @@ HARDHAT_CONFIG=./build/hardhat.config.js node build/ts/index.js genProofs \
     -w true 
 HARDHAT_CONFIG=./build/hardhat.config.js node build/ts/index.js proveOnChain \
     --poll-id 0 \
-    --proof-dir proofs/
+    --proof-dir proofs/ \
```
samajammin, line 50: ?

circuits/package.json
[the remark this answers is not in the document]
ctrlc03: it was needed to get it working with zk-kit, as there they use import mapping and circom_tester does not allow directly to resolve mapped imports. Not much has changed on our side with this as the testing framework we were using, circom_tester, is what circomkit uses hunder the hood

cli/testScript.sh
ctrlc03: that needed to be there for the next line to be considered part of the command, otherwise it would have been counted as a command of its own and result in a bash shell error

cli/ts/commands/proveOnChain.ts
```diff
@@ -77,7 +77,11 @@ export const proveOnChain = async (
   const maciContractAddress = maciAddress || readContractAddress("MACI");
   const messageProcessorContractAddress = messageProcessorAddress || readContractAddress(`MessageProcessor-${pollId}`);
   const tallyContractAddress = tallyAddress || readContractAddress(`Tally-${pollId}`);
-  const subsidyContractAddress = subsidyAddress || readContractAddress(`Subsidy-${pollId}`);
+
+  let subsidyContractAddress;
```
samajammin, line 81: These changes seem unrelated from the rest of this PR - are these updates fixing some bug?

cli/testScript.sh
ctrlc03: basically something we missed on a previous commit

cli/ts/commands/proveOnChain.ts
ctrlc03: yup, needed that to make manual test work properly so included here, but can definitely add in a separate one

circuits/circom/ecdh.circom
```diff
@@ -1,6 +1,7 @@
 pragma circom 2.0.0;
-include "../node_modules/circomlib/circuits/bitify.circom";
-include "../node_modules/circomlib/circuits/escalarmulany.circom";
+
+include "./bitify.circom";
```
samajammin, line 3: How are we able to import this template like this now (vs. referencing circomlib from node_modules)?

Makes it seem like this template is within the current directory 🤔

circuits/circom/poseidon/poseidon.circom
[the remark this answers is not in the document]
ctrlc03: I think it's something we should be tracking on zk-kit also, so that the poseidon-cipher circuits can be upgraded to use the new version (I've tried to implement with circomlib's new poseidon algorithm, but was unsuccessful in my first try)

circuits/circom/ecdh.circom
ctrlc03: https://github.com/privacy-scaling-explorations/maci/pull/1010#discussion_r1451600519 basically circom allows to map imports, and circomkit supports it directly via its json config file

circuits/circom/float.circom
```diff
@@ -1,4 +1,5 @@
 pragma circom 2.0.0;
+
 include "../node_modules/circomlib/circuits/bitify.circom";
```
samajammin, line 3: But here it seems we import the bitify the old way still?

ctrlc03: must have missed this import then, good catch

circuits/circom/ecdh.circom
samajammin: I see - so should we update this across all imports so we do it consistently?

samajammin: Maybe not a big deal but I think a downside of this approach is now it's less obvious where we're importing templates from (i.e. within our codebase vs. libraries)

circuits/circomkit.json
```diff
@@ -0,0 +1,17 @@
...
+  "prime": "bn128",
+  "version": "2.1.5",
+  "circuits": "./circom/circuits.json",
+  "dirPtau": "./ptau",
+  "dirCircuits": "./circom",
+  "dirInputs": "./inputs",
+  "dirBuild": "./build",
+  "optimization": 2,
+  "inspect": true,
+  "include": ["./node_modules/circomlib/circuits", "./node_modules/@zk-kit/circuits/circom"],
```
samajammin, line 12: So this is what allows us to import templates without specifying /node_modules/? i.e. https://github.com/privacy-scaling-explorations/maci/pull/1010#r1451601984

samajammin: If so, what happens if circomlib & zk-kit have a template of the same name? How do we know which we're using?

ctrlc03: zk-kit uses circomlib too, so it would never have a template with the same name. Same goes for every other circom project in existence I would think 🤔

ctrlc03: but yea if we used 2 libraries with the same template file name that would be a problem, though I don't think is one we will ever have

circuits/circom/ecdh.circom
ctrlc03: I see what you mean. Dunno really, I quite like it as it's shorter but I definitely see your point. Can always put it back if the team feels like we should keep as is

ctrlc03: I added comments throughout the templates so it is clear which templates are coming from circomlib/zk-kit vs internal

circuits/ts/__tests__/CalculateTotal.test.ts
```diff
@@ -1,16 +1,16 @@
-import { expect } from "chai";
-import tester from "circom_tester";
+import { WitnessTester } from "circomkit";
 
-import path from "path";
-
-import { getSignal } from "./utils/utils";
+import { circomkit } from "./utils/utils";
```
kittybest, line 3: The name is confusing to me hhh (I just found it's the instance instead of the package). Could we rename it? (if it's unnecessary, just ignore it)

circuits/ts/__tests__/CalculateTotal.test.ts
```diff
@@ -1,16 +1,16 @@
-import { expect } from "chai";
-import tester from "circom_tester";
+import { WitnessTester } from "circomkit";
```
kittybest, line 1: maybe we only need the type here? `import { type WitnessTester } from "circomkit"`

circuits/ts/__tests__/IncrementalQuinTree.test.ts
```diff
@@ -15,56 +13,71 @@ describe("IncrementalQuinTree circuit", function test() {
...
-    circuitGeneratePathIndices = await tester.wasm(
-      path.resolve(__dirname, "../../circom/test", `quinGeneratePathIndices_test.circom`),
-    );
-    circuitQuinSelector = await tester.wasm(path.resolve(__dirname, "../../circom/test", `quinSelector_test.circom`));
-    splicerCircuit = await tester.wasm(path.resolve(__dirname, "../../circom/test", `splicer_test.circom`));
+    circuitLeafExists = await circomkit.WitnessTester("quinLeafExists", {
+      file: "./trees/incrementalQuinTree",
+      template: "QuinLeafExists",
+      params: [3],
+    });
```
kittybest, line 26: Just a question, I saw there is another circom called `incrementalMerkleTree`, and there was tests toward it before. Does it need to be tested?

circuits/ts/__tests__/MessageValidator.test.ts
```diff
@@ -1,22 +1,60 @@
...
+    cmd: SignalValueType;
+    pubKey: SignalValueType;
+    sigR8: SignalValueType;
+    sigS: SignalValueType;
+    currentVoiceCreditBalance: SignalValueType;
+    currentVotesForOption: SignalValueType;
+    voteWeight: SignalValueType;
+    slTimestamp: SignalValueType;
+    pollEndTimestamp: SignalValueType;
+  }
```
kittybest, line 28: should we move this interface to **types.ts** file? since I saw the other interfaces were moved there.

cli/testScript.sh
kittybest: my fault sorr TT

circuits/ts/__tests__/CalculateTotal.test.ts
ctrlc03: changed to circomkitInstance to be clearer

circuits/ts/__tests__/IncrementalQuinTree.test.ts
ctrlc03: not sure I understand what you mean here, could you rephrase please

kittybest: I mean the `incrementalMerkleTree.circom` seems not be tested, will it be tested anywhere or we don’t need it?

ctrlc03: Oh right, yea this was not tested even before, and it's not even used actually. We always use the quinary one as that's what we use throughout the codebase

kittybest (approved):

PR #1031, opened 2024-01-16
refactor(optimisedmt): remove dependency and implement locally

ctrlc03:
# Description

- [x] Remove optimised dependency and implement locally 
- [x] Add a dev package to run benchmarks.

```bash
Optimisedmt - Insert, Update and Gen and verify proof for 25 leaves:
    13.7 ops/s, ±2.79%   | slowest, 2.84% slower
MACI - gen and verify proof for 25 leaves:
    14.1 ops/s, ±1.28%   | fastest
```

- [x] run benchmarks
- [x] add workflow to run benchmarks on ci

## Confirmation


crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+      throw new Error("The number of leaves must be a multiple of the tree arity");
+    }
+
+    // The number of leaves must be lower than the maximum tree capacity
+    if (numLeaves >= this.capacity) {
+      throw new Error("The number of leaves must be lower than the maximum tree capacity");
+    }
+
+    // The number of leaves must be the tree arity raised to some positive integer
+    let f = false;
```
0xmad, line 213: It's better to rename this variable

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+   * @param index The index of the leaf to generate a proof for
+   * @returns The proof
+   */
+  genProof(index: number): IMerkleProof {
+    if (index < 0) {
+      throw new Error("The leaf index must be greater than 0");
+    }
+    if (index >= this.capacity) {
+      throw new Error("The leaf index is too large");
+    }
```
0xmad, line 143:
```suggestion
    if (index < 0) {
      throw new Error("The leaf index must be greater than 0");
    }

    if (index >= this.capacity) {
      throw new Error("The leaf index is too large");
    }
```

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+    const { zeros, root } = calcInitialVals(this.arity, this.depth, this.zeroValue, this.hashFunc);
+    this.zeros = zeros;
+    this.root = root;
+
+    // calculate the number of nodes
+    let numNodes = 1;
+    for (let i = 0; i < this.depth; i += 1) {
+      numNodes += this.arity ** (i + 1);
+    }
+    this.numNodes = numNodes;
```
0xmad, line 86: I think this can be optimized to just one formula.
```suggestion
    this.numNodes = (this.arity ** (this.depth + 1)) / (this.arity - 1);
```

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+
+/**
+ * Calculate the zeroes and the root of a tree
+ * @param arity The arity of the tree
+ * @param depth The depth of the tree
+ * @param zeroValue The zero value of the tree
+ * @param hashFunc The hash function of the tree
+ * @returns The zeros and the root
+ */
+const calcInitialVals = (
```
0xmad, line 11: Let's move this function to utils or let's make it private for tree class.

crypto/ts/__tests__/IMT.test.ts
0xmad, line 1: If it's possible let's cover all the branches, statements for this tree. I usually look at coverage report and see missing tests.

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+  /**
+   * Copy the tree to a new instance
+   * @returns The new instance
+   */
+  copy(): MACITree {
+    const newTree = new MACITree(this.depth, this.zeroValue, this.arity, this.hashFunc);
+
+    newTree.nodes = this.nodes;
+    newTree.numNodes = this.numNodes;
+    newTree.zeros = this.zeros.map((z) => z);
```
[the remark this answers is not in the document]
ctrlc03: removed

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition, no-constant-condition
+    while (true) {
+      r += this.arity ** level;
+      if (index < r) {
+        break;
+      }
+
+      level -= 1;
+    }
```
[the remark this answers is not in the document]
ctrlc03: wouldn't work, though I managed to remove while(true) for another condition, lmk if that looks ok

benchmarks/.gitignore
0xmad, line 1: Let's add new line at the end

benchmarks/package.json
```diff
@@ -0,0 +1,26 @@
...
+  ],
+  "author": "PSE",
+  "license": "MIT",
+  "devDependencies": {
+    "@types/node": "^20.11.2",
+    "ts-node": "^10.9.1"
+  },
+  "dependencies": {
+    "benny": "^3.7.1",
+    "maci-core": "workspace:^1.1.2",
```
0xmad, line 22: Does version without workspace prefix work?

benchmarks/ts/src/optimisedmt.ts
```diff
@@ -0,0 +1,348 @@
+/* eslint-disable */
```
0xmad, line 1: Let's remove it

package.json
```diff
@@ -32,6 +33,7 @@
     "@commitlint/config-conventional": "^18.4.4",
     "@typescript-eslint/eslint-plugin": "^6.18.1",
     "@typescript-eslint/parser": "^6.18.1",
+    "benny": "^3.7.1",
```
0xmad, line 36: Why do you need these deps in root?

crypto/ts/quinTree.ts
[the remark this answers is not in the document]
0xmad: Seems it's still here

benchmarks/ts/suites/trees.ts
```diff
@@ -0,0 +1,50 @@
+import b from "benny";
+import { IncrementalQuinTree, hash5 } from "maci-crypto";
+
+import { OptimisedMT } from "../src/optimisedmt";
+
+const name = "merkle-trees";
```
0xmad, line 6: It's better to use uppercase for such constants

benchmarks/ts/suites/core.ts
```diff
@@ -0,0 +1,107 @@
...
+  coordinatorKeypair,
+  duration,
+  maxValues,
+  messageBatchSize,
+  stateTreeDepth,
+  treeDepths,
+  voiceCreditBalance,
+} from "./constants";
+
+const name = "maci-core";
```
0xmad, line 15: It's better to use uppercase for such constants

benchmarks/ts/suites/trees.ts
```diff
@@ -0,0 +1,50 @@
+import b from "benny";
```
0xmad, line 1: Let's use benny instead of b

benchmarks/ts/suites/constants.ts
```diff
@@ -0,0 +1,19 @@
...
+  maxMessages: 25,
+  maxVoteOptions: 25,
+};
+
+export const treeDepths = {
+  intStateTreeDepth: 2,
+  messageTreeDepth: 3,
+  messageTreeSubDepth: 2,
+  voteOptionTreeDepth: 4,
+};
```
0xmad, line 19: Let's use uppercase for constants

benchmarks/ts/src/optimisedmt.ts
0xmad, line 1: Why this file is copied here?

crypto/ts/quinTree.ts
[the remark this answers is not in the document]
ctrlc03: haven't pushed yet 😂

benchmarks/ts/src/optimisedmt.ts
ctrlc03: I'd rather keep it! I added this file as that was our old implementation https://github.com/weijiekoh/optimisedmt - wanted to run benchmarks against it but definitely doesn't make sense to fix their lint mistakes

package.json
ctrlc03: removed

benchmarks/ts/suites/trees.ts
ctrlc03: sure

benchmarks/ts/src/optimisedmt.ts
ctrlc03: to benchmark and not have to install that optimisedmt dependency which would ask for big-integer

0xmad: But you moved this tree from this package to crypto. I think it makes sense to keep benchmarks in crypto package.

benchmarks/ts/src/optimisedmt.ts
0xmad: I don't think we need to check in with old implementation in long term. So maybe it makes sense to get rid of it?

benchmarks/ts/src/optimisedmt.ts
ctrlc03: can't we use the benchmarks folder for benchmarking all of the other packages too though?

benchmarks/ts/src/optimisedmt.ts
ctrlc03: yea no problem I'll remove! guess it served its purpose now for the pr

package.json
```diff
@@ -52,6 +53,7 @@
     "prettier": "^3.2.2",
     "prettier-plugin-solidity": "^1.3.1",
     "solhint": "^4.1.1",
+    "ts-node": "^10.9.1",
```
0xmad, line 56: I think this can be removed too.

ctrlc03: yes good catch

benchmarks/ts/src/optimisedmt.ts
ctrlc03: mm so you say we should have one of these on each package? it would be cool if we could run all in one on a github action as well and have benchmarks with every changes

0xmad: Yes, it can be implemented the same way how we run tests.

ctrlc03: sure then I'll move it out

crypto/ts/quinTree.ts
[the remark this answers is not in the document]
0xmad: > wouldn't work, though I managed to remove while(true) for another condition, lmk if that looks ok

It's good, but if you can get rid of break it will be better :)

benchmarks/ts/src/optimisedmt.ts
ctrlc03: what about now?

core/ts/__tests__/benchmarks/results/maci-core.chart.html
0xmad, line 1: Let's ignore results

core/ts/__tests__/benchmarks/index.ts
```diff
@@ -0,0 +1,111 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
0xmad, line 1: Let's add benchmarks folder to exception for import/no-extraneous-dependencies linter rule in root config instead. And you can remove this line after it.

core/ts/__tests__/benchmarks/index.ts
```diff
@@ -0,0 +1,111 @@
...
+import benny from "benny";
+import { Keypair, PCommand } from "maci-domainobjs";
+
+import { MaciState } from "../../index";
```
0xmad, line 5:
```suggestion
import { MaciState } from "../..";
```

core/ts/__tests__/benchmarks/results/maci-core.chart.html
0xmad, line 1: Maybe we can remove these results and ignore them?

benchmarks/ts/src/optimisedmt.ts
0xmad: Maybe let's move it from `__tests__` folder so it will be next to it.

```
__tests__
__benchmarks__
```

core/ts/__tests__/benchmarks/results/maci-core.chart.html
ctrlc03: definitely

.eslintignore
```diff
@@ -5,3 +5,4 @@ build
 typechain-types
 .eslintrc.js
 commitlint.config.js
+__benchmarks__
```
0xmad, line 8: No, you don't need to ignore the whole benchmarks folder just add it to exceptions for rule import/no-extraneous-dependencies in root eslint config.

core/package.json
```diff
@@ -38,6 +40,8 @@
     ],
     "all": true,
     "exclude": [
+      "benchmarks/**",
```
0xmad, line 43:
```suggestion
      "__benchmarks__/**",
```

crypto/package.json
```diff
@@ -40,6 +42,8 @@
     ],
     "all": true,
     "exclude": [
+      "benchmarks/**",
```
0xmad, line 45:
```suggestion
      "__benchmarks__/**",
```

.eslintignore
0xmad: Just use this rule
```
"import/no-extraneous-dependencies": [
      "error",
      {
        devDependencies: ["**/*.test.ts", "**/__benchmarks__/*.ts"],
      },
    ],
```

ctrlc03: thanks

0xmad (approved):
@ctrlc03 thanks!

ctrlc03:
> @ctrlc03 thanks!

thanks for the review!

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,373 @@
...
+   * @param startIndex The index of the first leaf
+   * @param endIndex The index of the last leaf
+   * @returns The Merkle proof
+   */
+  genSubrootProof(
+    // inclusive
+    startIndex: number,
+    // exclusive
+    endIndex: number,
+  ): IMerkleProof {
```
kittybest, line 173: I saw we check if the `index` is negative in the `genProof` function, should we check it here too?

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,373 @@
+import type { Leaf, Node, IMerkleProof } from "./types";
+
+/**
+ * An implementation of an incremental Merkle tree
+ * @dev adapted from https://github.com/weijiekoh/optimisedmt
+ */
+export class MACITree {
```
kittybest, line 7: I think we call it `IncrementalQuinTree` through the whole project, why not just name it as `IncrementalQuinTree`? will it be conflicted with something else?

crypto/ts/__tests__/IMT.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  describe("insert", () => {
+    it("should update the root after one insertion", () => {
+      const mt1 = new IncrementalQuinTree(5, 0n, 5, hash5);
+
+      const rootBefore = mt1.root;
+      mt1.insert(1n);
+
+      expect(mt1.root).to.not.eq(rootBefore);
+    });
+  });
```
kittybest, line 25: Does the function `update` also need to be tested?

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+  };
+
+  /**
+   * Calculate the indices of the parent
+   * @param index The index of the leaf
+   * @returns The indices of the parent
+   */
+  calcParentIndices(index: number): number[] {
+    // can only calculate the parent for leaves not subroots
+    if (index >= this.capacity) {
```
kittybest, line 250: How about checking `index >= this.capacity || index < 0`?

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  /**
+   * Calculate the indices of the children of a node
+   * @param index The index of the node
+   * @returns The indices of the children
+   */
+  calcChildIndices(index: number): number[] {
+    // cannot get the children of a leaf
+    if (index < this.capacity) {
```
kittybest, line 274: same as above (`calcParentIndices`)

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+    return this.zeros[this.depth - level];
+  }
+
+  /**
+   * Set a node (not the root)
+   * @param index the index of the node
+   * @param value the value of the node
+   */
+  setNode(index: number, value: Leaf): void {
+    if (index > this.numNodes - 1) {
```
kittybest, line 328: `index > this.numNodes - 1 || index < 0`

crypto/ts/quinTree.ts
```diff
@@ -0,0 +1,377 @@
...
+
+    return indices;
+  }
+
+  /**
+   * Get a node at a given index
+   * @param index The index of the node
+   * @returns The node
+   */
+  getNode(index: number): Leaf {
```
kittybest, line 299: should we also check if it's **out of bounds** here?

ctrlc03: don't think there's any need here, as checks are done where we call it from and it's never used directly

crypto/ts/__benchmarks__/suites/trees.ts
```diff
@@ -0,0 +1,40 @@
...
+    benny.complete((results) => {
+      results.results.forEach((result) => {
+        // eslint-disable-next-line no-console
+        console.log(`${result.name}: mean time: ${result.details.mean.toFixed(2)}`);
+      });
+    }),
+
+    benny.save({ folder: "benchmarks/results", file: NAME, version: "1.0.0", details: true }),
+    benny.save({ folder: "benchmarks/results", file: NAME, format: "chart.html", details: true }),
+    benny.save({ folder: "benchmarks/results", file: NAME, format: "table.html", details: true }),
```
kittybest, line 38: here would create a folder called `benchmarks/results` in `crypto/`, which is not ignored in the `.gitignore` file.

core/ts/__benchmarks__/index.ts
```diff
@@ -0,0 +1,110 @@
...
+    benny.complete((results) => {
+      results.results.forEach((result) => {
+        // eslint-disable-next-line no-console
+        console.log(`${result.name}: mean time: ${result.details.mean.toFixed(2)}`);
+      });
+    }),
+
+    benny.save({ folder: "__benchmarks__/results", file: NAME, version: "1.0.0", details: true }),
+    benny.save({ folder: "__benchmarks__/results", file: NAME, format: "chart.html", details: true }),
+    benny.save({ folder: "__benchmarks__/results", file: NAME, format: "table.html", details: true }),
```
kittybest, line 106: I think it should be `ts/__benchmarks__/results`

crypto/ts/__benchmarks__/suites/trees.ts
```diff
@@ -0,0 +1,40 @@
...
+    benny.complete((results) => {
+      results.results.forEach((result) => {
+        // eslint-disable-next-line no-console
+        console.log(`${result.name}: mean time: ${result.details.mean.toFixed(2)}`);
+      });
+    }),
+
+    benny.save({ folder: "__benchmarks__/results", file: NAME, version: "1.0.0", details: true }),
+    benny.save({ folder: "__benchmarks__/results", file: NAME, format: "chart.html", details: true }),
+    benny.save({ folder: "__benchmarks__/results", file: NAME, format: "table.html", details: true }),
```
kittybest, line 38: same here, `ts/__benchmarks__/results`

kittybest (approved):

PR #1052, opened 2024-01-18
feat(website): update ui based on feedback

kittybest:
# Description

- updating sketch images
- update font-family and adjust font-size
- adjust hero-image
- update footer-link hover color

## Confirmation


website/src/components/HomepageFeatures/index.tsx
```diff
@@ -1,22 +1,29 @@
...
+import chainDarkImage from "../../../static/img/chain_dark.png";
 import chartImage from "../../../static/img/chart.png";
+import chartDarkImage from "../../../static/img/chart_dark.png";
 
 import styles from "./styles.module.css";
 
 interface FeatureItem {
   title: string;
   img: string;
+  // eslint-disable-next-line react/no-unused-prop-types
```
0xmad, line 17: Please, don't disable linter rules. I think this one can be easily removed because you use this prop.

website/src/pages/index.module.css
```diff
@@ -9,23 +9,30 @@
   overflow: hidden;
   background-image: linear-gradient(to bottom, rgba(0, 0, 0, 0.3) 0%, rgba(0, 0, 0, 0.3) 100%),
     url("../../static/img/hero.svg");
-  background-position: center;
+  background-position: 0% 75%;
```
0xmad, line 12: Have you checked this change for all the screen resolutions?

0xmad (commented):
@kittybest thanks, just one question and one improvement

website/src/pages/index.module.css
kittybest: I checked it. Making this change since the design team decide to use the lower part of the image

website/src/components/HomepageFeatures/index.tsx
kittybest: I think the link check failed of this, not sure how to rewrite this part could make it clearer

0xmad: So, you need to use props type for `Feature` component `Omit<FeatureItem, "imgDark">`. This is because this component doesn't use `imgDark` but it's defined in props.

kittybest: yep right, but I move the decision on whether to load dark or light image to that component, so I think it looks fine now.

0xmad (approved):
@kittybest thanks!

PR #1070, opened 2024-01-21
fix(build): fix bug that prevented c witness from being compiled

ctrlc03:
# Description

Currently, compiling the c witness generator will fail because execFileSync commands share no context, thus the second command will still run in the original directory, not the one where the first command moved to.

## Confirmation


0xmad (approved):
@ctrlc03 thanks!

Issue #1053, opened 2024-01-18
--dry-run flag for maci-cli


0xmad:
# Prerequisites

Please answer the following question for yourself before submitting a feature request.

- [x] I checked to make sure that this request has not already been filed

# Motivation

As a user I'd like to have `--dry-run` flag for cli to check if all operations are successful without paying for transaction.

# Details

We can use static calls for calling contract methods and check if the call is successful. It will allow us to get rid of hardcoded gas limits and user can configure gas limit without fearing of losing money.

Make sure that you have reviewed [MACI's Contributor Guidelines](https://github.com/privacy-scaling-explorations/maci/blob/dev/CONTRIBUTING.md)

0xmad:
I think it's still relevant. --dry-run can be used with contracts and platform.

PR #1087, opened 2024-01-23
fix(cli): remove hardcoded gas limit from signup and publish

### LINKED ISSUE #1086: Cannot signup on Arbitrum Sepolia network
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am not running a deprecated version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

# Expected Behavior

Should allow a user to signup using MACI's cli 

# Current Behavior

Transactions reverts.

# Failure Information

```bash
transaction execution reverted (action="sendTransaction", data=null, reason=null, invocation=null, revert=null, transaction={ "data": "", "from": "", "to": "" }, [...] code=CALL_EXCEPTION, version=6.10.0)
```

ctrlc03:
# Description

hardcoded gas limit might prevent transaction from succeeding on certain networks

## Related issue(s)

fix #1086

## Confirmation

### DIFF
--- a/cli/ts/commands/publish.ts
+++ b/cli/ts/commands/publish.ts
@@ -135,9 +135,7 @@ export const publish = async ({
 
   try {
     // submit the message onchain as well as the encryption public key
-    const tx = await pollContract.publishMessage(message.asContractParam(), encKeypair.pubKey.asContractParam(), {
-      gasLimit: 10000000,
-    });
+    const tx = await pollContract.publishMessage(message.asContractParam(), encKeypair.pubKey.asContractParam());
     const receipt = await tx.wait();
 
     if (receipt?.status !== 1) {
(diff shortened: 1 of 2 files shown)


kittybest (approved):
thanks!
I'm curious that does that mean the gas spent more than we set up?

ctrlc03:
> thanks! I'm curious that does that mean the gas spent more than we set up?

yes, usually you would set a gas limit when you want the tx to be sent regardless of whether it would fail or not. If you don't specify, and if the tx fails it will just not send it and give you the error. In our case, it seems that on this test network signin up spends more than 1M gas, so the tx would go through but fail because of not enough gas (limit was set to 1M)

0xmad:
@ctrlc03 I created a task for this case #1053. So the idea is to use static calls to check if transaction can be failed without paying for the gas.

PR #1101, opened 2024-01-24
fix(contracts): save numSignups after merging maciStateAq inside the Poll contract

### LINKED ISSUE #1091: Signups after logs are fetched would cause proofs to not verify on chain
Currently, if a user is able to sign up after a poll logs have been fetched locally, the number of signups on the MACI contract would change, causing a discrepancy between the data fetched offhcain and the one on chain. 

A fix could be to store the number of signups at the time that the poll completed (when `mergeMaciStateAq` is run) and then use it as a source of truth. This could also enable another Poll to run at the same time from the same MACI instance, where users won't have to signup twice (for instance)

ctrlc03:
# Description

prevent new signups from changing the record of users being signed for a specific poll. Instead store this value in the Poll contract after the stateAq is merged. This way, even if a user signs up after the logs are fetched and the state is reproduced offchain, the proof validation will not fail because of a numSignups mismatch.

## Additional Notes

Changes:

* when calling `mergeMaciStateAq` on the Poll contract, we now save the number of signups on a state variable
* the `numSignUpsAndMessages` function now returns the value from the Poll state and not from the MACI contract so if there's any new signup, this does not affect proof verification later on
* removed the state tree from MaciState. Instead, `MaciState.ts` will store the state leaves only. 
* amended the `copyStateFromMaci` function in `Poll.ts` to accept `numSignups` . This function is now called `updatePoll` and copies over `numSignups` leaves from `MaciState.ts` and creates its own state tree. 
* amended `genMaciStateFromContract` to call `updatePoll` with the number of signups taken directly from the on chain Poll contract. It also checks that the local state root matches the one saved on the Poll contract.
* amended tests where we mock calling `updatePoll` with the number of leaves 

Benefits:

* We remove one operation of copying over the stateAq for a poll, and instead fill the tree within the poll (should give a little speed boost) 
* Make it easier to then allow multiple concurrent polls as each would store independent data for finalization and not interfere with each other
* fixing the issue

## Related issue(s)

fix #1091

## Confirmation

### DIFF
--- a/contracts/contracts/Poll.sol
+++ b/contracts/contracts/Poll.sol
@@ -18,14 +18,16 @@ import { Utilities } from "./utilities/Utilities.sol";
 contract Poll is Params, Utilities, SnarkCommon, Ownable, EmptyBallotRoots, IPoll {
   using SafeERC20 for ERC20;
 
+  /// @notice Whether the Poll has been initialized
   bool internal isInit;
 
-  // The coordinator's public key
+  /// @notice The coordinator's public key
   PubKey public coordinatorPubKey;
 
   /// @notice Hash of the coordinator's public key
   uint256 public immutable coordinatorPubKeyHash;
 
+  /// @notice the state root of the state merkle tree
   uint256 public mergedStateRoot;
 
   // The timestamp of the block at which the Poll was deployed
@@ -46,7 +48,12 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable, EmptyBallotRoots, IPol
   /// the case that none of the messages are valid.
   uint256 public currentSbCommitment;
 
+  /// @notice The number of messages that have been published
   uint256 public numMessages;
+
+  /// @notice The number of signups that have been processed
+  /// before the Poll ended (stateAq merged)
+  uint256 public numSignups;
 
   /// @notice Max values for the poll
   MaxValues public maxValues;
@@ -68,7 +75,7 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable, EmptyBallotRoots, IPol
   event PublishMessage(Message _message, PubKey _encPubKey);
   event TopupMessage(Message _message);
   event MergeMaciStateAqSubRoots(uint256 _numSrQueueOps);
-  event MergeMaciStateAq(uint256 _stateRoot);
+  event MergeMaciStateAq(uint256 _stateRoot, uint256 _numSignups);
   event MergeMessageAqSubRoots(uint256 _numSrQueueOps);
   event MergeMessageAq(uint256 _messageRoot);
 
@@ -210,7 +217,9 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable, EmptyBallotRoots, IPol
     sb[2] = uint256(0);
 
     currentSbCommitment = hash3(sb);
-    emit MergeMaciStateAq(mergedStateRoot);
+
+    numSignups = extContracts.maci.numSignUps();
+    emit MergeMaciStateAq(mergedStateRoot, numSignups);
   }
 
   /// @inheritdoc IPoll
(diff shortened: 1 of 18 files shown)


contracts/contracts/Poll.sol
```diff
@@ -237,8 +247,8 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable, EmptyBallotRoots, IPol
   }
 
   /// @inheritdoc IPoll
-  function numSignUpsAndMessages() public view returns (uint256 numSignups, uint256 numMsgs) {
-    numSignups = extContracts.maci.numSignUps();
-    numMsgs = numMessages;
+  function numSignUpsAndMessages() public view returns (uint256 _numSignups, uint256 _numMessages) {
```
0xmad, line 250: I'd keep these variable names without underscore because on frontend they will be accessed by the names and it can conflict with code styles and linter. Let's rename them to avoid it.

core/ts/Poll.ts
```diff
@@ -130,6 +131,9 @@ export class Poll implements IPoll {
 
   emptyBallotHash?: bigint;
 
+  // how many users signed up
+  numSignups = 0n;
```
0xmad, line 135: Let's use private visibility here and create getter for external read.

core/ts/Poll.ts
```diff
@@ -173,12 +177,30 @@ export class Poll implements IPoll {
...
+    // not match. For this, we must only copy up to the number of signups
+
     // Copy the state tree, ballot tree, state leaves, and ballot leaves
-    this.stateLeaves = this.maciStateRef.stateLeaves.map((x) => x.copy());
-    this.stateTree = this.maciStateRef.stateTree.copy();
+
+    // start by setting the number of signups
+    this.setNumSignups(numSignups);
+    // copy up to numSignups state leaves
+    this.stateLeaves = this.maciStateRef.stateLeaves.slice(0, Number(this.numSignups)).map((x) => x.copy());
```
0xmad, line 196: Is it ok to cast bigint to number? Just want to make sure if we don't have overflow here.

core/ts/__tests__/Poll.test.ts
```diff
@@ -407,7 +414,7 @@ describe("Poll", function test() {
     poll.publishMessage(message, ecdhKeypair.pubKey);
 
     it("should throw if called before all messages have been processed", () => {
-      expect(() => poll.tallyVotes()).to.throw("Cannot read properties of undefined (reading 'root')");
+      expect(() => poll.tallyVotes()).to.throw("Cannot convert undefined to a BigInt");
```
0xmad, line 417: It's really strange error from UX perspective. Is it possible to use try/catch to tell what actually is the problem?

kittybest (approved):
lgtm.
Just wanna make sure I understand why we remove the state tree in MaciState ----> is that because we only use it for copy to poll before, but now we decide to build tree in each Poll based on the numSignups, thus we don't need the tree in MaciState anymore?

core/ts/Poll.ts
ctrlc03: so if we have `stateTreeDepth` at 10 we can have max `9765625` signups, at 20 we can have `95367431640625` signups and the maxSafeInt in JS is `9007199254740991` so we are good. Thinking though if it might be easier to just use a map for stateLeaves, and might be more efficient for the way we use the data

ctrlc03:
> lgtm. Just wanna make sure I understand why we remove the state tree in MaciState ----> is that because we only use it for copy to poll before, but now we decide to build tree in each Poll based on the numSignups, thus we don't need the tree in MaciState anymore?

Correct!

core/ts/Poll.ts
ctrlc03: I had a look and the changes would be substantial, we'd also need to update the merkle tree structure to use bigint indexing, so not sure it's worth the change. MaxSafeInt fits 1113716 the current world population so probably safe for now to keep using it. Lmk your thoughts

0xmad (approved):
@ctrlc03 thanks!

Issue #1133, opened 2024-01-30
Topup feature does not work - incorrectly throw insufficient funds error


yuetloo:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am not running a deprecated version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

# Expected Behavior

Given the following test case, `genProof` should generate correct tally result.

```
contribute 1 voice credits
topup 2 voice credits
vote 2 voice credit to project 1
vote 1 voice credits to project 2

``` 
The genProof output, tally.json file should produce the following result:

totalSpentVoiceCredits.spent = 3
perVOSpentVoiceCredits.tally[1] = 2
perVOSpentVoiceCredits.tally[2] = 1

# Current Behavior

What is the current behavior?

genProofs throws InsufficientVoiceCredits error.

https://github.com/privacy-scaling-explorations/maci/blob/dev/core/ts/Poll.ts#L281

```
      // If the remaining voice credits is insufficient, do nothing
      if (voiceCreditsLeft < 0n) {
        throw new ProcessMessageError(ProcessMessageErrors.InsufficientVoiceCredits);
      }
```

# Failure Information

The issue is that `topup` and `vote` messages are processed in reverse order instead of the order they show up in the logs, see the following line here:
https://github.com/privacy-scaling-explorations/maci/blob/dev/core/ts/Poll.ts#L507

As a result, when it processes the first vote message with 2 voice credits, it hasn't processed the topup message, which is supposed to update the state tree balance here:

https://github.com/privacy-scaling-explorations/maci/blob/dev/core/ts/Poll.ts#L580-L582
```
newStateLeaf.voiceCreditBalance += amount;
              // save it
              this.stateLeaves[stateIndex] = newStateLeaf;
``` 

## Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. step 1
2. step 2
3. you get it...

## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

- MACI version:
- Firmware Version:
- Operating System:
- SDK version:
- Toolchain version:

## Failure Logs

Please include any relevant log snippets or files here.

yuetloo:
@ctrlc03 , I don't see how I can reorder the log events as suggested. The events are logged by the contract and the contract has to enforce the order and verify the contributor has enough voice credits before they can vote.

```
contribute 1 voice credits
[...]

PR #1170, opened 2024-02-08
fix(circuits): enforce use of stateIndex from message

ctrlc03:
# Description

Prevent passing unmatched state leaves to censor votes by the coordinator. Instead, ensure that the state leaf sent for each message, is the state leaf at position message.stateIndex within the stateTree. Also, force decryption on the TS message processing, to ensure that a malicious user might not send an invalid message which when decrypted by the circuit (using decryptWithoutCheck and thus possibily failing decryption) would result in a valid stateIndex, causing a DoS on the coordinator which would not be able to generate a proof.

Also, the padding messages are not anymore the last message sent (this.messages[this.messages.length - 1]), but actually an empty command with a state index 0 so that when processed in the circuit, it will have no effect.

cc @kcharbo3 

## Related issue(s)

re #1133 

## Confirmation

### DIFF
--- a/circuits/circom/processMessages.circom
+++ b/circuits/circom/processMessages.circom
@@ -48,6 +48,8 @@ template ProcessMessages(
     var STATE_LEAF_PUB_Y_IDX = 1;
     var STATE_LEAF_VOICE_CREDIT_BALANCE_IDX = 2;
     var STATE_LEAF_TIMESTAMP_IDX = 3;
+
+    var N_BITS = 252;
     
     // Note that we sha256 hash some values from the contract, pass in the hash
     // as a public input, and pass in said values as private inputs. This saves
@@ -277,6 +279,7 @@ template ProcessMessages(
     component processors[batchSize]; 
     // topup type processor
     component processors2[batchSize]; 
+
     for (var i = batchSize - 1; i >= 0; i --) {
         // process it as vote type message
         processors[i] = ProcessOne(stateTreeDepth, voteOptionTreeDepth);
@@ -349,6 +352,7 @@ template ProcessMessages(
                     <== currentStateLeavesPathElements[i][j][k];
             }
         }
+
         // pick the correct result by msg type
         tmpStateRoot1[i] <== processors[i].newStateRoot * (2 - msgs[i][0]); 
         tmpStateRoot2[i] <== processors2[i].newStateRoot * (msgs[i][0] - 1);
@@ -378,6 +382,8 @@ template ProcessTopup(stateTreeDepth) {
     var STATE_LEAF_VOICE_CREDIT_BALANCE_IDX = 2;
     var STATE_LEAF_TIMESTAMP_IDX = 3;
 
+    var N_BITS = 252;
+
     signal input msgType;
     signal input stateTreeIndex;
     signal input amount;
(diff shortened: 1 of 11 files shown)


samajammin:
Thanks @ctrlc03!

> Also, the padding messages are not the last message sent, but actually the noting up my sleeve message with a state index 0 so that we can process it as null in the circuit.

Not sure what you mean hear, could you please clarify?

ctrlc03:
> Thanks @ctrlc03!
> 
> 
> 
> > Also, the padding messages are not the last message sent, but actually the noting up my sleeve message with a state index 0 so that we can process it as null in the circuit.
> 
> 
> 
> Not sure what you mean hear, could you please clarify?

Have a look at core/ts/Poll.ts in genProcessMessagesPartialInputs, for some reason we were padding with the last message in the array (basically there might be one batch which needs to be padded to be full, which is the first one unless messages % batch size === 0). Now instead of sending the last message in the message array, which could have been a top up or a vote message (random based on whatever is sent last in a voting round), we just send a command  with a state index of 0 so when processed by the circuit it'll just do nothing.

circuits/circom/processMessages.circom
```diff
@@ -526,22 +528,34 @@ template ProcessOne(stateTreeDepth, voteOptionTreeDepth) {
...
 
-    component stateIndexMux = Mux1();
-    stateIndexMux.s <== transformer.isValid + msgType - 1;
-    stateIndexMux.c[0] <== 0;
-    stateIndexMux.c[1] <== indexByType;
+    // we can validate if the state index is within the numSignups
+    // if not, we use 0
+    // this is because decryption of an invalid message 
+    // might result in random packed vals
+    component validStateLeafIndex = SafeLessThan(252);
```
0xmad, line 543: Seems 252 is a magic constant. Can you move it to a separate variable?

0xmad (commented):
@ctrlc03 thanks, just one minor comment.

0xmad (approved):
@ctrlc03 thanks!

core/ts/Poll.ts
```diff
@@ -554,21 +554,62 @@ export class Poll implements IPoll {
...
+                  currentBallotsPathElements.unshift(this.ballotTree!.genProof(0).pathElements);
+
+                  // Since the command is invalid, we use a zero vote weight
+                  currentVoteWeights.unshift(this.ballots[0].votes[0]);
+
+                  // create a new quinary tree and add an empty vote
+                  const vt = new IncrementalQuinTree(this.treeDepths.voteOptionTreeDepth, 0n, STATE_TREE_ARITY, hash5);
+                  vt.insert(this.ballots[0].votes[0]);
+                  // get the path elements for this empty vote weight leaf
+                  currentVoteWeightsPathElements.unshift(vt.genProof(0).pathElements);
```
kittybest, line 611: have a question here: if a voter sent an invalid message, will the whole tree being updated (and insert the empty vote?) --> if yes, would that cause a tree full of empty votes since someone keep on sending invalid message?

circuits/ts/__tests__/ProcessMessages.test.ts
```diff
@@ -76,7 +76,7 @@ describe("ProcessMessage circuit", function test() {
   describe("1 user, 2 messages", () => {
```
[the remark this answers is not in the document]
ctrlc03: this is added so it replicates a real voting round. The first message sent by the poll contract is this nothing message

core/ts/Poll.ts
ctrlc03: this is just a local tree generated to hold the current user (stateIndex) vote. in this case, the vote was completely invalid, so we just send dummy data which does not prevent a proof from being generated, but really has no effect in the state machine logic. this.ballots[0] will be the ballot of the stateIndex 0 which is the empty state leaf we put in the tree, so it'll just be 0

circuits/ts/__tests__/ProcessMessages.test.ts
```diff
@@ -554,7 +575,7 @@ describe("ProcessMessage circuit", function test() {
 
         // Second batch is not a full batch
         const numMessages = messageBatchSize * NUM_BATCHES - 1;
-        for (let i = 0; i < numMessages; i += 1) {
+        for (let i = 0; i < 6; i += 1) {
```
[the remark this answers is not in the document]
ctrlc03: don't remember now, but will put back as it was

kcharbo3:
LGTM!

PR #1193, opened 2024-02-14
fix(tests): remove wrong chai matcher from integrationTests

ctrlc03:
# Description

Remove wrong except matcher which causes invalid tests to pass.

## Confirmation

### DIFF
--- a/integrationTests/ts/__tests__/integration.test.ts
+++ b/integrationTests/ts/__tests__/integration.test.ts
@@ -347,7 +347,7 @@ describe("Integration tests", function test() {
           subsidyAddress: pollContracts.subsidy,
           signer,
         }),
-      ).to.eventually.not.rejectedWith();
+      ).to.not.be.rejected;
 
       // verify the proofs
       await expect(
@@ -365,7 +365,7 @@ describe("Integration tests", function test() {
             : undefined,
           signer,
         }),
-      ).to.eventually.not.rejectedWith();
+      ).to.not.be.rejected;
     });
   });
 });


integrationTests/ts/__tests__/integration.test.ts
```diff
@@ -336,36 +336,32 @@ describe("Integration tests", function test() {
...
+      await proveOnChain({
+        pollId,
+        proofDir: path.resolve(__dirname, "../../../cli/proofs"),
+        subsidyEnabled,
+        maciAddress: contracts.maciAddress,
+        messageProcessorAddress: pollContracts.messageProcessor,
+        tallyAddress: pollContracts.tally,
+        subsidyAddress: pollContracts.subsidy,
+        signer,
+      }).catch((err) => expect(err).to.eq(undefined));
```
0xmad, line 348: If the error is not thrown this catch won't be fired but in case if there is an error it will fail because error is not undefined.
Is it possible to check return value or if it resolves properly?

0xmad (commented):

integrationTests/ts/__tests__/integration.test.ts
ctrlc03: maybe we don't even add except and just let it throw if it throws? so the test fails?

ctrlc03: none of those functions return anything, and the matcher we used before seem to not work. is there any other matcher you know of that could work?

0xmad: Usually, it's a good practice to have expect for resolved values as well, because it can be resolved with undefined or null values, but it will be considered as valid tests.

ctrlc03: I added .to.not.be.rejected and seems to work fine

0xmad (approved):
@ctrlc03 thanks!

PR #1224, opened 2024-02-20
feat(cli): add get poll cli command

0xmad:
# Description

- [x] Add get poll cli command
- [x] Add indexed params for deploy poll event

## Additional Notes

N/A

## Related issue(s)

Related to https://github.com/privacy-scaling-explorations/maci-rpgf/issues/10

## Confirmation

### DIFF
--- a/cli/ts/commands/poll.ts
+++ b/cli/ts/commands/poll.ts
@@ -1,0 +1,61 @@
+import { ZeroAddress } from "ethers";
+import { MACI__factory as MACIFactory, Poll__factory as PollFactory } from "maci-contracts/typechain-types";
+
+import type { IGetPollArgs, IGetPollData } from "../utils/interfaces";
+
+import { banner } from "../utils/banner";
+import { logError, logGreen, success } from "../utils/theme";
+
+/**
+ * Get deployed poll from MACI contract
+ * @param {IGetPollArgs} args - The arguments for the get poll command
+ * @returns {IGetPollData} poll data
+ */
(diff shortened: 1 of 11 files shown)


cli/ts/commands/poll.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (pollAddress === ZeroAddress) {
+    logError("MACI contract doesn't have any deployed poll");
+  }
+
+  const pollContract = PollFactory.connect(pollAddress, signer);
+
+  const [[deployTime, duration], isStateAqMerged, numSignups] = await Promise.all([
+    pollContract.getDeployTimeAndDuration(),
+    pollContract.stateAqMerged(),
+    pollContract.numSignups(),
```
ctrlc03, line 36: so the number of signups will only be known after the state aq is merged. what about taking it from maci instead which will have up to date info if the poll is not completed yet?

ctrlc03 (changes requested):
Nice thank you! just two comments

cli/ts/commands/poll.ts
ctrlc03: or maybe to be even more precise, we first can check if the state aq are merged, if they are, take from poll. if not take from maci

0xmad: Does it update when new poll is deployed? If no, I can just remove it.

ctrlc03: on maci, numSignups can keep growing poll after poll as new users signup. whereas on a specific poll, we store the number of users signed up up until the poll is finalized

ctrlc03 (approved):
Thank you!!

PR #1251, opened 2024-02-26
feat(cli): add publish batch method for cli sdk

0xmad:
# Description

Add publish batch method for maci-cli sdk.

## Additional Notes

N/A

## Related issue(s)

Related to https://github.com/privacy-scaling-explorations/maci-rpgf/issues/7

## Confirmation

### DIFF
--- a/cli/ts/commands/publish.ts
+++ b/cli/ts/commands/publish.ts
@@ -1,8 +1,15 @@
 import { MACI__factory as MACIFactory, Poll__factory as PollFactory } from "maci-contracts/typechain-types";
 import { genRandomSalt } from "maci-crypto";
-import { Keypair, PCommand, PrivKey, PubKey } from "maci-domainobjs";
+import {
+  type IG1ContractParams,
+  type IMessageContractParams,
+  Keypair,
+  PCommand,
+  PrivKey,
+  PubKey,
+} from "maci-domainobjs";
 
-import type { PublishArgs } from "../utils/interfaces";
+import type { IPublishBatchArgs, IPublishBatchData, PublishArgs } from "../utils/interfaces";
 
 import { banner } from "../utils/banner";
 import { contractExists } from "../utils/contracts";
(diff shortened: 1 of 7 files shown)


cli/ts/utils/interfaces.ts
```diff
@@ -708,63 +708,112 @@ export interface ProveOnChainArgs {
...
 
   /**
    * The address of the MACI contract
    */
   maciContractAddress: string;
 
+  /**
+   * The public key of the user
+   */
+  publicKey: string;
```
kittybest, line 765: nit pick: it is named `pubKey` in the `PublishArgs`

kittybest (commented):
maybe also add publish test cmd in `package.json`?

0xmad:
> maybe also add publish test cmd in `package.json`?

Unit tests can be run via `test:unit` command.

kittybest:
> > maybe also add publish test cmd in `package.json`?
> 
> Unit tests can be run via `test:unit` command.

I mean something like `test:publish`? since I saw there are commands for other single unit test

0xmad:
> > > maybe also add publish test cmd in `package.json`?
> > 
> > 
> > Unit tests can be run via `test:unit` command.
> 
> I mean something like `test:publish`? since I saw there are commands for other single unit test

Ok, why do you need it? If I run tests for cli, I just run all the tests to make sure if nothing is broken.

kittybest (approved):

cli/ts/commands/publish.ts
```diff
@@ -129,3 +136,108 @@ export const publish = async ({
...
+      throw new Error("invalid salt");
+    }
+  });
+
+  const coordinatorPubKey = new PubKey([
+    BigInt(coordinatorPubKeyResult.x.toString()),
+    BigInt(coordinatorPubKeyResult.y.toString()),
+  ]);
+
+  const encryptionKeypair = new Keypair();
```
ctrlc03, line 206: I see the point of using the same key here for all messages in the batch, we just assume that all messages in the batch are from the same user either way for simplicity sake (though that doesn't necessarily have to be the case), so I would suggest we move line 226 out of the loop to avoid computing the ECDH key every time; and also we should probably return this key like we do in publishMessage.

PR #1299, opened 2024-03-14
feat(cli): add initial voice credits from signup event

0xmad:
# Description

Add initial voice credits for poll sdk method

## Additional Notes

N/A

## Related issue(s)

Related to https://github.com/privacy-scaling-explorations/maci-rpgf/pull/39

## Confirmation

### DIFF
--- a/cli/ts/commands/signup.ts
+++ b/cli/ts/commands/signup.ts
@@ -89,14 +89,14 @@ export const signup = async ({
 /**
  * Checks if user is registered with public key
  * @param IRegisteredArgs - The arguments for the register check command
- * @returns user registered or not and state index
+ * @returns user registered or not and state index, voice credit balance
  */
 export const isRegisteredUser = async ({
   maciAddress,
   maciPubKey,
   signer,
   quiet = true,
-}: IRegisteredUserArgs): Promise<{ isRegistered: boolean; stateIndex?: string }> => {
+}: IRegisteredUserArgs): Promise<{ isRegistered: boolean; stateIndex?: string; voiceCredits?: string }> => {
   banner(quiet);
 
   const maciContract = MACIFactory.connect(maciAddress, signer);
@@ -104,11 +104,13 @@ export const isRegisteredUser = async ({
 
   const events = await maciContract.queryFilter(maciContract.filters.SignUp(undefined, publicKey.x, publicKey.y));
   const stateIndex = events[0]?.args[0].toString() as string | undefined;
+  const voiceCredits = events[0]?.args[3].toString() as string | undefined;
 
   logGreen(quiet, success(`State index: ${stateIndex?.toString()}, registered: ${stateIndex !== undefined}`));
 
   return {
     isRegistered: stateIndex !== undefined,
     stateIndex,
+    voiceCredits,
   };
 };


cli/ts/commands/poll.ts
```diff
@@ -11,28 +16,44 @@ import { logError, logGreen, success } from "../utils/theme";
...
+  pollId,
+  ivcpDataArg,
+  quiet = true,
+}: IGetPollArgs): Promise<IGetPollData> => {
   banner(quiet);
 
   const maciContract = MACIFactory.connect(maciAddress, signer);
   const id =
     pollId === undefined ? await maciContract.nextPollId().then((nextPollId) => nextPollId - 1n) : BigInt(pollId);
+  const ivcpData = ivcpDataArg || DEFAULT_IVCP_DATA;
```
ctrlc03, line 31: so the idea is that for each user we call this function with their own ivcpData? as the ivcpData might differ from user to user for certain applications (random example - it could be based on the number of tokens in a wallet at signup time). Would it be easier to get the initial voice credits from the user signup event instead?

0xmad: So I chose this one because it's using in signup function (so application already has this data and if not, we just use default one). Plus, I think it's not a constant value so it can be changed so that's why I use contract method and not event.

0xmad: @ctrlc03 I added this functionality to `isRegisteredUser` so if someone wants just initial value, they can get it from event with state index.

ctrlc03: why would someone want a value that is not the initial one given it's the only valid one?

0xmad: So it possible can be used as how many votes are left after voting. If we take a look at the abstract class for InitialVoiceCreditProxy it doesn't show that it can't be used this way. For instance, your vote weight depends on your DAI amount, but if you transfer your DAI, your vote weight will be decreased.

ctrlc03: Right, so it's not really easy to implement a fair system where the voice credits are given based on a token balance, due to what you just said (or even flashloans for what is worth, you could flashloan 100000 DAI, signup with 100000 voice credits and not have any more DAI the next block.) However, votes are encrypted, so really there is no way to validate on chain that your vote is using <= voice credits left. Same way, off chain, if you signed up on chain with 100 voice credits for whatever reason, when processing votes we can only process with the knowledge that when you signed up you had 100 tokens. (The state leaf stored on the merkle tree on chain never changes, and the state tree root is very important for the validity of the snark proof).

Now, I'm not too familiar with DAO voting, though I assume that for a dao vote I can vote with a power equal to my balance at the time (or weighted balance for a period of x days, or $ staked). Though next time I vote for a new proposal (in my mind here proposal = maci poll) I would have a different amount of tokens so a different voting power. In this case, I think we could not reuse the same signup tree for dao voting with maci due to this, but that's a discussion for another time haha

Anyway, this is why I suggested to only account for the credit balance at signup time, because it's the only one that really counts when processing messages and generating snark proofs, and wanted to keep MACI as general as possible as people might implement several different ways of calculating the initial voice credit amount.

ctrlc03 (approved):
Thank you!

kittybest (approved):
thanks! (though I am not sure if I understood your discussion 0.0

PR #1304, opened 2024-03-17
fix: for genLocalState '-o' represents the path where to write the local state and not the pollId and must be be parsed as int.

GaetanoMondelli:
# Description

In the cli the command `genLocalState` has ha '-o' parameter used to specifiy the path where to write the local state. 

## Additional Notes

For other commands '-o' represented the pollId and the parameter required to be parsed into an integer but here trying to parse a path does nto allow to specify the path ending up to store the local store in a file called 'NaN'

## Related issue(s)

--

## Confirmation

### DIFF
--- a/cli/ts/index.ts
+++ b/cli/ts/index.ts
@@ -635,7 +635,7 @@ program
 program
   .command("genLocalState")
   .description("generate a local MACI state from the smart contracts events")
-  .requiredOption("-o, --output <outputPath>", "the path where to write the state", parseInt)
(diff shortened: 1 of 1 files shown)


kittybest (commented):
good catch, just need to fix the commitlint

ctrlc03 (changes requested):
Thanks for the fix! could you please amend the commit to conform with commitlint and force push it? (also rebase to the latest dev changes please)

ctrlc03 (approved):
Thank you!

PR #1336, opened 2024-04-01
chore: use non-zero start and end blocks for proof generation

0xmad:
# Description

- [x] Update maci-contracts task
- [x] Update maci-cli task
- [x] Add indexed params for poll events
- [x] Change deploy config example maci public key

## Additional Notes

Now you don't need to specify start and end blocks. By default start block is a first signup event block; end - last merge aq event block. 

## Related issue(s)

MACI rpgf feedback discussion

## Confirmation

### DIFF
--- a/cli/ts/commands/genProofs.ts
+++ b/cli/ts/commands/genProofs.ts
@@ -240,10 +240,28 @@ export const genProofs = async ({
     }
   } else {
     // build an off-chain representation of the MACI contract using data in the contract storage
-    let fromBlock = startBlock ? Number(startBlock) : 0;
+    const [defaultStartBlockSignup, defaultStartBlockPoll, stateRoot, numSignups, messageRoot] = await Promise.all([
+      maciContract.queryFilter(maciContract.filters.SignUp()).then((events) => events[0]?.blockNumber ?? 0),
+      maciContract.queryFilter(maciContract.filters.DeployPoll()).then((events) => events[0]?.blockNumber ?? 0),
+      maciContract.getStateAqRoot(),
+      maciContract.numSignUps(),
+      messageAqContract.getMainRoot(messageTreeDepth),
+    ]);
+    const defaultStartBlock = Math.min(defaultStartBlockPoll, defaultStartBlockSignup);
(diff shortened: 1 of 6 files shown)


cli/ts/commands/genProofs.ts
```diff
@@ -240,10 +240,19 @@ export const genProofs = async ({
     }
   } else {
     // build an off-chain representation of the MACI contract using data in the contract storage
-    let fromBlock = startBlock ? Number(startBlock) : 0;
+    const defaultStartBlock = await maciContract
+      .queryFilter(maciContract.filters.SignUp())
```
ctrlc03, line 244: I think the problem with this approach is that if someone deploys a poll before the first signup is made, then any log before the first signup will be lost, like the poll deployment event.
My suggestion is that for the cli we can put it as mandatory param, and for the contract task we save (same way as we save all other data about the deployments) the Maci deployment block and use it later

0xmad: So, I added poll event parsing as well so if poll is deployed before signup, it takes poll event block and vice versa.

ctrlc03 (approved):
Great stuff, thank you

ctrlc03:
@0xmad I think we missed something here, and surprised tests did not catch it. We should be counting the start block as the block where the MACI contract was deployed. This is because we add the first leaf to the state tree inside the constructor, so if we start fetching logs from either the first signup, or when the first poll is deployed, then we will have a mismatch here: https://github.com/privacy-scaling-explorations/maci/blob/dev/contracts/ts/genMaciState.ts#L362

ctrlc03:
> @0xmad I think we missed something here, and surprised tests did not catch it. We should be counting the start block as the block where the MACI contract was deployed. This is because we add the first leaf to the state tree inside the constructor, so if we start fetching logs from either the first signup, or when the first poll is deployed, then we will have a mismatch here: https://github.com/privacy-scaling-explorations/maci/blob/dev/contracts/ts/genMaciState.ts#L362 will make a PR for it

nope actually what I just said is not true haha it's fine to fetch from the first signup or first poll deployment because anyways we are not emitting an event for the first leaf and replicating the same logic off chain - my bad

PR #1369, opened 2024-04-10
feat: use mode instead of flag for voting option

0xmad:
# Description

- [x] Add mode for contracts
- [x] Support cli changes

## Additional Notes

N/A

## Related issue(s)

Related #1355 

## Confirmation

### DIFF
--- a/contracts/contracts/utilities/DomainObjs.sol
+++ b/contracts/contracts/utilities/DomainObjs.sol
@@ -8,6 +8,12 @@ contract DomainObjs {
   /// @notice the length of a MACI message
   uint8 public constant MESSAGE_DATA_LENGTH = 10;
 
+  /// @notice voting modes
+  enum Mode {
+    QV,
+    NON_QV
+  }
+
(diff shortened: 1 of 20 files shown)


contracts/contracts/utilities/DomainObjs.sol
```diff
@@ -8,6 +8,12 @@ contract DomainObjs {
   /// @notice the length of a MACI message
   uint8 public constant MESSAGE_DATA_LENGTH = 10;
 
+  /// @notice voting modes
+  enum Mode {
```
kittybest, line 12: just a stupid question: why we call it EMode elsewhere, but use Mode here?

kittybest (commented):
Thanks!

contracts/contracts/utilities/DomainObjs.sol
0xmad: Because it's solidity, not js/ts

kittybest (approved):

PR #1408, opened 2024-04-29
fix: do not accept invalid maci keys

ctrlc03:
# Description

Ensure MACI keys are correctly validated to be a point on the babyjujub curve

## Confirmation

### DIFF
--- a/contracts/contracts/MACI.sol
+++ b/contracts/contracts/MACI.sol
@@ -13,6 +13,7 @@ import { Params } from "./utilities/Params.sol";
 import { TopupCredit } from "./TopupCredit.sol";
 import { Utilities } from "./utilities/Utilities.sol";
 import { DomainObjs } from "./utilities/DomainObjs.sol";
+import { CurveBabyJubJub } from "./crypto/BabyJubJub.sol";
 import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 
 /// @title MACI - Minimum Anti-Collusion Infrastructure Version 1
@@ -101,7 +102,7 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable(msg.sender) {
   error CallerMustBePoll(address _caller);
   error PoseidonHashLibrariesNotLinked();
   error TooManySignups();
-  error MaciPubKeyLargerThanSnarkFieldSize();
+  error InvalidPubKey();
   error PreviousPollNotCompleted(uint256 pollId);
   error PollDoesNotExist(uint256 pollId);
   error SignupTemporaryBlocked();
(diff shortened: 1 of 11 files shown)


0xmad (commented):
@ctrlc03 thanks, just one question: can we add this check for cli, so user don't need to initiate the transaction?

contracts/contracts/MACI.sol
```diff
@@ -168,8 +169,9 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable(msg.sender) {
     // ensure we do not have more signups than what the circuits support
     if (numSignUps >= uint256(TREE_ARITY) ** uint256(stateTreeDepth)) revert TooManySignups();
 
-    if (_pubKey.x >= SNARK_SCALAR_FIELD || _pubKey.y >= SNARK_SCALAR_FIELD) {
-      revert MaciPubKeyLargerThanSnarkFieldSize();
+    // ensure that the public key is on the baby jubjub curve
+    if (!CurveBabyJubJub.isOnCurve(_pubKey.x, _pubKey.y)) {
```
kittybest, line 173: this includes the case that insure `_pubKey.x < SNARK_SCALAR_FIELD && _pubKey.y < SNARK_SCALAR_FIELD` right?

contracts/contracts/crypto/BabyJubJub.sol
```diff
@@ -0,0 +1,138 @@
...
+  // A = 168700
+  uint256 public constant A = 0x292FC;
+  // D = 168696
+  uint256 public constant D = 0x292F8;
+  // Prime Q = 21888242871839275222246405745257275088548364400416034343698204186575808495617
+  uint256 public constant Q = 0x30644E72E131A029B85045B68181585D2833E84879B9709143E1F593F0000001;
+
+  /**
+   * @dev Add 2 points on baby jubjub curve
+   * Formulae for adding 2 points on a twisted Edwards curve:
```
kittybest, line 22: nit: Formulae -> Formulate

kittybest (commented):
thanks!

contracts/contracts/crypto/BabyJubJub.sol
ctrlc03: mm thinking it was supposed to be "formula" instead?

contracts/contracts/MACI.sol
ctrlc03: yup

contracts/contracts/crypto/BabyJubJub.sol
kittybest: I think you are right😂

domainobjs/ts/publicKey.ts
```diff
@@ -19,11 +19,18 @@ export class PubKey {
 
   /**
    * Create a new instance of a public key
+   * @dev You might want to allow an invalid raw key,
```
kittybest, line 22: I think I don’t quite understand this part. Why decrypting invalid message should use invalid raw key?

ctrlc03: Basically imagine you send an invalid ciphertext (the encrypted message) and a key which wasn't used to encrypt it. when decrypting the message forcefully, the data will be invalid because the key is not able to decrypt it, but we still choose to get the decryption result. And because we need to forcefully decrypt messages here https://github.com/privacy-scaling-explorations/maci/blob/dev/core/ts/Poll.ts#L554 the `Command.Decrypt` operation would fail when creating the public key https://github.com/privacy-scaling-explorations/maci/blob/dev/domainobjs/ts/commands/PCommand.ts#L210

0xmad (approved):
@ctrlc03 thanks!

crisgarner (approved):

PR #1417, opened 2024-05-02
feat(multiple-polls): allow concurrent polls

### LINKED ISSUE #1418: Allow multiple (and concurrent) Polls
Currently in MACI, it is not possible to re use the same Maci contract for multiple polls due to the AccQueue structure inner working. After a merge operation, the index would skip ahead to the smallest subtree capacity, and this would not be replicated offchain causing state root mismatches. As the modification would include quite some changes to the logic, and these structures might be removed soon from MACI, we should look into fixing this in another way. 

A proposed solution is to use a LazyIMT structure https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/imt.sol/contracts/LazyIMT.sol which is cheap to use and does not require any merge operation.

ctrlc03:
# Description

* use lazyIMT for signups to allow for concurrent polls
* add tests
* amend maci components to support the change

Benefits:
* cheaper signups overall
* less code
* no merge required
* concurrent polls

Drawbacks:
* new trusted setup needed
* slightly more constraints in the circuits

## Related issue(s)

fix #1418 

## Confirmation


circuits/circom/trees/incrementalMerkleTree.circom
```diff
@@ -104,10 +106,93 @@ template CheckRoot(levels) {
...
-}
\ No newline at end of file
+}
+
+/**
+ * Calculates the path indices required for Merkle proof verifications. 
+ * Given a node index within an IMT and the total tree levels, it outputs the path indices leading to that node.
+ * The template handles the modulo and division operations to break down the tree index into its constituent path indices.
+ */
+template MerkleGeneratePathIndices(levels) {
```
0xmad, line 122: Maybe let's move it to a separate file?

circuits/circom/trees/incrementalMerkleTree.circom
```diff
@@ -104,10 +106,93 @@ template CheckRoot(levels) {
...
+    n[levels] <-- m;
+
+    for (var i = 0; i < levels; i++) {
+        // Check that each output element is less than the base.
+        var computedIsOutputElementLessThanBase = SafeLessThan(3)([out[i], BASE]);
+        computedIsOutputElementLessThanBase === 1;
+
+        // Re-compute the total sum.
+        computedResults[i] = out[i] * (BASE ** i);
+    }
```
0xmad, line 149: Can we combine these two loops?

circuits/circom/trees/incrementalMerkleTree.circom
```diff
@@ -104,10 +106,93 @@ template CheckRoot(levels) {
...
+    var root = 0;
+
+    for (var i = 0; i < MAX_DEPTH; i++) {
+        var isDepth = IsEqual()([depth, i]);
+
+        roots[i] <== isDepth * nodes[i];
+
+        root += roots[i];
+
+        var c[2][2] = [ [nodes[i], siblings[i][0]], [siblings[i][0], nodes[i]] ];
```
0xmad, line 189:
```suggestion
        var c[2][2] = [
            [nodes[i], siblings[i][0]],
            [siblings[i][0], nodes[i]]
        ];
```

contracts/contracts/Poll.sol
```diff
@@ -224,17 +228,6 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable(msg.sender), EmptyBallo
...
-    if (stateAqMerged) revert StateAqAlreadyMerged();
-
-    // merge subroots
-    extContracts.maci.mergeStateAqSubRoots(_numSrQueueOps, _pollId);
-
-    emit MergeMaciStateAqSubRoots(_numSrQueueOps);
-  }
-
   /// @inheritdoc IPoll
   function mergeMaciStateAq(uint256 _pollId) public onlyOwner isAfterVotingDeadline {
```
0xmad, line 232: Can we remove this _pollId arg?

contracts/contracts/MACI.sol
```diff
@@ -259,26 +234,8 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable(msg.sender) {
...
-    subtreesMerged = false;
-
-    root = stateAq.merge(stateTreeDepth);
-  }
-
-  /// @inheritdoc IMACI
-  function getStateAqRoot() public view returns (uint256 root) {
-    root = stateAq.getMainRoot(stateTreeDepth);
+  function getStateTreeRoot() public view returns (uint256 root) {
+    root = InternalLazyIMT._root(lazyIMTData);
```
0xmad, line 238: Why do these methods for IMT start with underscore?

contracts/contracts/trees/LazyIMT.sol
```diff
@@ -0,0 +1,222 @@
...
+    }
+    return _root(self, numberOfLeaves, depth);
+  }
+
+  /// @notice Returns the root of the LazyIMT
+  /// @param self The LazyIMTData
+  /// @param depth The depth of the tree
+  /// @return The root of the LazyIMT
+  function _root(LazyIMTData storage self, uint8 depth) internal view returns (uint256) {
+    require(depth > 0, "LazyIMT: depth must be > 0");
```
0xmad, line 158: Let's use custom errors for this file

contracts/tasks/helpers/TreeMerger.ts
0xmad, line 1: Is it right that IMT is used only for signups?
What's about messages? Are we good to use AccQueue for it or you have a plan to migrate to IMT later?

contracts/tasks/runner/merge.ts
```diff
@@ -60,9 +53,6 @@ task("merge", "Merge signups and messages")
     await treeMerger.checkPollOwner();
     await treeMerger.checkPollDuration();
 
-    await treeMerger.mergeSignupSubtrees(poll, queueOps);
-    await treeMerger.mergeSignups(poll);
```
0xmad, line 64: Do we need to merge signups here or no?

contracts/tests/MACI.test.ts
```diff
@@ -200,7 +185,24 @@ describe("MACI", () => {
...
         ),
-      ).to.be.revertedWithCustomError(maci, "TooManySignups");
+      ).to.be.revertedWithCustomError(maciContract, "TooManySignups");
+    });
+
+    it("should signup 2 ** 10 users", async () => {
+      const stateTreeDepthTest = 10;
+      const maxUsers = 2 ** stateTreeDepthTest;
+      const maci = (await deployTestContracts(initialVoiceCreditBalance, stateTreeDepthTest, signer, true))
+        .maciContract;
```
0xmad, line 195:
```suggestion
      const { maciContract: maci } = await deployTestContracts(initialVoiceCreditBalance, stateTreeDepthTest, signer, true);
```

cli/ts/commands/genProofs.ts
```diff
@@ -155,8 +155,8 @@ export const genProofs = async ({
   const messageAqContractAddr = extContracts.messageAq;
   const messageAqContract = AccQueueFactory.connect(messageAqContractAddr, signer);
 
-  // Check that the state and message trees have been merged for at least the first poll
-  if (!(await pollContract.stateAqMerged()) && pollId.toString() === "0") {
```
0xmad, line 159: Can you add this change for maci-contracts proof generator?

0xmad (changes requested):
@ctrlc03 thanks, just added some comments and questions

contracts/contracts/Poll.sol
ctrlc03: good shout, totally

contracts/contracts/MACI.sol
ctrlc03: cause they are internal functions of a library

0xmad: Is there a reason why they are internal? Maybe it makes sense to change them to public like for poseidon libs?

ctrlc03: https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/imt.sol/contracts/internal/InternalLazyIMT.sol

contracts/tasks/helpers/TreeMerger.ts
ctrlc03: For now only for signups, then we can use for messages too, but by then we would probably already have the hash chain implemented by the 3227 team on their proposal for anon poll signups + hash chain

contracts/tasks/runner/merge.ts
ctrlc03: we do yes, will put back

cli/ts/commands/genProofs.ts
ctrlc03: sure

circuits/circom/trees/incrementalMerkleTree.circom
ctrlc03: let me do this in a separate PR, I saw that the one in `incrementalQuinTree` is the same and can be simplified as you suggested + remove unused variables

circuits/circom/trees/incrementalMerkleTree.circom
ctrlc03: I think we could keep it inside this file - We have all binary tree templates in this, and all quinary tree templates in `incrementalQuinTree`.  But perhaps you had a more detailed structure in mind? open to moving stuff around if we apply it across all other imt files

contracts/contracts/MACI.sol
ctrlc03: right so I'm trying to convert them to public, the next step is to use the directive `use InternalLazyIMT for lazyIMTData` -> though for this I would need to create a wrapper contract which calls these functions, deploy with linked library, then inherit from this contract within MACI, when we can directly call the library 🤔  Reckon it's something worth doing?

circuits/circom/trees/incrementalMerkleTree.circom
0xmad: It's just because of code style like: one file for one class (or in our case template).

circuits/circom/trees/incrementalMerkleTree.circom
0xmad: Sure, it works

contracts/contracts/MACI.sol
0xmad: Let's keep it simple instead :)

circuits/circom/trees/incrementalMerkleTree.circom
ctrlc03: What you think about then having two folders inside `circuits/circom/trees` one for `quinTree` and one for `binaryTree` and inside these split the templates one per file?

0xmad: Yes, it works!

cli/tests/ceremony-params/ceremonyParams.test.ts
```diff
@@ -128,7 +128,7 @@ describe("Stress tests with ceremony params (6,9,2,3)", function test() {
       await setVerifyingKeys({ ...verifyingKeysArgs, signer });
     });
 
-    describe("1 user, 2 messages", () => {
+    describe.only("1 user, 2 messages", () => {
```
0xmad, line 131: Don't forget to remove it

contracts/contracts/trees/LazyIMT.sol
```diff
@@ -0,0 +1,241 @@
...
+  /// @return The root of the LazyIMT
+  function _root(LazyIMTData storage self, uint8 depth) internal view returns (uint256) {
+    if (depth == 0) {
+      revert DepthCannotBeZero();
+    }
+    if (depth > MAX_DEPTH) {
+      revert DepthTooLarge();
+    }
+    uint40 numberOfLeaves = self.numberOfLeaves;
+    require(uint40(2) ** uint40(depth) >= numberOfLeaves, "LazyIMT: ambiguous depth");
```
0xmad, line 173: Let's use custom error here as well

circuits/circom/trees/incrementalMerkleTree.circom
ctrlc03: opened issue here #1419

circuits/circom/trees/incrementalMerkleTree.circom
ctrlc03: opened issue here #1420

0xmad (approved):
@ctrlc03 thanks!
I'll test it when new zkeys are available, for local env looks good to me.

ctrlc03:
> @ctrlc03 thanks! I'll test it when new zkeys are available, for local env looks good to me.

gonna build them and push to s3 :)

PR #1426, opened 2024-05-06
refactor: general cleanup, variables/events/functions renaming

ctrlc03:
# Description

Ensure functions/events/variables/test descriptions reflect the latest code changes optimize contracts and remove unnecessary logic

- [x] Remove ownable from MACI - anyone can deploy a poll now and they all are independent
- [x] Remove ownable from Poll - anyone can technically merge the message tree if they are willing to pay the gas for it
- [x] Remove transferOwnership in MpFactory and TallyFactory - directly use caller from Maci.deployPoll
- [x] Rename functions/events to reflect removal of state acc queue
- [x] Remove dead code
- [x] Add tests to ensure above changes do not affect logic

## Confirmation

### DIFF
--- a/contracts/contracts/MACI.sol
+++ b/contracts/contracts/MACI.sol
@@ -12,21 +12,19 @@ import { TopupCredit } from "./TopupCredit.sol";
 import { Utilities } from "./utilities/Utilities.sol";
 import { DomainObjs } from "./utilities/DomainObjs.sol";
 import { CurveBabyJubJub } from "./crypto/BabyJubJub.sol";
-import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 import { InternalLazyIMT, LazyIMTData } from "./trees/LazyIMT.sol";
 
 /// @title MACI - Minimum Anti-Collusion Infrastructure Version 1
 /// @notice A contract which allows users to sign up, and deploy new polls
-contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable(msg.sender) {
+contract MACI is IMACI, DomainObjs, Params, Utilities {
   /// @notice The state tree depth is fixed. As such it should be as large as feasible
-  /// so that there can be as many users as possible.  i.e. 5 ** 10 = 9765625
+  /// so that there can be as many users as possible.  i.e. 2 ** 23 = 8388608
   /// this should also match the parameter of the circom circuits.
-  uint8 public immutable stateTreeDepth;
-
   /// @notice IMPORTANT: remember to change the ballot tree depth
   /// in contracts/ts/genEmptyBallotRootsContract.ts file
   /// if we change the state tree depth!
-  uint8 internal constant STATE_TREE_SUBDEPTH = 2;
+  uint8 public immutable stateTreeDepth;
+
   uint8 internal constant TREE_ARITY = 2;
   uint8 internal constant MESSAGE_TREE_ARITY = 5;
 
@@ -86,20 +84,11 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable(msg.sender) {
     PollContracts pollAddr
   );
 
-  /// @notice Only allow a Poll contract to call the modified function.
-  modifier onlyPoll(uint256 _pollId) {
-    if (msg.sender != address(polls[_pollId])) revert CallerMustBePoll(msg.sender);
-    _;
-  }
-
   /// @notice custom errors
-  error CallerMustBePoll(address _caller);
   error PoseidonHashLibrariesNotLinked();
   error TooManySignups();
   error InvalidPubKey();
-  error PreviousPollNotCompleted(uint256 pollId);
   error PollDoesNotExist(uint256 pollId);
-  error SignupTemporaryBlocked();
 
   /// @notice Create a new instance of the MACI contract.
   /// @param _pollFactory The PollFactory contract
@@ -190,7 +179,7 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable(msg.sender) {
     address _verifier,
     address _vkRegistry,
     Mode _mode
-  ) public virtual onlyOwner returns (PollContracts memory pollAddr) {
+  ) public virtual returns (PollContracts memory pollAddr) {
     // cache the poll to a local variable so we can increment it
     uint256 pollId = nextPollId;
 
@@ -210,20 +199,13 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable(msg.sender) {
       maxVoteOptions: uint256(MESSAGE_TREE_ARITY) ** _treeDepths.voteOptionTreeDepth
     });
 
-    address _owner = owner();
+    // the owner of the message processor and tally contract will be the msg.sender
+    address _msgSender = msg.sender;
 
-    address p = pollFactory.deploy(
-      _duration,
-      maxValues,
-      _treeDepths,
-      _coordinatorPubKey,
-      address(this),
-      topupCredit,
-      _owner
-    );
+    address p = pollFactory.deploy(_duration, maxValues, _treeDepths, _coordinatorPubKey, address(this), topupCredit);
 
-    address mp = messageProcessorFactory.deploy(_verifier, _vkRegistry, p, _owner, _mode);
-    address tally = tallyFactory.deploy(_verifier, _vkRegistry, p, mp, _owner, _mode);
+    address mp = messageProcessorFactory.deploy(_verifier, _vkRegistry, p, _msgSender, _mode);
+    address tally = tallyFactory.deploy(_verifier, _vkRegistry, p, mp, _msgSender, _mode);
 
     polls[pollId] = p;
 
--- a/contracts/contracts/Poll.sol
+++ b/contracts/contracts/Poll.sol
@@ -1,11 +1,11 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.20;
 
-import { Params } from "./utilities/Params.sol";
-import { SnarkCommon } from "./crypto/SnarkCommon.sol";
-import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import { Params } from "./utilities/Params.sol";
+import { SnarkCommon } from "./crypto/SnarkCommon.sol";
 import { EmptyBallotRoots } from "./trees/EmptyBallotRoots.sol";
 import { IPoll } from "./interfaces/IPoll.sol";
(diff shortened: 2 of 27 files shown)


contracts/tests/MACI.test.ts
```diff
@@ -259,6 +259,39 @@ describe("MACI", function test() {
...
+        vkRegistryContract,
+        EMode.QV,
+        { gasLimit: 10000000 },
+      );
+      const receipt = await tx.wait();
+      expect(receipt?.status).to.eq(1);
+      expect(await maciContract.nextPollId()).to.eq(2);
+    });
+
+    it("should allow any user to deploy a poll", async () => {
```
kittybest, line 278: I'm just curious that why we wanna allow any user to deploy poll?

contracts/contracts/Poll.sol
```diff
@@ -80,8 +80,7 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable(msg.sender), EmptyBallo
 
   event PublishMessage(Message _message, PubKey _encPubKey);
   event TopupMessage(Message _message);
-  event MergeMaciStateAqSubRoots(uint256 indexed _numSrQueueOps);
-  event MergeMaciStateAq(uint256 indexed _stateRoot, uint256 indexed _numSignups);
+  event MergeMaciState(uint256 indexed _stateRoot, uint256 indexed _numSignups);
```
kittybest, line 83: just curious: why remove `Aq` here?

contracts/tests/MACI.test.ts
ctrlc03: actually good question, I removed the modifier as now each poll is independent from each other and can run concurrently, so it doesn't really matter who deploys it (+ we save a tiny bit of gas), it will have its own set of contracts deployed and their own coordinator key. Anyone could now re use the same signups (so users which passed a certain anti-sybil requirements) and run their poll for these users 
Do you think there's a reason why we should keep it `onlyOwner`?

kittybest: No I'm just curious why you consider to make this change, also might change little bit to the UI display of maci-rpgf design. Agree with that.

contracts/contracts/Poll.sol
kittybest: seems like a long story I just started reading 😨 thanks

contracts/tests/MACI.test.ts
ctrlc03: I could see an argument where it might be best to keep it callable only by the owner, if for instance you don't want other users from deploying polls from your contract, and filling up you state tree (assuming the gatekeeping requirements are not very custom, like for instance proof of humanhood vs owning a NFT)

kittybest: that's also a good way, but would need to add an argument here and affecting a lot of codes calling this function 🤔 not sure if it's worth doing this.

contracts/contracts/MACI.sol
```diff
@@ -190,7 +179,7 @@ contract MACI is IMACI, DomainObjs, Params, Utilities, Ownable(msg.sender) {
     address _verifier,
     address _vkRegistry,
     Mode _mode
-  ) public virtual onlyOwner returns (PollContracts memory pollAddr) {
+  ) public virtual returns (PollContracts memory pollAddr) {
```
0xmad, line 182: Is it expected that everyone can deploy polls?

contracts/contracts/Poll.sol
```diff
@@ -259,17 +258,17 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable(msg.sender), EmptyBallo
 
     actualStateTreeDepth = depth;
 
-    emit MergeMaciStateAq(mergedStateRoot, numSignups);
+    emit MergeMaciState(mergedStateRoot, numSignups);
   }
 
   /// @inheritdoc IPoll
-  function mergeMessageAqSubRoots(uint256 _numSrQueueOps) public onlyOwner isAfterVotingDeadline {
+  function mergeMessageAqSubRoots(uint256 _numSrQueueOps) public isAfterVotingDeadline {
```
0xmad, line 265: Just an item for discussion: we need to think about access control. After coordinator work, I can start working on this.

0xmad (commented):
@ctrlc03 thanks, just one question but everything looks good

contracts/contracts/MACI.sol
0xmad: The only one concern here: poll finalization depends on block timestamp and if someone can manipulate with it, potentially poll can be finished before the deadline.

contracts/contracts/Poll.sol
ctrlc03: Yes totally a good discussion to have. My thoughts here are: it doesn't really matter who calls these functions, first of all they can cost quite a lot of gas, so technically if someone wants to call them, they are doing the coordinator/operator a favour. Second, these functions can only be called after the voting period ended, so no message would be able to be sent anymore either way, so it would not influence the voting period.

contracts/contracts/MACI.sol
ctrlc03: > Is it expected that everyone can deploy polls?

Up to discussion here. I think it could open up scenarios where single MACI instances are re-used by different teams. Imagine Someone uses  MACI for a vote which is gatekept by an event ticket, let's say Devcon. Now imagine each sub-event (or side event) in Devcon wants to launch their own poll (for whatever reason). With one contract only, individual teams can launch completely independent polls (each of them would have a different set of Poll/Mp/Tally contracts) but all of them would be able to re use the already signed up users which are known to have passed the sybil checks at the gatekeeper level.

ctrlc03: > The only one concern here: poll finalization depends on block timestamp and if someone can manipulate with it, potentially poll can be finished before the deadline.

So you are referring to a block proposer/validator which would try to validate a block with a different block.timestamp to be able to "complete" a poll earlier? I guess eventually there might even be financial incentives to do that, though this would still be a problem (less of a problem of course) with the `onlyOwner` modifier, as it would need to be coordinated between the owner and validator. Perhaps in general relying on block numbers vs block timestamp might be a safer solution. We can keep `onlyOwner` for tree merging for now, and think more about it.

kittybest: is that possible that we could have a list of addresses that could manipulate on the contracts? in the above example (devcon), who need to deploy their own poll would know beforehand, thus they could ask the maci deployer to put them on the whitelist and they are the ones who could deploy polls? limit the risk to trusted addresses.

ctrlc03: Sure there are access control libraries which you can use to give multiple roles or same role to more people. Though what exactly is the risk of someone deploying a poll? Before we had only owner because you could only have one poll at a time, so if someone deployed a poll before you (contract deployer) could then you'd need a new contract, but now there's not that problem anymore

kittybest: I don't know if there's any risk now. actually I'm just thinking about something maybe unrelated to these discussions, that if we wanna allow anyone to deploy a poll, but the EASGatekeeper still only accept attestation from a certain address, would that be a problem? maybe would be good to keep a list of addresses for EASGatekeeper?

ctrlc03: > I don't know if there's any risk now. actually I'm just thinking about something maybe unrelated to these discussions, that if we wanna allow anyone to deploy a poll, but the EASGatekeeper still only accept attestation from a certain address, would that be a problem? maybe would be good to keep a list of addresses for EASGatekeeper?

Good questions to bring up, thanks for that. So let's start with the gatekeepers. Any solution should not be focused on a particular gatekeeper like the EAS for instance, so we should talk gatekeepers in general. After all, a Gatekeeper is a contract that given some inputs, either reverts or gives back execution to the MACI contract if they believe the user should be allowed to vote. 

When you say "maybe would be good to keep a list of addresses for EASGatekeeper?" do you mean that we should have more gatekeepers on MACI's side or that an EAS gatekeeper should have a list of trusted attesters vs only one?

Second, just because everyone could deploy a poll from a MACI instance, it doesn't mean that they should nor that they will. They might not know that there's a MACI instance deployed there, or if they do know and they see that the instance uses an EAS gatekeeper, well it's their choice whether they want to run a poll that only users with that attestation can access, or instead they could deploy their own Maci instance. 
In terms of benefit for the original MACI deployers, this comes down to smaller contract due to Ownable not being included, as well as a tiny saving on the `deployPoll` function as there's no `onlyOwner` modifier.

kittybest: > When you say "maybe would be good to keep a list of addresses for EASGatekeeper?" do you mean that we should have more gatekeepers on MACI's side or that an EAS gatekeeper should have a list of trusted attesters vs only one?

I'm thinking about this: *an EAS gatekeeper should have a list of trusted attesters*.
That's why I think this is not quite related to this PR, only related to how we implement EASGatekeeper.

> Second, just because everyone could deploy a poll from a MACI instance, it doesn't mean that they should nor that they will. They might not know that there's a MACI instance deployed there, or if they do know and they see that the instance uses an EAS gatekeeper, well it's their choice whether they want to run a poll that only users with that attestation can access, or instead they could deploy their own Maci instance.

That's quite great to imagine about. Users could deploy their own rounds in a, let's say, certain event, making voting occurs more often. Even though some small things (e.g. what's for dinner? ...) are not quite related to using MACI (?), but it would be interesting that we could do several voting rounds in a group (e.g. devcon attendees, students from class 309, ...) safe and sound.

contracts/contracts/Poll.sol
```diff
@@ -229,13 +228,13 @@ contract Poll is Params, Utilities, SnarkCommon, Ownable(msg.sender), EmptyBallo
   }
 
   /// @inheritdoc IPoll
-  function mergeMaciStateAq() public onlyOwner isAfterVotingDeadline {
+  function mergeMaciState() public isAfterVotingDeadline {
     // This function can only be called once per Poll after the voting
     // deadline
-    if (stateAqMerged) revert StateAqAlreadyMerged();
+    if (stateMerged) revert StateAqAlreadyMerged();
```
baumstern, line 234:
```suggestion
    if (stateMerged) revert StateAlreadyMerged();
```

contracts/contracts/MACI.sol
ctrlc03: > I'm thinking about this: _an EAS gatekeeper should have a list of trusted attesters_. That's why I think this is not quite related to this PR, only related to how we implement EASGatekeeper.

right yea we can have a EASGatekeeperMultipleAttester contract which has multiple attester. That could be useful but needs to be used with caution.

> That's quite great to imagine about. Users could deploy their own rounds in a, let's say, certain event, making voting occurs more often. Even though some small things (e.g. what's for dinner? ...) are not quite related to using MACI (?), but it would be interesting that we could do several voting rounds in a group (e.g. devcon attendees, students from class 309, ...) safe and sound.

Yea totally a good use case, which is possible now but only with 1 gatekeeper at a time (for now)

contracts/contracts/Poll.sol
ctrlc03: good catch thank you

PR #1462, opened 2024-05-14
fix(nightly): checkout tag v1.2.1 on nightly-ceremony workflow

ctrlc03:
# Description

Checkout v1.2.1 instead of dev for nightly-ceremony workflow

## Related issue(s)

re #1414 

## Confirmation

- [ ] I have read and understand MACI's [GitHub processes](https://github.com/privacy-scaling-explorations/maci/discussions/847).
- [ ] I have read and understand MACI's [testing guide](https://maci.pse.dev/docs/testing).


.github/workflows/nightly-ceremony.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: dev
+          ref: v1.2.1
```
0xmad, line 17: Can we use wildcard here or we don't need to support previous and future versions?

0xmad (approved):
@ctrlc03 thanks, just one question.

.github/workflows/nightly-ceremony.yml
ctrlc03: so these ceremony zkeys only work with 1.2.0 and 1.2.1, now with dev we have changes to the circuits and require a new ceremony. But we should probably get an audit first. 
And the problem here is that unless we publish a 1.2.2 (starting from 1.2.1 and not current dev) with the ceremony test fix we won't be able to run successfully these tests 🤔

PR #1507, opened 2024-05-23
feat: coordinator public key method

0xmad:
# Description

- [x] Add api method for getting rsa public key
- [x] Use dependency injection for crypto service
- [x] Move file related logic to a separate service
- [x] Add explicit public by-pass for authorization

## Additional Notes

N/A

## Related issue(s)

Closes https://github.com/privacy-scaling-explorations/maci-rpgf/issues/58

## Confirmation

### DIFF
--- a/coordinator/ts/file/file.service.ts
+++ b/coordinator/ts/file/file.service.ts
@@ -1,0 +1,81 @@
+import { Injectable, Logger } from "@nestjs/common";
+
+import fs from "fs";
+import path from "path";
+
+import type { IGetPrivateKeyData, IGetPublicKeyData, IGetZkeyFilePathsData } from "./types";
+
+import { ErrorCodes } from "../common";
+
+/**
+ * FileService is responsible for working with local files like:
+ * 1. RSA public/private keys
+ * 2. Zkey files
+ */
+@Injectable()
+export class FileService {
+  /**
+   * Logger
+   */
+  private readonly logger: Logger;
+
+  /**
+   * Initialize service
+   */
+  constructor() {
+    this.logger = new Logger(FileService.name);
+  }
+
+  /**
+   * Get RSA private key for coordinator service
+   *
+   * @returns serialized RSA public key
+   */
+  async getPublicKey(): Promise<IGetPublicKeyData> {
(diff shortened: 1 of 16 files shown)


coordinator/ts/file/file.service.ts
```diff
@@ -0,0 +1,81 @@
...
+   */
+  getZkeyFilePaths(name: string, useQuadraticVoting: boolean): IGetZkeyFilePathsData {
+    const root = path.resolve(process.env.COORDINATOR_ZKEY_PATH!);
+    const index = name.indexOf("_");
+    const type = name.slice(0, index);
+    const params = name.slice(index + 1);
+    const mode = useQuadraticVoting ? "" : "NonQv";
+    const filename = `${type}${mode}_${params}`;
+
+    const zkey = path.resolve(root, `${filename}/${filename}.0.zkey`);
```
ctrlc03, line 66: something to note here is that we add this `0` to the name on test files, but in general the name could be anything. So maybe we can remove the `0` and just take the filename?

coordinator/ts/file/file.service.ts
```diff
@@ -0,0 +1,81 @@
...
+    const wasm = path.resolve(root, `${filename}/${filename}_js/${filename}.wasm`);
+    const witgen = path.resolve(root, `${filename}/${filename}_cpp/${filename}`);
+
+    if (!fs.existsSync(zkey) || !fs.existsSync(wasm) || !fs.existsSync(witgen)) {
```
ctrlc03, line 70: I think it might be ok if there's either the wasm or c witness gen - in the end I think we would probably use wasm for testing, as we can get much faster proof generation with the c witness gen

ctrlc03 (changes requested):
Thanks, left some comments

coordinator/ts/file/file.service.ts
0xmad: So the reason I added it, it downloads with this name but wasm and witgen doesn't have this `.0`.

ctrlc03: ok right we can then maybe remove from the test artifacts and keep it here for now

ctrlc03 (approved):
Thanks!

PR #1631, opened 2024-07-05
feat(contracts): add first registry and payout extensions

### LINKED ISSUE #1633: Implement a simple project registry
Implement a simple project registry to hold the addresses (or data) of projects signed up to MACI. This way applications like maci-rpgf can use this contract to provide an immutable order of vote options

### LINKED ISSUE #1634: Implement a simple payout contract
Implement, as an extension to MACI's core protocol, a contract which can be used to distribute funds according to a MACI Poll results.

ctrlc03:
# Description

Add an extensions folder inside contracts and add a simple implementation of a registry and a payout contract.

## Additional Info

Can merge after v2.0 release

## Related Issues

fix #1634 
fix #1633 

## Confirmation


contracts/contracts/extensions/payout/SimplePayout.sol
```diff
@@ -0,0 +1,188 @@
...
+    // Handle Ether payment
+    if (address(this).balance < _amount) revert NotEnoughEther();
+    uint256 gas = gasleft();
+    (bool success, ) = _dest.call{ value: _amount, gas: gas }("");
+    // If the Ether transfer fails, wrap the Ether and try to send it as WETH.
+    if (!success) {
+      WETH.deposit{ value: _amount }();
+      IERC20(address(WETH)).safeTransfer(_dest, _amount);
+    }
+  }
```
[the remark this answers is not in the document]
ctrlc03: This is used so if there is a malicious recipient which reverts on receive() we handle this case by trying catching the revert, and wrapping ether to WETH so we can force transfer to them

contracts/contracts/extensions/payout/SimplePayout.sol
```diff
@@ -0,0 +1,190 @@
...
+
+/// @title SimplePayout
+/// @notice A simple payout contract that works with a SimpleProjectRegistry
+/// and a MACI Tally contract
+contract SimplePayout is Ownable(msg.sender) {
+  /// @notice use safe ERC20 functions
+  using SafeERC20 for IERC20;
+
+  /// @notice WETH address
+  IWETH public immutable WETH;
```
0xmad, line 20: Maybe we can call it `weth`?

contracts/contracts/extensions/payout/SimplePayout.sol
```diff
@@ -0,0 +1,190 @@
...
+    totalVoiceCreditsSpent = _totalSpent;
+  }
+
+  /// @notice Deposit the amount of tokens to the contract
+  /// @dev This function is only callable by the owner
+  function deposit() external payable onlyOwner {
+    if (payoutToken != address(0)) {
+      // transfer the ERC20 token amount
+      IERC20(payoutToken).safeTransferFrom(msg.sender, address(this), totalAmount);
+    }
```
0xmad, line 99: If contract is deployed with `payoutToken = address(0)` this deposit will take no effect.
Is it expected behaviour?

contracts/contracts/extensions/payout/SimplePayout.sol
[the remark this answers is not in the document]
0xmad: Is it possible if project from registry can be replaced with something different?
And is there a reason why you use `address#call`?

contracts/contracts/gatekeepers/SignUpTokenGatekeeper.sol
```diff
@@ -2,16 +2,17 @@
...
 import { SignUpGatekeeper } from "./SignUpGatekeeper.sol";
-import { SignUpToken } from "../SignUpToken.sol";
 
 /// @title SignUpTokenGatekeeper
 /// @notice This contract allows to gatekeep MACI signups
 /// by requiring new voters to own a certain ERC721 token
 contract SignUpTokenGatekeeper is SignUpGatekeeper, Ownable(msg.sender) {
   /// @notice the reference to the SignUpToken contract
-  SignUpToken public token;
+  ERC721 public immutable token;
```
0xmad, line 14: Do we have any plans for supporting other token types?

contracts/contracts/mocks/MockWETH.sol
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    (bool success, ) = msg.sender.call{ value: amount }("");
+    if (!success) revert WithdrawFailed();
+  }
+
+  /// @notice Transfer ETH
+  /// @param to The address to transfer to
+  /// @param amount The amount to transfer
+  function transfer(address to, uint256 amount) external {
```
0xmad, line 64: Can your return boolean to match ERC20 interface?

contracts/tests/extensions/SimplePayout.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    /*
+            uint256 _voteOptionIndex,
+            uint256 _spent,
+            uint256[][] calldata _proof,
+            uint256 _spentSalt,
+            uint256 _resultsCommitment,
+            uint256 _spentVoiceCreditsCommitment,
+            uint8 _voteOptionTreeDepth
+        */
```
0xmad, line 107: Can you align it or remove?

contracts/tests/extensions/SimpleProjectRegistry.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  let owner: Signer;
+  let user: Signer;
+
+  let ownerAddress: string;
+  let userAddress: string;
+
+  before(async () => {
+    [owner, user] = await getSigners();
+    ownerAddress = await owner.getAddress();
+    userAddress = await user.getAddress();
```
0xmad, line 18:
```suggestion
    [ownerAddress, userAddress] = await Promise.all([owner.getAddress(), user.getAddress()]);
```

0xmad (changes requested):
@ctrlc03 thanks, just left some comments

contracts/contracts/extensions/payout/SimplePayout.sol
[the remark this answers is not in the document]
ctrlc03: not sure what exactly you are referring by "replaced with something different"?
I use call so there is more flexibility on what the receiver can do in their receive/fallback, and also on the sender side + I believe call is the recommended method of sending Ether nowawayds. I remember this particular pattern from my auditing days, as contracts can revert on receive and if they did and we don't handle that, we will never be able to complete the payout. 
Any other way you'd do this?

contracts/contracts/extensions/payout/SimplePayout.sol
ctrlc03: the function is payable so if payoutToken = address(0) then it just accepts the ether. 

but actually we should have a check that msg.value == totalAmount so I'll add that

contracts/contracts/gatekeepers/SignUpTokenGatekeeper.sol
ctrlc03: mm not sure, this one was meant to work with ERC721 specifically, but eventually we should be getting rid of gatkeepers within MACI and using Excubiae's https://github.com/privacy-scaling-explorations/zk-kit.solidity/tree/main/packages/excubiae

contracts/contracts/extensions/payout/SimplePayout.sol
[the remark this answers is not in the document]
0xmad: I think it's fine regarding `call` so 
By "replaced with something different" I mean, if coordinator can add different project address or even user address to get the payout. So target project won't get anything because registry has different addresses.

0xmad (approved):

Issue #1614, opened 2024-06-25
MACI audit v2.0.0


0xmad:
Support MACI audit for v2.0.0 version release

PR #1700, opened 2024-07-25
chore: optimize message processor and tally

0xmad:
# Description

- [x] Remove sha256 hashing
- [x] Support multiple params for verifier

## Additional Notes

N/A

## Related issue(s)

Related to #1614 

## Confirmation


circuits/circom/core/non-qv/processMessages.circom
```diff
@@ -48,17 +47,10 @@ include "../../trees/incrementalQuinaryTree.circom";
...
-
-    // The SHA-256 hash of values provided by the contract.
-    signal input inputHash;
-    signal input packedVals;
     // Number of users that have completed the sign up.
-    signal numSignUps;
+    signal input numSignUps;
     // Number of options for this poll.
-    signal maxVoteOptions;
+    signal input maxVoteOptions;
```
ctrlc03, line 53: given maxVoteOptions is always 5 ** voteOptionTreeDepth we can probably get rid of it completely as a input and just use the calculated one inside the circuit (MESSAGE_TREE_ARITY ** voteOptionTreeDepth)

contracts/contracts/MessageProcessor.sol
```diff
@@ -119,18 +119,9 @@ contract MessageProcessor is Ownable, SnarkCommon, Hasher, CommonUtilities, IMes
...
-        messageRoot,
-        sbCommitment,
-        _newSbCommitment,
-        messageTreeSubDepth,
-        messageTreeDepth,
-        voteOptionTreeDepth,
-        _proof
-      )
-    ) {
+    uint256[] memory publicCircuitInputs = getPublicCircuitInputs(currentMessageBatchIndex, _newSbCommitment);
```
ctrlc03, line 122: you should be able to pass uint8 messageTreeSubDepth, uint8 messageTreeDepth so we don't need to fetch twice, same for these values:  batchEndIndex, _currentMessageBatchIndex

ctrlc03: same for messageRoot

contracts/contracts/Tally.sol
```diff
@@ -154,59 +116,60 @@ contract Tally is Ownable, SnarkCommon, CommonUtilities, Hasher, DomainObjs {
...
-
     (uint256 numSignUps, ) = poll.numSignUpsAndMessages();
 
     // Require that there are untallied ballots left
     if (batchStartIndex >= numSignUps) {
       revert AllBallotsTallied();
     }
 
-    bool isValid = verifyTallyProof(_proof, numSignUps, batchStartIndex, tallyBatchSize, _newTallyCommitment);
+    uint256[] memory circuitPublicInputs = getPublicCircuitInputs(_newTallyCommitment);
```
ctrlc03, line 126: you already have numSignups, batchStartIndex, tallyBatchSize in this scope, can you pass the directly to verifyTallyProof?

contracts/contracts/Tally.sol
```diff
@@ -154,59 +116,60 @@ contract Tally is Ownable, SnarkCommon, CommonUtilities, Hasher, DomainObjs {
...
-  /// @param _tallyBatchSize batch size for the tally
-  /// @param _newTallyCommitment the tally commitment to be verified at a given batch index
   /// @return isValid whether the proof is valid
   function verifyTallyProof(
-    uint256[8] calldata _proof,
-    uint256 _numSignUps,
-    uint256 _batchStartIndex,
-    uint256 _tallyBatchSize,
-    uint256 _newTallyCommitment
+    uint256[] memory _publicInputs,
```
ctrlc03, line 163: given we know how many inputs, maybe we constraint the array to a fixed size?

contracts/contracts/crypto/Verifier.sol
```diff
@@ -26,15 +26,15 @@ contract Verifier is IVerifier, SnarkConstants, SnarkCommon {
...
+  /// @param inputs The public inputs to the circuit
   /// @return isValid Whether the proof is valid given the verifying key and public
   ///          input. Note that this function only supports one public input.
   ///          Refer to the Semaphore source code for a verifier that supports
   ///          multiple public inputs.
   function verify(
     uint256[8] memory _proof,
     VerifyingKey memory vk,
-    uint256 input
+    uint256[] memory inputs
```
ctrlc03, line 37: would calldata work here?

ctrlc03 (commented):
e2e tests work locally 🤔

contracts/contracts/MessageProcessor.sol
0xmad: The problem here is related to stack issue I mentioned before + `getPublicCircuitInputs` should be called without additional data passed as params. This is a public method which is also using for proof verification and public inputs check for prover and cli functions.

contracts/contracts/Tally.sol
0xmad: So, here is the same problem and idea as for MP.

contracts/contracts/Tally.sol
0xmad: I can try to check it, but my idea is: if we want to add more params, we don't need to change it later.

contracts/contracts/crypto/Verifier.sol
0xmad: Yes, it works

0xmad:
> e2e tests work locally 🤔

I think it's something related to github because there are different flaky tests across different jobs.

contracts/contracts/Tally.sol
0xmad: So, it won't work because verifier needs dynamic size.

kittybest (approved):
Not sure if I understood what do these changes improve, but the code itself looks good to me

ctrlc03 (approved):
Great work, thanks

PR #1714, opened 2024-07-30
docs(blog): add q3 roadmap post

ctrlc03:
# Description

Add Q3 roadmap post

## Confirmation

### DIFF
--- a/website/blog/2024-07-30-roadmap-q3.md
+++ b/website/blog/2024-07-30-roadmap-q3.md
@@ -1,0 +1,60 @@
+---
+slug: 2024-q3-roadmap
+title: 2024 Q3 Roadmap
+description: Our plans for MACI in Q3 2024
+authors:
+  name: ctrlc03
+  title: MACI team lead
+  url: https://x.com/ctrlc03
+  image_url: https://avatars.githubusercontent.com/u/93448202?v=4
+tags: [voting, security, anonymity, roadmap, grants]
+---
+
+# Q3 Roadmap
+
+Greetings anon,
+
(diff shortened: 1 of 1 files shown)


website/blog/2024-07-30-roadmap-q3.md
```diff
@@ -0,0 +1,60 @@
...
+- Support quadratic voting
+- Support the use of different signup mechanisms
+- Move application's data storage from Vercel to IPFS
+- Support contributing/voting to multiple rounds on the same deployment
+
+Our efforts will include supporting community organisers who want to fork and operate MACI Platform rounds in production. If you're interested in running a round for your community, [please get in touch](https://maci.pse.dev/docs/use-cases/public-goods-funding/quadratic-funding/apply)!
+
+## Coordinator Service
+
+After successfully implementing a proof generation service as part of our effort of reducing the burden on round organisers, we have plans to continue automating all MACI related operations. This includes contract deployment, proof generation and submission, contract details storage, automatic upload of final tally results to IPFS and more.
```
kittybest, line 36: just thinking about the `contract details storage` --> this service actually is only accessible for coordinators, so the deployed contracts details should be put on ipfs or some public place to let every user could fetch. So I'm not sure what content included in this `contract details storage`

ctrlc03: well actually it will not be accessible just to coordinators, only privileged functionality like proof gen require admin signatures. For instance, now the endpoint to get the public key is open to all. When we deploy the platform, the frontend will only know the coordinator service IP, an operator will then deploy the first instance of MACI and then we need a way for users to also know this address, but we don't want to go and add this address in the .env of the frontend, we just want to get it from the coordinator service. Whether the service stores it on IPFS or locally, or both, the data will come from it

kittybest (approved):
Thanks! lgtm

PR #1715, opened 2024-07-30
feat(contracts): added message processor and tally addresses to the maci contract

yashgo0018:
# Description
Changed the polls mapping to return the PollContracts struct instead of the poll address, it will now return the message processor address, tally contract address and poll contract address, when we will fetch the poll from the polls mapping and the getPoll function.

## Confirmation


ctrlc03 (approved):
Thank you!!

PR #1751, opened 2024-08-09
chore(contracts): deploy contracts to L2s, store addresses and benchmark message batch size

### LINKED ISSUE #1689: Add a page with supported networks and limitations
Currently, it is not clear on which live networks MACI is supported, and which networks are best for its usage. For instance on Scroll we can only send three messages in a batch, whereas on Optimism we can send 83-84. 

This should be a new entry in the documentation website

Requirements
- [x] deploy contracts on OP/Arbitrum/Base
- [x] Submit batches of messages to understand maximum values
- [x] Store reusable contract addresses on https://maci.pse.dev/docs/developers-references/smart-contracts/DeployedContracts

ctrlc03:
# Description

Deploy contracts to main L2 networks. Add a benchmark task for message batch costs and size. Store all new addresses for re-use.

## Related issue(s)

fix #1689 

## Confirmation

### DIFF
--- a/packages/contracts/tasks/helpers/benchmarks.ts
+++ b/packages/contracts/tasks/helpers/benchmarks.ts
@@ -1,0 +1,74 @@
+/* eslint-disable no-console */
+import { Message, Keypair } from "maci-domainobjs";
+
+import { MACI, Poll } from "../../typechain-types";
+
+import { Deployment } from "./Deployment";
+import { EContracts } from "./types";
+
+/**
+ * Utility function to send a batch of messages to a poll contract
+ * @dev This function will attempt to send a batch of messages to a poll contract
+ * and if it fails, it will attempt to send a smaller batch size
+ * @param deployment - The deployment instance
+ * @param message - The message to send
+ * @param keypair - The keypair to sign the message with
+ * @param batchSize - The number of messages to send
+ */
+export async function publishBatch(
+  deployment: Deployment,
+  message: Message,
+  keypair: Keypair,
+  batchSize: number,
+): Promise<void> {
+  // then try and publish a large batch of votes
+  const maciContract = await deployment.getContract<MACI>({
+    name: EContracts.MACI,
+  });
+  const lastPollId = (await maciContract.nextPollId()) - 1n;
+
+  const pollContract = await deployment.getContract<Poll>({
+    name: EContracts.Poll,
+    key: `poll-${lastPollId}`,
+  });
+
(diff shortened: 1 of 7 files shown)


packages/contracts/tasks/helpers/benchmarks.ts
```diff
@@ -0,0 +1,39 @@
...
+  } catch (err) {
+    console.log(`Failed to submit a batch of ${batchSize} messages\n`);
+    if (batchSize > 1) {
+      const newBatchSize = batchSize - 1;
+      console.log(`Trying with ${newBatchSize} messages\n`);
+      await publishBatch(pollContract, message, keypair, newBatchSize);
+    } else {
+      console.error("Unable to submit even a single message\n");
+    }
+  }
```
0xmad, line 38: It's better to avoid using catch for logic handling. Let's use static call to check the right batch size. Ideally, it can be done using binary search.

packages/contracts/tasks/runner/benchmarks.ts
```diff
@@ -0,0 +1,66 @@
...
+import { Keypair, PCommand } from "maci-domainobjs";
+
+import { publishBatch } from "../helpers/benchmarks";
+import { ContractStorage } from "../helpers/ContractStorage";
+import { Deployment } from "../helpers/Deployment";
+import { EContracts } from "../helpers/types";
+
+task("benchmark", "Run benchmarks").setAction(
+  // eslint-disable-next-line no-empty-pattern
+  async ({}, hre) => {
```
0xmad, line 12:
```suggestion
  async (_, hre) => {
```

0xmad (commented):
@ctrlc03 thanks, just some improvement suggestions

packages/contracts/tasks/helpers/benchmarks.ts
ctrlc03: good call, will do that :)

ctrlc03: seems like with static call it always gives Result(0) [] and I can't seem to find a way to determine whether it'd succeed or not. How have you used static call before?

```ts
    const res = await pollContract.publishMessageBatch.staticCallResult(testMessageBatch, testPubKeyBatch);
```

0xmad: It should work like this
```
await pollContract.publishMessageBatch.staticCall(...args).catch(() => undefined);
```

0xmad (approved):
@ctrlc03 thanks!

PR #1808, opened 2024-09-03
fix: do not cleanup thread all the time while doing parallel proof gen

ctrlc03:
# Description

There currently is a hanging issue with generating proofs within the hardhat tasks due to the different promises opening and closing handles to the same curve (reference: https://github.com/iden3/snarkjs/issues/152)

Also, do not extract Vk every time but only once when generating proofs for multiple batches

## Confirmation


kittybest (approved):
Thanks!

PR #1836, opened 2024-09-23
fix(zupass-gatekeeper): event id was stored at the wrong index

crisgarner:
# Description

The PCD libraries we use to generate the witness to send the data to the gatekeeper store the event IDs in an array on the 1 and 15 to 37 positions. At the contract level, we were checking the 15 positions, but unless the request is made in a certain way, the 15 value might be invalid. As we are only supporting one event ID for now, there is no need to check the other positions

## Additional Notes

Updated the tests also to reflect and tests these changes

## Confirmation


ctrlc03 (approved):
Thanks

PR #1961, opened 2024-12-09
feat: use gatekeeper per poll

### LINKED ISSUE #1858: Custom gatekeeper per poll
As a coordinator, I'd like to use custom gatekeeper per poll

ctrlc03:
# Description

Use a gatekeeper per each poll so that each can have its own gatekeeping (maci level more general, poll more specific)

## Related issue(s)

fix #1858 

## Confirmation

### DIFF
--- a/packages/cli/ts/commands/deployPoll.ts
+++ b/packages/cli/ts/commands/deployPoll.ts
@@ -1,4 +1,4 @@
-import { MACI__factory as MACIFactory, EMode } from "maci-contracts";
+import { MACI__factory as MACIFactory, EMode, deployFreeForAllSignUpGatekeeper } from "maci-contracts";
 import { PubKey } from "maci-domainobjs";
 
 import {
@@ -26,6 +26,7 @@ export const deployPoll = async ({
   coordinatorPubkey,
   maciAddress,
   vkRegistryAddress,
+  gatekeeperAddress,
   signer,
   quiet = true,
   useQuadraticVoting = false,
@@ -48,6 +49,18 @@ export const deployPoll = async ({
   }
 
   const maci = maciAddress || maciContractAddress;
+
+  const maciContract = MACIFactory.connect(maci, signer);
+  const pollId = await maciContract.nextPollId();
+
+  // check if we have a signupGatekeeper already deployed or passed as arg
+  let signupGatekeeperContractAddress =
+    gatekeeperAddress || (await readContractAddress(`SignUpGatekeeper-${pollId.toString()}`, network?.name));
+
+  if (!signupGatekeeperContractAddress) {
+    const contract = await deployFreeForAllSignUpGatekeeper(signer, true);
+    signupGatekeeperContractAddress = await contract.getAddress();
+  }
 
   // required arg -> poll duration
   if (pollDuration <= 0) {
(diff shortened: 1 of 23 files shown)


packages/contracts/tasks/deploy/poll/01-gatekeepers.ts
```diff
@@ -0,0 +1,293 @@
...
+      (semaphoreGatekeeperContractAddress || skipDeploySemaphoreGatekeeper) &&
+      (hatsGatekeeperContractAddress || skipDeployHatsGatekeeper) &&
+      (merkleProofGatekeeperContractAddress || skipDeployMerkleProofGatekeeper) &&
+      (!skipDeployFreeForAllGatekeeper ||
+        !skipDeployEASGatekeeper ||
+        !skipDeployGitcoinGatekeeper ||
+        !skipDeployZupassGatekeeper ||
+        !skipDeploySemaphoreGatekeeper ||
+        !skipDeployHatsGatekeeper ||
+        !skipDeployMerkleProofGatekeeper);
```
0xmad, line 86: Let's simplify it like this:

```
const canSkipDeploy =
  incremental &&
  ([
    freeForAllGatekeeperContractAddress,
    easGatekeeperContractAddress,
    gitcoinGatekeeperContractAddress,
    zupassGatekeeperContractAddress,
    semaphoreGatekeeperContractAddress,
    hatsGatekeeperContractAddress,
    merkleProofGatekeeperContractAddress
  ].some(Boolean)) &&
  [
    skipDeployFreeForAllGatekeeper,
    skipDeployEASGatekeeper,
    skipDeployGitcoinGatekeeper,
    skipDeployZupassGatekeeper,
    skipDeploySemaphoreGatekeeper,
    skipDeployHatsGatekeeper,
    skipDeployMerkleProofGatekeeper
  ].some(skip => !skip);
```

0xmad (commented):
@ctrlc03 thanks, just one suggestion.

packages/cli/ts/commands/deployPoll.ts
```diff
@@ -49,6 +50,18 @@ export const deployPoll = async ({
...
 
+  const maciContract = MACIFactory.connect(maci, signer);
+  const pollId = await maciContract.nextPollId();
+
+  // check if we have a signupGatekeeper already deployed or passed as arg
+  let signupGatekeeperContractAddress =
+    gatekeeperAddress || (await readContractAddress(`SignUpGatekeeper-${pollId.toString()}`, network?.name));
+
+  if (!signupGatekeeperContractAddress) {
+    const contract = await deployFreeForAllSignUpGatekeeper(signer, true);
```
kittybest, line 61: I don't quite understand this part: why we deploy `FreeForAllSignupGatekeeper` contract instead of letting them indicate what kind of gatekeeper they wanna deploy?

kittybest (commented):
Thanks! Just left a comment.
btw, I'm curious about if this PR would affect the user need to signup to each poll everytime? would it be possible that some polls wanna share the same signup information, only few of them wanna have their own gatekeeper? or is this because of the anon poll joining?
Just thinking about if one wanna attend 10 polls of a single event (let's say Devcon, for example), maybe the user don't want to sign up 10 times?

packages/cli/ts/commands/deployPoll.ts
ctrlc03: this is the same logic as in maci deployment in the cli/sdk. You are supposed to deploy it separately unless you just want the free for all

ctrlc03:
> only few of them wanna have their own gatekeeper

yes correct, in most cases this might be free for all 

> or is this because of the anon poll joining

Yes The poll joining is for the anon poll joining flow, so that you can join with a new anonymous key for each poll

0xmad (approved):
@ctrlc03 thanks!

PR #1967, opened 2024-12-12
feat: voice credits per poll

### LINKED ISSUE #1854: Custom voice credit proxy per poll
As a coordinator I'd like to use custom proxy per poll.

ctrlc03:
# Description

Add custom credits per each poll

## Related issue(s)

fix #1854 

## Confirmation


packages/crypto/ts/constants.ts
```diff
@@ -10,3 +10,5 @@ export const SNARK_FIELD_SIZE = r;
 export const NOTHING_UP_MY_SLEEVE = BigInt(keccak256(toUtf8Bytes("Maci"))) % SNARK_FIELD_SIZE;
 
 assert(NOTHING_UP_MY_SLEEVE === BigInt("8370432830353022751713833565135785980866757267633941821328460903436894336785"));
+
+export const PAD_KEY_HASH = BigInt("1309255631273308531193241901289907343161346846555918942743921933037802809814");
```
0xmad, line 14: Why do we use exactly this value?

0xmad (commented):
@ctrlc03 thanks, just some questions and minor suggestions

packages/crypto/ts/constants.ts
ctrlc03: That is the poseidon hash of the the PAD_KEY (https://github.com/privacy-scaling-explorations/maci/blob/dev/packages/contracts/contracts/crypto/SnarkConstants.sol#L12) - it's the same key that was part of the blank state leaf, and now because we only need to store the key at the maci state tree, we can directly use this value as init value for the tree

ctrlc03: noticed the last two lines of this comment are not accurate (https://github.com/privacy-scaling-explorations/maci/blob/dev/packages/contracts/contracts/crypto/SnarkConstants.sol#L18-L19), that hash starting 67 refers to the hash of the blank state leaf

PR #1974, opened 2024-12-20
feat: make nullifier not leak identity between polls

ctrlc03:
# Description

Currently, the nullifier leaks user identity between polls, as it's the poseidon hash of the user original private key. By adding the poll Id as a public input, we can check inside the circuit that nullifier = poseidon(privateKey, pollId), thus resulting in completely different nullifiers per each poll, and also keeping the use of the nullifier to prevent double signup.

## Confirmation

### DIFF
--- a/packages/circuits/circom/anon/pollJoining.circom
+++ b/packages/circuits/circom/anon/pollJoining.circom
@@ -41,8 +41,11 @@ template PollJoining(stateTreeDepth) {
     signal input stateRoot;
     // The actual tree depth (might be <= stateTreeDepth) Used in BinaryMerkleRoot
     signal input actualStateTreeDepth;
+    // The poll id
+    signal input pollId;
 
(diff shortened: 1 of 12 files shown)


kittybest (approved):
Thanks! But would the identity be leaked if we only hash the private key? I think the poseidon function is not reversible?

ctrlc03:
> Thanks! But would the identity be leaked if we only hash the private key? I think the poseidon function is not reversible?

The issue is that if you register to 5 polls, each poll will have the same nullifier (which is, well was, the hash of the original private key signed up to the maci contract), so even if you use different poll keys and different addresses, others might observe that one user is the same and has x keys on x polls. Adding the poll Id makes it a completely new value which does not leak you joining other polls

0xmad (approved):
@ctrlc03 thanks!

PR #1976, opened 2024-12-23
refactor: remove redundant init from poll contract

ctrlc03:
# Description

Poll init is redundant so it can be removed to simplify the code

(needs #1854 merged first and new anon poll joining artifacts built)

## Confirmation

### DIFF
--- a/packages/contracts/contracts/Poll.sol
+++ b/packages/contracts/contracts/Poll.sol
@@ -16,9 +16,6 @@ import { CurveBabyJubJub } from "./crypto/BabyJubJub.sol";
 /// @dev Do not deploy this directly. Use PollFactory.deploy() which performs some
 /// checks on the Poll constructor arguments.
 contract Poll is Params, Utilities, SnarkCommon, IPoll {
-  /// @notice Whether the Poll has been initialized
-  bool internal isInit;
-
   /// @notice The coordinator's public key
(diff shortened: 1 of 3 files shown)


packages/contracts/contracts/Poll.sol
```diff
@@ -186,33 +202,6 @@ contract Poll is Params, Utilities, SnarkCommon, IPoll {
     _;
   }
 
-  /// @notice The initialization function.
-  /// @dev Should be called immediately after Poll creation
-  function init() public virtual {
-    if (isInit) revert PollAlreadyInit();
-    // set to true so it cannot be called again
-    isInit = true;
```
0xmad, line 194: What's about maci-platform integration? There is a feature that poll can be created without initialization and after all the data ready (registries, applications) then poll can be initialized.

0xmad (commented):

packages/contracts/contracts/Poll.sol
ctrlc03: It would not be affected. The maci platform smart contracts already extend the poll factory to change the deploy function, so after this change they would need to add poll.init and implement their own init function inside the poll (or any other function they require).

0xmad (approved):
@ctrlc03 thanks!

PR #2002, opened 2025-01-14
feat(contracts): relay messages

0xmad:
# Description

- [x] Add relay messages method to Poll contract
- [x] Use struct for deploy poll args
- [x] Add relayer addresses to Poll contract

## Additional Notes

This is a part of happy offchain path

## Related issue(s)

Related to https://github.com/privacy-scaling-explorations/maci-platform/issues/279

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/packages/contracts/contracts/Poll.sol
+++ b/packages/contracts/contracts/Poll.sol
@@ -81,6 +81,12 @@ contract Poll is Params, Utilities, SnarkCommon, IPoll {
   /// @notice Poll state tree for anonymous joining
   LazyIMTData public pollStateTree;
 
+  /// @notice IPFS hashes of messages batches
+  bytes32[] public ipfsHashes;
+
+  /// @notice Relayer address
+  mapping(address => bool) public relayers;
+
   /// @notice The hash of a blank state leaf
   uint256 internal constant BLANK_STATE_LEAF_HASH =
     uint256(6769006970205099520508948723718471724660867171122235270773600567925038008762);
@@ -103,6 +109,7 @@ contract Poll is Params, Utilities, SnarkCommon, IPoll {
   error BatchHashesAlreadyPadded();
   error UserAlreadyJoined();
   error InvalidPollProof();
+  error NotRelayer();
 
   event PublishMessage(Message _message, PubKey _encPubKey);
   event MergeState(uint256 indexed _stateRoot, uint256 indexed _numSignups);
@@ -133,7 +140,8 @@ contract Poll is Params, Utilities, SnarkCommon, IPoll {
     PubKey memory _coordinatorPubKey,
     ExtContracts memory _extContracts,
     uint256 _emptyBallotRoot,
-    uint256 _pollId
+    uint256 _pollId,
+    address[] memory _relayers
   ) payable {
     // check that the coordinator public key is valid
     if (!CurveBabyJubJub.isOnCurve(_coordinatorPubKey.x, _coordinatorPubKey.y)) {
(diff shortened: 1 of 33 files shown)


packages/contracts/contracts/Poll.sol
```diff
@@ -228,6 +251,32 @@ contract Poll is Params, Utilities, SnarkCommon, IPoll {
     emit PublishMessage(_message, _encPubKey);
   }
 
+  /// @notice Allows relayer to publish messages using IPFS.
+  /// @param _messageHashes The message hashes
+  /// @param _ipfsHash The IPFS hash of the messages batch
+  function relayMessagesBatch(
+    uint256[] calldata _messageHashes,
```
ctrlc03, line 258: I thought we'd be relaying the full batch hash rather than each message hash, but this would work too, just a little bit more expensive for the relayer, but something to consider

Also if we go like this, it might be worth considering to constrain the number of hashes sent to be <= messageBatchSize, to keep things a bit more clean (only the last batch might be smaller)

ctrlc03 (commented):
Thanks, left some comments

packages/contracts/contracts/Poll.sol
0xmad: I use message hashes to keep `numMessages` sync with offchain messages.
And why do we need to check `messageBatchSize`? When we update chain hash we have this check when adding new chain hash:

```
if (numMessages % messageBatchSize == 0) {
  batchHashes.push(newChainHash);
}
```

I think it's the same logic for `publishMessageBatch`, there is no check for `messageBatchSize`.

ctrlc03: I'm thinking also in terms of how many messages can be at most in a ipfs hash, would be neat if each message was stored on ipfs, ok if we have fixed size batches stored on ipfs all in one, and not so great if we end up having different sized batches stored on ipfs, as that might make it very confusing when processing votes. was this something accounted for?

ctrlc03: so actually in this case messageBatchSize is what is used as param for the circuits, but maybe we end up using different values for the offchain batches? what do you think? would that make your thinking any different?

0xmad: > I'm thinking also in terms of how many messages can be at most in a ipfs hash, would be neat if each message was stored on ipfs, ok if we have fixed size batches stored on ipfs all in one, and not so great if we end up having different sized batches stored on ipfs, as that might make it very confusing when processing votes. was this something accounted for?

I think it doesn't matter because all the messages are handled without knowing about batch sizes.
In case if we want to support the same size for batches, we need to send only specific number of message hashes and it won't work for the last batch (can be smaller). But if there is an option to send smaller amount of hashes, it will be possible to send different batches.

Plus, if there are several relayers, some of them can use different size of messages in the batch (gas optimisations).

ctrlc03: > all the messages are handled without knowing about batch sizes
to which component of the system are you referring?

> and it won't work for the last batch
so my suggestion was to check <= batch size

0xmad: > to which component of the system are you referring?

This is implemented for `genMaciState` and subgraph event handing.

>

Right, but relayer can send `batch size - n` for all the other calls. There is no check for `publishMessageBatch` as well because if users need to publish a lot of messages, they need to use more gas for the transaction

ctrlc03 (approved):
Thanks

Issue #1981, opened 2025-01-02
Refactor SDK


ctrlc03:
Currently, SDK functions are deeply connected with the CLI (for example there is logging included). The goal is to separate the two so that the SDK is cleaner

PR #2070, opened 2025-01-27
refactor: verifying keys sdk

ctrlc03:
# Description

Refactor functions related to verifying keys

## Related issue(s)

re #1981 
## Confirmation

> [!IMPORTANT]

### DIFF
--- a/packages/cli/ts/commands/checkVerifyingKeys.ts
+++ b/packages/cli/ts/commands/checkVerifyingKeys.ts
@@ -1,13 +1,12 @@
-import { extractVk } from "maci-circuits";
-import { EMode, VkRegistry__factory as VkRegistryFactory } from "maci-contracts";
-import { VerifyingKey } from "maci-domainobjs";
+import { EMode } from "maci-contracts";
+import { extractAllVks, compareVks } from "maci-sdk";
+import { getAllOnChainVks } from "maci-sdk/build/ts/verifyingKeys";
 
 import fs from "fs";
 
 import {
   CheckVerifyingKeysArgs,
   banner,
-  compareVks,
   contractExists,
   info,
   logError,
@@ -52,8 +51,6 @@ export const checkVerifyingKeys = async ({
     logError("The VkRegistry contract does not exist");
   }
 
-  const vkRegistryContractInstance = VkRegistryFactory.connect(vkContractAddress, signer);
-
   // we need to ensure that the zkey files exist
   const isProcessMessagesZkeyPathExists = fs.existsSync(processMessagesZkeyPath);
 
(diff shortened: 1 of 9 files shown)


packages/sdk/ts/utils/interfaces.ts
```diff
@@ -469,3 +471,125 @@ export interface IIsJoinedUser {
...
+  /**
+   * The mode to use for the contract calls
+   */
+  mode: EMode;
+}
+
+/**
+ * MACI's verifying keys
+ */
+export interface MaciVerifyingKeys {
```
0xmad, line 512: Let's add `I` as for other interfaces (applicable for GetAllVksArgs too)

packages/sdk/ts/utils/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+declare global {
+  interface ITerminatable {
+    terminate: () => Promise<unknown>;
+  }
+
+  // eslint-disable-next-line vars-on-top, no-var, camelcase
+  var curve_bn128: ITerminatable | undefined;
+
+  // eslint-disable-next-line vars-on-top, no-var, camelcase
+  var curve_bls12381: ITerminatable | undefined;
```
0xmad, line 12: Is it possible to use `curve_bn128?: ITerminatable`?

packages/cli/ts/commands/checkVerifyingKeys.ts
```diff
@@ -68,38 +64,43 @@ export const checkVerifyingKeys = async ({
...
+      stateTreeDepth,
+      voteOptionTreeDepth,
+      messageBatchSize,
+      intStateTreeDepth,
+      mode,
+    });
 
     // do the actual validation
-    if (!compareVks(pollJoiningVk, pollJoiningVkOnChain)) {
+    if (!compareVks(pollJoiningVk!, pollJoiningVkOnChain)) {
```
0xmad, line 91: I think it makes sense to allow passing undefined value and check inside `compareVks` if value is defined.

0xmad (commented):
@ctrlc03 thanks, just left some minor comments

packages/sdk/ts/utils/utils.ts
ctrlc03: no that doesn't work

packages/cli/ts/commands/checkVerifyingKeys.ts
ctrlc03: but if it's not defined then what do we do? return true?

ctrlc03: maybe makes sense to actually check first in checkVerifyingKeys if (pollJoiningvk) { compareVk(poll...)}?

ctrlc03: either way the caller will need to check after if the result is false and not handle it if the vk is undefined, so why do that? 

I think it's better to keep as is, then when calling do:

if (vk) { compare(vk, vkOnChain) }

otherwise it'd need to be

const res = compare(vk, vkOnChain)
if (!res && !vk) { do nothing }

or am I missing something?

0xmad: If you pass undefined to compare it will fail and go to catch with ic is not exist in undefined.
And there is no case where you don't have key but it exists onchain.
So my idea is to update compare function so it looks like this:

```
export const compare = (vk?: Vk, onchainVk: Vk): boolean => {
  if (!vk) {
    throw new Error("No vk provided")
  }
}
```

This will show more readable error and the logic is the same.

0xmad (approved):
@ctrlc03 thanks!

PR #2089, opened 2025-01-29
refactor: vote sdk

ctrlc03:
# Description

Extract common code from publish cli command

## Related issue(s)

re #1981 
## Confirmation

> [!IMPORTANT]


apps/relayer/package.json
```diff
@@ -46,9 +46,9 @@
     "helmet": "^8.0.0",
     "lodash": "^4.17.21",
     "maci-cli": "workspace:^2.5.0",
-    "maci-contracts": "workspace:^2.5.0",
     "maci-sdk": "workspace:^0.0.1",
     "maci-domainobjs": "workspace:^2.5.0",
+    "maci-contracts": "workspace:^2.5.0",
```
0xmad, line 51: Do we still need it?

0xmad (approved):

apps/relayer/package.json
ctrlc03: looks like it's needed for the contract artifacts on hardhat config, not sure whether we can reference from sdk there, or add as dev dep only?

0xmad: Right, let's keep it in deps because we need artifacts to call message publishing.

PR #2167, opened 2025-02-20
docs(relayer): add general documentation for service

0xmad:
# Description

Add docs for relayer service

## Additional Notes

This is a part of happy offchain path

## Related issue(s)

Closes https://github.com/privacy-scaling-explorations/maci-platform/issues/279

## Confirmation

> [!IMPORTANT]


apps/relayer/README.md
```diff
@@ -1,5 +1,146 @@
...
+![MACI Relayer Flow](./relayer-diagram.png)
+
+## High-Level Overview of the Relayer:
+
+### Functionality
+
+The Relayer serves as an intermediary between users and the MACI system. It performs the following tasks:
+
+- Allows users to send messages without having to pay gas fees by processing these messages off-chain.
+- Receives user messages, such as votes or sign-ups.
```
ctrlc03, line 18: let's remove sign-ups for now as we don't support them (yet)?

apps/relayer/README.md
```diff
@@ -1,5 +1,146 @@
...
+- Prepare messages batch and upload it to IPFS
```
ctrlc03, line 19: batches in plural?

apps/relayer/README.md
```diff
@@ -1,5 +1,146 @@
...
+- Submits the message hashes to the Poll contract with IPFS hash of the batch.
+
+This process ensures that user actions are recorded on-chain without compromising their privacy.
+
+### Privacy Preservation
+
+By handling the creation and submission of zero-knowledge proofs, the Relayer ensures that sensitive user information remains confidential. This design choice is essential for maintaining the integrity and trustworthiness of the voting process.
```
ctrlc03, line 26: don't think the services handles creating any proofs, users generate their own proofs client side to prove they joined a poll

apps/relayer/README.md
```diff
@@ -1,5 +1,146 @@
...
+
+### Decentralization and Trust
+
+The Relayer can be self-hosted, which enhances privacy guarantees and reduces reliance on a single trusted party. This decentralization aligns with the broader goals of the MACI framework to minimize trust assumptions and potential points of failure.
```
ctrlc03, line 30: I'm not sure this is true,  the service is run by the coordinator/admin so not really decentralised - and currently allowing others to run it and add them as relayers on the system would open up to malicious activity as they could publish invalid batch hashes and dos the round

apps/relayer/README.md
```diff
@@ -1,5 +1,146 @@
...
+## Getting Started
+
+Follow these steps to run the MACI Relayer service using Docker.
+
+### 1. Clone the Repository
+
+First, clone the repository to your local machine.
+
+```bash
+git clone https://github.com/your-organization/your-repo.git
```
ctrlc03, line 55: should we add the MACI repo url?

apps/relayer/README.md
```diff
@@ -1,5 +1,146 @@
...
+
+This will stop and remove the containers but retain the volumes (like MongoDB data). To also remove the volumes, run:
+
+```bash
+docker-compose down --volumes
+```
+
+## 7. MongoDB Persistence
+
+MongoDB data is persisted using Docker volumes (`mongodb-data`). Even if you remove and recreate the MongoDB container, the data will persist.
```
ctrlc03, line 130: nice

ctrlc03 (commented):
Thanks, left some comments

apps/relayer/README.md
0xmad: I think it's true for the party that all the users trust. The same as for coordinator but it can be a different 3rd party.

ctrlc03: but that does not minimise points of failure, it would actually increase them

0xmad: I added a comment for this block that relayer must be trusted by users and coordinator.

apps/relayer/README.md
```diff
@@ -1,5 +1,146 @@
...
+- Allows users to send messages without having to pay gas fees by processing these messages off-chain.
+- Receives user messages.
+- Prepare messages batches and upload it to IPFS
+- Submits the message hashes to the Poll contract with IPFS hash of the batch.
+
+This process ensures that user actions are recorded on-chain without compromising their privacy.
+
+### Privacy Preservation
+
+By handling the zero-knowledge proofs, the Relayer ensures that only authorized users can send messages. This design choice is essential for maintaining the integrity and trustworthiness of the voting process.
```
ctrlc03, line 26: this is very vague, which zk proofs? also verifying would be more appropriate. Let's explain that users authenticate by providing a proof that they know the pre-image to a state leaf in the poll state tree

apps/relayer/README.md
ctrlc03: I still don't think that's a true statement unfortunately.
Also: "Anyway, relayer must be trusted by coordinator and users due to current status of MACI design." it's very vague and doesn't really give an explanation 

My suggestion is to remove this part, and mention that it's recommended that the coordinator only runs this service for now. That is because if the service malfunctions and batch hashes are pushed on chain without the messages actually being stored on ipfs, then rounds are automatically botched

ctrlc03 (approved):
Thanks

PR #2193, opened 2025-02-28
Return processProofs, tallyProofs and tallyData in generateProofs function

NicoSerranoP:
# Description

The previous `generateProofs` function was returning only `tallyData` but `processProofs` and `tallyProofs` were missing. The coordinator service was built using these two missing variables so I needed to return them in the sdk function. By using an interface, we are not modifying the function usage to much (instead of `const tallyData = await generateProofs( ... )` we use `const { tallyData } = await generateProofs( ... )`

## Confirmation

> [!IMPORTANT]


ctrlc03 (approved):
Thanks

Issue #2141, opened 2025-02-14
Refactor test suite


ctrlc03:
Refactor the test suite as now it's separated between two packages (cli and integrationTests) and has a lot of redundant test scenarios, with several more important missing

PR #2229, opened 2025-03-06
refactor: make separate testing package

ctrlc03:
# Description

Create separate package for testing (maci-testing) and move over test suite. 

## Related issue(s)

related to #2141 
## Confirmation

> [!IMPORTANT]


packages/testing/ts/constants.ts
```diff
@@ -0,0 +1,242 @@
+import { Keypair } from "maci-domainobjs";
+import {
+  EMode,
+  extractAllVks,
+  generateTallyCommitments,
+  getPollParams,
+  ICheckVerifyingKeysArgs,
+  IDeployPollArgs,
+  ISetVerifyingKeysArgs,
```
0xmad, line 9:
```suggestion
  type ICheckVerifyingKeysArgs,
  type IDeployPollArgs,
  type ISetVerifyingKeysArgs,
```

packages/testing/ts/types.ts
```diff
@@ -0,0 +1,136 @@
...
+  messages: IIpfsMessage[];
+  maciAddress: string;
+  signer: Signer;
+  pollId: number;
+}
+
+/**
+ * A class representing a user and its votes
+ */
+export class User {
```
0xmad, line 86: Let's move this to a separate file

0xmad (commented):
@ctrlc03 thanks, let's remove `.skip` and `.only` for tests and some minor code style comments.

packages/testing/ts/types.ts
ctrlc03: if I move that to a separate file, then I will need to move out something else to another file to avoid an import cycle - either move IVote too or move IContractsData - what you think it's best?

0xmad: So to prevent dependency cycle, you can use `import type`. All the types can have cycles and in the result bundle (js) types are not available. So it's not a problem if you're using `import type` for `IVote` or `IContractsData`.

ctrlc03: oh nice thanks

0xmad (approved):
@ctrlc03 thanks!

PR #2275, opened 2025-03-24
refactor: contracts optimisations and cleanup

ctrlc03:
# Description

* remove ownable from mp and tally (can only post valid proofs)
* remove _votingPeriodOver() from processMessages() and tallyVotes() -> those functions can only be called after the voting period is over because:
  * processMessages calls poll.padLastBatchHash() the first time which checks poll end time is passed
  * tallyVotes checks that processMessages was called for all batches (thus can only be after poll end time)
* use memory for MACI.deployPoll arg, as when integrating one might be passing arguments stored and constructing the struct object locally, thus not able to set it as calldata. furthermore, extending contracts might be modifying the passed struct 
* MACI.getStateIndex throws for invalid keys

## Confirmation

> [!IMPORTANT]


0xmad (approved):
@ctrlc03 thanks!

PR #2346, opened 2025-04-07
fix: ensure browser sdk does not import maci-contracts

ctrlc03:
# Description

Ensure we only export functions which do not use @maci-protocol/contracts (only types) as hardhat is not supported on browsers

## Confirmation

> [!IMPORTANT]


packages/sdk/ts/browser/index.ts
```diff
@@ -1,28 +1,13 @@
...
-  extractVk,
-  genProofSnarkjs,
-  formatProofForVerifierContract,
-  verifyProof,
-  cleanThreads,
-  unlinkFile,
-  getBlockTimestamp,
-  genEmptyBallotRoots,
-} from "@maci-protocol/contracts";
+export { getSignedupUserData, signup, getJoinedUserData, hasUserJoinedPoll, hasUserSignedUp } from "../user";
```
0xmad, line 8: Let's use `export * from "../user";` as for other exports

0xmad (commented):
@ctrlc03 thanks, just one comment!

packages/sdk/ts/browser/index.ts
ctrlc03: we don't want to export joinPoll from `../user` because it relies on maci-contracts functions

ctrlc03: I made a separate version of `joinPoll` that uses wasm specifically and functions implemented from within maci-sdk directly (inside ts/browser/)

0xmad (approved):
@ctrlc03 thanks!

PR #2368, opened 2025-04-14
feat: add poll state tree depth

### LINKED ISSUE #2272: Specify stateTreeDepth per poll
Instead of relying to MACI.stateTreeDepth, let's specify a stateTreeDepth for each poll so that trees can be smaller (if needed)

0xmad:
# Description

- [x] Use poll state tree depth instead of maci state tree depth
- [x] Use maci state tree depth as a fallback
- [x] Update params for setting keys for vk registry
- [x] Update deploy scripts and compatibility fixes
- [x] Set verifying keys for poll if signature is not matched

## Additional Notes

Requires to compile circuits with `pollStateTreeDepth` params.

## Related issue(s)

Closes #2272

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/packages/contracts/contracts/Poll.sol
+++ b/packages/contracts/contracts/Poll.sol
@@ -17,6 +17,12 @@ import { CurveBabyJubJub } from "./crypto/BabyJubJub.sol";
 /// @dev Do not deploy this directly. Use PollFactory.deploy() which performs some
 /// checks on the Poll constructor arguments.
 contract Poll is Clone, Params, Utilities, SnarkCommon, IPoll {
+  /// @notice The max number of poll signups
+  uint256 public maxSignups;
+
+  /// @notice The state tree arity
+  uint8 internal constant STATE_TREE_ARITY = 2;
+
   /// @notice The coordinator's public key
   PubKey public coordinatorPubKey;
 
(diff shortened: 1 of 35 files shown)


packages/contracts/contracts/Poll.sol
```diff
@@ -181,6 +185,9 @@ contract Poll is Clone, Params, Utilities, SnarkCommon, IPoll {
     // store message batch size
     messageBatchSize = _messageBatchSize;
 
+    // store  state tree depth
+    stateTreeDepth = _stateTreeDepth;
+
     // store tree depth
     treeDepths = _treeDepths;
```
ctrlc03, line 192: why not store it inside treeDepths (seems like the best place logically to keep it)?

packages/contracts/contracts/Poll.sol
```diff
@@ -181,6 +185,9 @@ contract Poll is Clone, Params, Utilities, SnarkCommon, IPoll {
     // store message batch size
     messageBatchSize = _messageBatchSize;
 
+    // store  state tree depth
+    stateTreeDepth = _stateTreeDepth;
```
ctrlc03, line 189: now that we can allow for a different state depth than the maci tree, we are going to need to ensure that when a user calls `joinPoll` the tree is not already full (before it would not be needed as the capacity check was done on the MACI contract already)

ctrlc03 (changes requested):
Thanks, left some comments

ctrlc03 (approved):
Thank you

PR #2483, opened 2025-05-14
fix(sdk): remove stateIndex from joinPoll + verify in ETH sepolia

NicoSerranoP:
# Description

I am removing the `stateIndex` function input from `joinPoll` because it is computed inside the sdk function. Thanks to the latest version of `joinPoll` there was no need to send the `stateIndex` and in my opinion it is better because it could cause confusion (e.g. should we use the state index of the maci contract or of the poll contract?)

## Confirmation

> [!IMPORTANT]


packages/testing/ts/__tests__/e2e.nonQv.test.ts
```diff
@@ -701,12 +697,14 @@ describe("e2e tests with non quadratic voting", function test() {
     });
 
     it("should join two users", async () => {
-      for (let index = 0; index < users.length; index += 1) {
+      // we cannot use Promise.all( users.map( ... ) ) because they will run concurrently
```
NicoSerranoP, line 700: es-lint was giving an error asking me to change from
```javascript
for (let index = 0; index < users.length; index += 1) {
``` 

to

```javascript
for (const user of users) {
```

and later on even to
```javascript
Promise.all( users.map( ... ) )
```


The problem with `users.map( ... )` is that it runs concurrently and not sequentially. This could cause non-deterministic errors between function executions. Remember that each user can join the poll but they need to do it sequentially.


I decided that the best way to solve this was to disable es-lint for this line and add a comment explaining why

ctrlc03: can't we keep it as  for (let index = 0; index < users.length; index += 1) as that's what we usually use?

0xmad, line 700: No need to add this comment

packages/testing/ts/__tests__/e2e.test.ts
```diff
@@ -1476,12 +1466,14 @@ describe("e2e tests", function test() {
     it("join the second and third polls", async () => {
       // eslint-disable-next-line @typescript-eslint/prefer-for-of
       for (let pollId = 1; pollId <= 2; pollId += 1) {
-        for (let i = 0; i < users.length; i += 1) {
+        // we cannot use Promise.all( users.map( ... ) ) because they will run concurrently
+        // and we need to wait for the transaction to be mined before joining the next user
```
0xmad, line 1470: Same here, just remove the comment

packages/testing/ts/__tests__/e2e.test.ts
```diff
@@ -1476,12 +1466,14 @@ describe("e2e tests", function test() {
     it("join the second and third polls", async () => {
       // eslint-disable-next-line @typescript-eslint/prefer-for-of
       for (let pollId = 1; pollId <= 2; pollId += 1) {
-        for (let i = 0; i < users.length; i += 1) {
```
0xmad, line 1479: You can keep this

0xmad (commented):
@NicoSerranoP thanks, just some comments

packages/testing/ts/__tests__/e2e.nonQv.test.ts
NicoSerranoP: I really don't mind but the es-lint throws an error, to silence it we have 2 options:

```javascript
// eslint-disable-next-line @typescript-eslint/prefer-for-of
for (let i= 0; i< users.length; i+= 1) {
```

```javascript
// eslint-disable-next-line no-restricted-syntax
for (const user of users) {
```

Let me know which one you prefer 😊 @ctrlc03 @0xmad

packages/testing/ts/__tests__/e2e.test.ts
NicoSerranoP: image

I am getting the es-lint error :S I can see that we are using `// eslint-disable-next-line @typescript-eslint/prefer-for-of` on the for above, I could use that one as well but please check Ale's comment to recommend me which es-lint silencer we use

0xmad: We have this code working in other tests with for loop and eslint ignore comment (`eslint-disable-next-line @typescript-eslint/prefer-for-of`)

0xmad (approved):
@NicoSerranoP thanks!

Issue #2369, opened 2025-04-15
feat (coordinator service): automate poll tallying start


ctrlc03:
Add a cronjob to the coordinator service to automate the start of poll finalisation (processing and tallying)

NicoSerranoP:
I understand there are two issues:

1. **Storing poll state (endDate, isTallied, etc):** I think that using `lowdb` or `SQLite3` might reduce the over engineering issue (I completed agree with it, we should keep it as simple as possible). @0xmad do you have any preference on these ones? Do you think redis would be better than `lowdb` or `SQLite3` in this case?

2. **Cron jobs management:** My idea was to have one cron job that runs every N seconds and checks if any poll needs to be deployed. Therefore there will be only one Cron job that would read into the database every N seconds. NestJS allows you to set up cron jobs in the service logic.

ctrlc03:
I think we also need some fallback in case a registration of a cron is not successful (for whatever reason that might be)

> 1. `lowdb` is just an abstraction to read files, it's not a database. `sqlite` is database but it will require adding the additional logic like repositories for the entities. `redis` will have the same logic changes but it's a more lightweight because it's just a key-value storage (which is perfect for caching).
> 2. Why do you need to run it every `N` seconds? It should be registered to reduce resources using for it.

NicoSerranoP:
1. Ok! So I will go with `redis` then 😊 

2.1. My initial idea was to use [declarative cron jobs](https://docs.nestjs.com/techniques/task-scheduling#declarative-cron-jobs) so it would be easy to manage (not many different cron jobs created but only one that would check if there are pending polls. It would check ALL polls in the maci contract).

2.2. We could use [dynamic cron jobs](https://docs.nestjs.com/techniques/task-scheduling#dynamic-cron-jobs) to register one every time a new poll is deployed. Using this solution we need to consider the case of a poll being deployed outside the coordinator service (e.g. in the Aragon plugin). How would we register that? A end user should register that?


In the Aragon - MACI case we could use 2.1. and the end user would never need to interact with the coordinator service at all (it could be safely deployed in a private network or something)

0xmad:
If the coordinator service is going to be used for poll finalization, it should be registered explicitly for it. It's more transparent and you don't need to parse event every time. Anyway, you need to register maci contract for it because it's not a generic service per every maci instances.
[...]

PR #2509, opened 2025-05-30
Coordinator: changes to use it with Aragon UI

NicoSerranoP:
# Description

## Confirmation

> [!IMPORTANT]


apps/coordinator/.env.example
```diff
@@ -7,6 +7,8 @@ LIMIT=10
 COORDINATOR_PUBLIC_KEY_PATH="./public.key"
 COORDINATOR_PRIVATE_KEY_PATH="./private.key"
 
+COORDINATOR_MACI_PRIVATE_KEY="macisk.a2c1dbb1ab568fdb1a34d3b3db5dbe7544e7f46bc0429ada2ac2c38c2ed956af"
```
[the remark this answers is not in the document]
NicoSerranoP: We are using this test private key across the Aragon UI, Maci Voting plugin and the coordinator. Could we leave it just as a default so it is clear which key is being used in the whole flow?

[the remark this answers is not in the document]
0xmad: There should be a different place (like password vault) to keep this secret. 
`.env.example` shouldn't have this.

apps/coordinator/ts/file/file.service.ts
```diff
@@ -100,15 +100,16 @@ export class FileService {
...
     const root = path.resolve(process.env.COORDINATOR_ZKEY_PATH!);
     const index = name.indexOf("_");
     const type = name.slice(0, index);
     const params = name.slice(index + 1);
     const modePrefixes = {
-      [EMode.QV]: "",
+      [EMode.QV]: "QV",
       [EMode.NON_QV]: "NonQv",
       [EMode.FULL]: "Full",
+      noMode: "",
```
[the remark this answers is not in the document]
NicoSerranoP: That is a great idea 🥳 . I will implement it like that.

apps/coordinator/.env.example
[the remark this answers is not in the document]
NicoSerranoP: Got it! 🚀

apps/coordinator/ts/proof/types.ts
```diff
@@ -65,7 +65,7 @@ export interface IGenerateArgs {
   /**
    * Encrypted coordinator private key with RSA public key (see .env.example)
    */
-  encryptedCoordinatorPrivateKey: string;
+  encryptedCoordinatorPrivateKey?: string;
```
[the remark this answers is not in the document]
NicoSerranoP: We considered scenario 2 before but we noticed that it would be easy to have scenario 1 with the Aragon UI.

All DAO members (address with enough governance tokens) can create proposals. Each proposal deploys a poll and the voting period starts. After the voting has ended, any user should be able to finalize the poll by clicking a "Finalize" button.


At the same time this feature would help with the schedule poll finalization https://github.com/privacy-scaling-explorations/maci/issues/2369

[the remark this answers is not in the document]
JohnGuilding: The rationale was we wanted an intermediate solution that would allow us to get the demo working quickly, without proposing larger architectural changes to the coordinator service. In this sense, allowing both scenarios was both backward and forward compatible.

That said, if there is buy-in for this change (adding coordinator private key to the coordinator service env), then we don't necessarily need to have both scenarios accounted for at the same time. I agree with your sentiment that supporting one scenario is better than supporting both. I view us as having three paths forward:
1. Keep it as optional and make it not optional in a subsequent PR
2. Make it not optional in this PR
3. If you disagree with this approach, consider reverting and using authentication mechanism as envisioned

**Rationale for storing private key in coordinator service**

The rationale for storing the coordinator private key in the coordinator service is that otherwise, we would need to enter the private key in the Aragon frontend before encrypting it and using it for authentication. We felt like this would not make for a good demo when we show Aragon, and the security implications of handling a private key with a frontend were worse than storing a private key in the backend. 

We of course did not design the coordinator service and are not super familiar with the design decisions and rationale. We know that it was an explicit choice to NOT store the private key in the coordinator service, so keen to hear previous rationale if you think this new approach is wrong.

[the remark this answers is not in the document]
0xmad: So, why do you need to support scenario 2? It makes sense to remove encrypted private key param and keep it only as env.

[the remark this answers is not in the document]
0xmad: > The rationale was we wanted an intermediate solution that would allow us to get the demo working quickly, without proposing larger architectural changes to the coordinator service. In this sense, allowing both scenarios was both backward and forward compatible.
> 
> That said, if there is buy-in for this change (adding coordinator private key to the coordinator service env), then we don't necessarily need to have both scenarios accounted for at the same time. I agree with your sentiment that supporting one scenario is better than supporting both. I view us as having three paths forward:
> 
>     1. Keep it as optional and make it not optional in a subsequent PR
> 
>     2. Make it not optional in this PR
> 
>     3. If you disagree with this approach, consider reverting and using authentication mechanism as envisioned
> 
> 
> **Rationale for storing private key in coordinator service**
> 
> The rationale for storing the coordinator private key in the coordinator service is that otherwise, we would need to enter the private key in the Aragon frontend before encrypting it and using it for authentication. We felt like this would not make for a good demo when we show Aragon, and the security implications of handling a private key with a frontend were worse than storing a private key in the backend.
> 
> We of course did not design the coordinator service and are not super familiar with the design decisions and rationale. We know that it was an explicit choice to NOT store the private key in the coordinator service, so keen to hear previous rationale if you think this new approach is wrong.

I think there are no users of coordinator service who uses latest changes so it makes sense to remove this private key param from dto and use private key from `.env` file. It's not a big change and doesn't break Aragon UI logic.

apps/coordinator/ts/deployer/types.ts
```diff
@@ -289,6 +294,7 @@ export interface IDeployMaciConfig {
   MACI: {
     stateTreeDepth: number;
     policy: EPolicies;
+    mode?: EMode;
```
0xmad, line 297: Why is it optional?

apps/coordinator/ts/deployer/deployer.service.ts
```diff
@@ -290,12 +290,12 @@ export class DeployerService {
   ): Promise<ISetVerifyingKeysArgs> {
     const { zkey: pollJoiningZkeyPath } = this.fileService.getZkeyFilePaths(
       process.env.COORDINATOR_POLL_JOINING_ZKEY_NAME!,
-      EMode.QV,
+      "noMode",
```
0xmad, line 293: This can be removed and for pollJoined, too

apps/coordinator/hardhat.config.cjs
```diff
@@ -13,7 +13,7 @@ module.exports = {
   defaultNetwork: "sepolia",
   networks: {
     sepolia: {
-      url: process.env.COORDINATOR_RPC_URL || "http://localhost:8545",
+      url: `https://opt-sepolia.g.alchemy.com/v2/${process.env.RPC_API_KEY}`,
```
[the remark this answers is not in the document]
0xmad: Sorry, but no. It was implemented before and removed with current change.

apps/coordinator/ts/common/chain.ts
```diff
@@ -4,23 +4,23 @@ import { ErrorCodes } from "./errors";
...
-  if (!apiKey) {
-    throw new Error(ErrorCodes.RPC_API_KEY_NOT_SET.toString());
+  if (!COORDINATOR_RPC_URL) {
+    throw new Error(ErrorCodes.COORDINATOR_RPC_URL_NOT_SET.toString());
   }
 
   switch (network) {
     case ESupportedNetworks.OPTIMISM_SEPOLIA:
-      return `https://opt-sepolia.g.alchemy.com/v2/${apiKey}`;
+      return COORDINATOR_RPC_URL;
```
ctrlc03, line 21: we don't need the switch here, it's the same env var for all

apps/coordinator/ts/common/coordinatorKeypair.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+export const getCoordinatorKeypair = (): Keypair => {
+  const privateKey = String(process.env.COORDINATOR_MACI_PRIVATE_KEY);
+
+  if (!privateKey) {
+    throw new Error("COORDINATOR_MACI_PRIVATE_KEY environment variable is not set.");
+  }
+
+  const maciPrivateKey = PrivateKey.deserialize(privateKey);
+  return new Keypair(maciPrivateKey);
```
0xmad, line 15:
```suggestion
  const maciPrivateKey = PrivateKey.deserialize(privateKey);

  return new Keypair(maciPrivateKey);
```

apps/coordinator/ts/common/chain.ts
```diff
@@ -4,23 +4,23 @@ import { ErrorCodes } from "./errors";
...
+ * Get the RPC url for the chain we need to interact with
  *
  * @param network - the network we want to interact with
- * @returns the RPCUrl for the network
+ * @returns the RPC url for the network
  */
-export const getAlchemyRpcUrl = (network: ESupportedNetworks): string => {
-  const apiKey = process.env.RPC_API_KEY;
+export const getRpcUrl = (network: ESupportedNetworks): string => {
+  const { COORDINATOR_RPC_URL } = process.env;
```
0xmad, line 13: You should use it like this because sometimes env values are empty when you use destructuring.

```suggestion
  const rpcUrl  = process.env.COORDINATOR_RPC_URL;
```

apps/coordinator/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+USER node
+
+# monorepo root directory
+WORKDIR /maci
+COPY --chown=node:node  --from=builder /builder/ ./
+
+RUN pnpm run download-zkeys:test:coordinator
+
+RUN mkdir -p /home/node/rapidsnark/build; \
```
0xmad, line 31: I think it should be like for the previous version:

```
RUN mkdir -p ~/rapidsnark/build; \
    wget -qO ~/rapidsnark/build/prover https://maci-devops-zkeys.s3.ap-northeast-2.amazonaws.com/rapidsnark-linux-amd64-1c137; \
    chmod +x ~/rapidsnark/build/prover
RUN wget -qO ~/circom https://github.com/iden3/circom/releases/download/v2.1.6/circom-linux-amd64; \
    chmod +x ~/circom; \
    mv ~/circom /bin
```

apps/coordinator/tests/e2e.aa.test.ts
```diff
@@ -62,7 +60,7 @@ describe("E2E Account Abstraction Tests", () => {
   describe("merge", () => {
     test("should return true when there are no errors", async () => {
       // wait until we can complete a poll
-      await sleep(40000);
+      await sleep(90 * 1000); // 50 start + 35 end + 5 seconds of grace time
```
0xmad, line 63: Why did you change it?

apps/coordinator/tests/e2e.deploy.test.ts
```diff
@@ -246,22 +247,33 @@ describe("E2E Deployment Tests", () => {
...
+        tag: `v0.0.${maciCreatedAt}`, // different versions per test using block number
       } as IDeploySubgraphArgs),
     });
-    const body = (await response.json()) as { url: string };
-    const { url } = body;
+
+    const body = await response.json();
+
+    if (response.status === 400) {
+      const errorMessage = (body as { message: string }).message;
```
0xmad, line 260: Empty line after

apps/coordinator/package.json
```diff
@@ -24,7 +24,7 @@
     "start:prod": "pnpm run run:node build/ts/main.js",
     "test": "jest --testPathIgnorePatterns=e2e.deploy.test.ts --testPathIgnorePatterns=e2e.test.ts --forceExit",
     "test:coverage": "jest --coverage --forceExit",
-    "test:e2e": "jest ./tests/*.test.ts --forceExit",
+    "test:e2e": "jest ./tests/*.test.ts --runInBand",
```
0xmad, line 27: Why did you change it?

NicoSerranoP: Because the e2e.aa.test.ts were running in parallel and we  needed the `--runInBand` flag. I forgot to readd the `--forceExit` one, I just pushed it with the new changes

0xmad (approved):
@NicoSerranoP thanks!

PR #2556, opened 2025-06-12
feat(contracts): enhance ITally interface as per Tally.sol

lordshashank:
# Description

- Enhances the `Itally.sol` to include most of the functions from `Tally.sol` which can then be used by developers for various purposes  like to check if poll `isTallied()`, etc.
- Makes the `tallyResults` mapping internal and creates getter functions to access them. With public, compiler was throwing error for default getter function ([issue](https://ethereum.stackexchange.com/questions/125241/how-to-override-an-interface-function-that-returns-a-struct-with-a-mapping))
- Replaces `tallyResult` with `getTallyResult` in scripts and tests.
- Deletes the `MockTally.sol` as was not needed.

## Additional Notes

Been working on privote.live, would help me there.

## Confirmation

> [!IMPORTANT]


packages/contracts/contracts/Tally.sol
```diff
@@ -113,6 +105,23 @@ contract Tally is Clone, SnarkCommon, Hasher, DomainObjs, ITally {
...
+
+  /// @notice Get complete tally results
+  /// @return The tally results
+  function getCompleteTallyResults() public view returns (TallyResult[] memory) {
+    TallyResult[] memory results = new TallyResult[](totalTallyResults);
+    for (uint256 i = 0; i < totalTallyResults; i++) {
+      results[i] = tallyResults[i];
+    }
+    return results;
+  }
```
0xmad, line 123: Copying all the tally results into a memory is not a good solution.
If you need results for specific options, you can use `getTallyResults` with specified indices or you can use subgraph.

packages/contracts/contracts/Tally.sol
```diff
@@ -113,6 +105,23 @@ contract Tally is Clone, SnarkCommon, Hasher, DomainObjs, ITally {
     }
   }
 
+  /// @notice Get the tally results for a specific index
+  /// @param index The index to query
+  /// @return The tally result
```
0xmad, line 110: If the code is moved to the interface, you should inherit docs from it to prevent duplication.

0xmad (changes requested):
@lordshashank thanks, just changes requested.

0xmad (approved):
@lordshashank thanks!

PR #2606, opened 2025-07-03
feat: dockerize coordinator

NicoSerranoP:
# Description

1. Dockerize coordinator service so you can easily run it using the README.md instructions
2. Added a health module to check if rapidsnark is available, zkeys are downloaded and the coordinator wallet has funds

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/apps/coordinator/ts/health/health.controller.ts
+++ b/apps/coordinator/ts/health/health.controller.ts
@@ -1,0 +1,25 @@
+import { Controller, Get, HttpStatus } from "@nestjs/common";
+import { ApiResponse, ApiTags } from "@nestjs/swagger";
+
+import { HealthService } from "./health.service";
+import { IHealthCheckResponse } from "./types";
+
+@ApiTags("v1/health")
+@Controller("v1/health")
+export class HealthController {
+  /**
+   * Initialize HealthController
+   */
+  constructor(private readonly healthService: HealthService) {}
+
+  /**
+   * Health check api method.
+   *
+   * @returns true if the service is running
+   */
+  @ApiResponse({ status: HttpStatus.OK, description: "The service is running" })
+  @Get("check")
+  check(): Promise<IHealthCheckResponse> {
+    return this.healthService.checkCoordinatorHealth();
+  }
+}
--- a/apps/coordinator/ts/health/health.service.ts
+++ b/apps/coordinator/ts/health/health.service.ts
@@ -1,0 +1,175 @@
+import { EMode } from "@maci-protocol/sdk";
+import { Injectable } from "@nestjs/common";
+import { formatEther } from "ethers";
+import { zeroAddress } from "viem";
+
+import fs from "fs";
+import path from "path";
+
+import type { ICheckRapidsnark, ICheckWalletFunds, ICheckZkeysDirectory, IHealthCheckResponse } from "./types";
+
+import { ESupportedNetworks, getSigner } from "../common";
+import { FileService } from "../file/file.service";
+
+/**
+ * Health service functions to check the status of the coordinator service.
+ */
+@Injectable()
+export class HealthService {
+  /**
+   * Create a new instance of the HealthService.
+   *
+   * @param fileService - file service
+   */
+  constructor(private readonly fileService: FileService) {}
+
+  /**
+   * Check if rapidsnark is available and executable.
+   * @returns the rapidsnark executable path and its executability status
+   */
+  async checkRapidsnark(): Promise<ICheckRapidsnark> {
+    const rapidsnarkExecutablePath = process.env.COORDINATOR_RAPIDSNARK_EXE || "";
+
+    const [rapidsnarkIsAccessible, rapidsnarkIsExecutable] = await Promise.all([
+      fs.promises
+        // eslint-disable-next-line no-bitwise
+        .access(rapidsnarkExecutablePath, fs.constants.F_OK | fs.constants.X_OK)
+        .then(() => true)
+        .catch(() => false),
+      fs.promises
+        // eslint-disable-next-line no-bitwise
+        .stat(rapidsnarkExecutablePath)
+        .then((stats) => stats.isFile())
+        .catch(() => false),
+    ]);
+
+    return {
+      rapidsnarkExecutablePath,
+      rapidsnarkIsAccessible,
+      rapidsnarkIsExecutable,
+    };
+  }
+
+  /**
+   * Check if zkeys directory exists
+   * @returns the existence of the zkeys directory and the required zkeys
+   */
+  async checkZkeysDirectory(): Promise<ICheckZkeysDirectory> {
+    const zkeysDirectoryPath = path.resolve(process.env.COORDINATOR_ZKEY_PATH!);
+    const requiredZkeyNames = [
+      {
+        name: process.env.COORDINATOR_TALLY_ZKEY_NAME,
+        modes: [EMode.QV, EMode.NON_QV],
+      },
+      {
+        name: process.env.COORDINATOR_MESSAGE_PROCESS_ZKEY_NAME,
+        modes: [EMode.QV, EMode.NON_QV, EMode.FULL],
+      },
+      {
+        name: process.env.COORDINATOR_POLL_JOINING_ZKEY_NAME,
+      },
+      {
+        name: process.env.COORDINATOR_POLL_JOINED_ZKEY_NAME,
+      },
+    ];
+
+    const zkeysDirectoryExists = await fs.promises
+      .stat(zkeysDirectoryPath)
+      .then((stats) => stats.isDirectory())
+      .catch(() => false);
+
+    const availableZkeys = requiredZkeyNames.reduce<ICheckZkeysDirectory["availableZkeys"]>((acc, { name, modes }) => {
+      if (!modes) {
+        try {
+          acc[name!] = this.fileService.getZkeyFilePaths(name!);
+        } catch (error) {
+          acc[name!] = undefined;
+        }
+
+        return acc;
+      }
+
+      const modeEntries = modes.reduce<ICheckZkeysDirectory["availableZkeys"]>((modeAcc, mode) => {
+        const pathname = `${name!}_Mode_${String(mode)}`;
+
+        try {
+          // eslint-disable-next-line no-param-reassign
+          modeAcc[pathname] = this.fileService.getZkeyFilePaths(name!, mode);
+        } catch (error) {
+          // eslint-disable-next-line no-param-reassign
+          modeAcc[pathname] = undefined;
+        }
+
+        return modeAcc;
+      }, {});
+
+      return { ...acc, ...modeEntries };
+    }, {});
+
+    return {
+      zkeysDirectoryExists,
+      availableZkeys,
+    };
+  }
+
+  /**
+   * Check if coordinator wallet has funds
+   * @returns the address and its funds in all networks
+   */
+  async checkWalletFunds(): Promise<ICheckWalletFunds> {
+    let address: string;
+    const networks = Object.values(ESupportedNetworks);
+
+    try {
+      const signer = getSigner(networks[0]);
+      address = await signer.getAddress();
+    } catch (error) {
+      return {
+        address: zeroAddress,
+        fundsInNetworks: [],
+      };
+    }
+
+    const fundsInNetworks = await Promise.all(
+      networks.map(async (network) => {
+        const signer = getSigner(network);
+
+        return signer.provider
+          ?.getBalance(address)
+          .then((balance) => ({
+            network,
+            balance: formatEther(balance),
+            status: true,
+          }))
+          .catch(() => ({
+            network,
+            balance: formatEther(0n),
+            status: false,
+          }));
+      }),
+    );
+
+    return {
+      address,
+      fundsInNetworks,
+    };
+  }
+
+  /**
+   * Check coordinator service health
+   * @returns an object containing the results of all the health checks
+   */
+  async checkCoordinatorHealth(): Promise<IHealthCheckResponse> {
+    const [rapidsnark, zkeysDirectory, coordinatorWalletFunds] = await Promise.all([
+      this.checkRapidsnark(),
+      this.checkZkeysDirectory(),
+      this.checkWalletFunds(),
+    ]);
+
+    return {
+      rapidsnark,
+      zkeysDirectory,
+      coordinatorWalletFunds,
+    };
+  }
+}
--- a/apps/coordinator/ts/health/types.ts
+++ b/apps/coordinator/ts/health/types.ts
@@ -1,0 +1,91 @@
+import type { IGetZkeyFilePathsData } from "../file/types";
+
+/**
+ * Interface for checking the rapidsnark setup
+ */
+export interface ICheckRapidsnark {
+  /**
+   * Rapidsnark executable path
+   */
+  rapidsnarkExecutablePath: string;
+
+  /**
+   * Is rapidsnark accessible
+   */
+  rapidsnarkIsAccessible: boolean;
+
+  /**
+   * Is rapidsnark executable
+   */
+  rapidsnarkIsExecutable: boolean;
+}
+
+/**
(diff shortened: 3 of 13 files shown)


NicoSerranoP:
I decided to remove the coordinator.yml Github workflow until we find out why the build process is failing on it and not on local. To run the docker commands on local you can check out the coordinator's README.md

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,164 @@
...
+   */
+  checkRapidsnark(): ICheckRapidsnark {
+    const rapidsnarkExecutablePath = process.env.COORDINATOR_RAPIDSNARK_EXE || "";
+    let rapidsnarkIsExecutable = false;
+
+    try {
+      // Check if it's a file and has execute permissions
+      // eslint-disable-next-line no-bitwise
+      accessSync(rapidsnarkExecutablePath, constants.F_OK | constants.X_OK);
+      const stats = statSync(rapidsnarkExecutablePath);
```
0xmad, line 37: Do not use sync methods

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,164 @@
...
+
+      rapidsnarkIsExecutable = stats.isFile();
+    } catch (error) {
+      rapidsnarkIsExecutable = false;
+    }
```
0xmad, line 42: Why the error is silenced (applied to all catches)?

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,164 @@
...
+          if (!modes) {
+            const zkeyPath = this.fileService.getZkeyFilePaths(name!);
+            availableZkeys[name!] = zkeyPath;
+            return;
+          }
+
+          modes.forEach((mode) => {
+            const zkeyPath = this.fileService.getZkeyFilePaths(name!, mode);
+            availableZkeys[`${name!}_Mode_${String(mode)}`] = zkeyPath;
+          });
```
0xmad, line 91: It's better to use `reduce` to build the immutable `availableZkeys` instead of updating it.

apps/coordinator/ts/health/health.controller.ts
0xmad, line 1: This file should have tests

apps/coordinator/ts/health/health.service.ts
0xmad, line 1: This file should have tests

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,164 @@
...
+import { Injectable } from "@nestjs/common";
+import { formatEther } from "ethers";
+
+import { accessSync, constants, statSync } from "fs";
+import path from "path";
+
+import { ESupportedNetworks, getSigner } from "../common";
+import { FileService } from "../file/file.service";
+
+import { ICheckRapidsnark, ICheckWalletFunds, ICheckZkeysDirectory, IHealthCheckResponse } from "./types";
```
0xmad, line 11:
```suggestion
import type { ICheckRapidsnark, ICheckWalletFunds, ICheckZkeysDirectory, IHealthCheckResponse } from "./types";
```

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,164 @@
...
+   * @returns the address and its funds in all networks
+   */
+  async checkWalletFunds(): Promise<ICheckWalletFunds> {
+    let address = "0x0";
+    const fundsInNetworks: ICheckWalletFunds["fundsInNetworks"] = {};
+
+    const networks = Object.values(ESupportedNetworks);
+
+    try {
+      const signer = getSigner(ESupportedNetworks.ETHEREUM_SEPOLIA);
```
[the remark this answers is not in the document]
0xmad: But it will be using every time, not only for tests. I think you need to specify the supported network.

apps/coordinator/ts/health/health.service.ts
0xmad: There is a async version with promise version:

```
const rapidsnarkIsExecutable = await fs.promises.stat(rapidsnarkExecutablePath)
  .then((stats) => stats.isFile())
  .catch(() => false);
```

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,164 @@
...
+          fundsInNetworks[network] = {
+            balance: formatEther(balance ?? 0n),
+            status: true,
+          };
+        } catch (error) {
+          fundsInNetworks[network] = {
+            balance: formatEther(0n),
+            status: false,
+          };
+        }
```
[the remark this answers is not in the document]
NicoSerranoP: My idea of using 
```javascript
await Promise.all(
      networks.map(async (network) => {
      // ...
      });
 });
 ```
 was to get all the supported networks with their respective gas balance and their status (true if active, false if not). I have a couple of follow up question 😊 :
 
>  I think it's better to use .catch method on promise (if it can fail) instead of try/catch.

 
How could I use `.catch` in this situation? If I do `const balance = await signer.provider?.getBalance(address).catch( () => { return 0n });` I could "catch" the error but how would we know if the network is down (status false) or if only the address doesn't have funds

File: apps/coordinator/ts/health/types.ts [hunk 7]
```diff
@@ -0,0 +1,66 @@
...
+export interface ICheckWalletFunds {
+  /**
+   * Coordinator address
+   */
+  address: string;
+
+  /**
+   * Funds in all networks
+   */
+  fundsInNetworks: { network: string; balance: string; status: boolean }[];
```
0xmad, line 45: Let's move `{ network: string; balance: string; status: boolean }` to a separate interface

apps/coordinator/ts/health/health.service.ts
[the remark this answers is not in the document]
NicoSerranoP: The idea is to simply request a GET to `/v1/health/check` to get all the available info. I would prefer not adding a param to specify the supported network because at the end of the day the address will always be the same. Remember that the results show you all the supported network + balances (if the RPC does not fail. If it fails it shows balance = 0n and status = false)

I think we should not add a param for the network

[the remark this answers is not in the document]
0xmad: So, in this case this code can be removed because you get the new signer for each network below (you just need to pass signer instead of address to `getBalance` method)

apps/coordinator/ts/health/health.service.ts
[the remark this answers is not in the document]
NicoSerranoP: It fails when the RPC call fails. I just changed it to use .then() and .catch instead of try catch

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,178 @@
...
+        let zkeyPath: IGetZkeyFilePathsData | boolean;
+
+        try {
+          zkeyPath = this.fileService.getZkeyFilePaths(name!, mode);
+        } catch (error) {
+          zkeyPath = false;
+        }
+
+        return { ...modeAcc, [`${name!}_Mode_${String(mode)}`]: zkeyPath };
+      }, {});
```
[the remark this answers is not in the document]
NicoSerranoP: I tried this but I am getting a lint error: `Assignment to property of function parameter 'modeAcc'.eslint[no-param-reassign](https://eslint.org/docs/latest/rules/no-param-reassign)`

I asked Copilot and it says:

The error "Assignment to property of function parameter 'modeAcc'" occurs because ESLint is detecting that you're modifying a parameter object directly inside the `reduce` callback. This is generally discouraged as it can lead to unexpected side effects and makes the code harder to reason about.

Looking at the code in the `checkZkeysDirectory` method, the issue is in this section:

```typescript
const modeEntries = modes.reduce((modeAcc: ICheckZkeysDirectory["availableZkeys"], mode) => {
  const pathname = `${name!}_Mode_${String(mode)}`;

  try {
    modeAcc[pathname] = this.fileService.getZkeyFilePaths(name!, mode); // ← This line modifies the parameter
  } catch (error) {
    modeAcc[pathname] = undefined; // ← This line also modifies the parameter
  }

  return modeAcc;
}, {});
```

The best practice is to avoid mutating the accumulator parameter directly. Instead, you should return a new object that includes the previous properties plus the new ones. This follows the functional programming principle of immutability and makes the code more predictable.

To fix this, you should use the spread operator to create a new object instead of modifying the existing one:

```typescript
const modeEntries = modes.reduce((modeAcc: ICheckZkeysDirectory["availableZkeys"], mode) => {
  const pathname = `${name!}_Mode_${String(mode)}`;
  let zkeyPath;

  try {
    zkeyPath = this.fileService.getZkeyFilePaths(name!, mode);
  } catch (error) {
    zkeyPath = undefined;
  }

  return {
    ...modeAcc,
    [pathname]: zkeyPath,
  };
}, {});
```

This approach creates a new object for each iteration rather than mutating the existing accumulator, which is safer and follows functional programming best practices.

apps/coordinator/ts/health/health.service.ts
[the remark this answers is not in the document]
0xmad: I think you just need to use the right key type for the object and it will be fixed

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,178 @@
...
+            balance: formatEther(balance ?? 0n),
+            status: true,
+          };
+        } catch (error) {
+          return {
+            network,
+            balance: formatEther(0n),
+            status: false,
+          };
+        }
```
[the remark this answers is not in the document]
NicoSerranoP: There are a couple of throw errors in the `getSigner` -> `getRpcUrl`: https://github.com/privacy-scaling-explorations/maci/blob/8eef7d12c050582d8846268998cda80a93585e33/apps/coordinator/ts/common/chain.ts#L16

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,176 @@
...
+  constructor(private readonly fileService: FileService) {}
+
+  /**
+   * Check if rapidsnark is available and executable.
+   * @returns the rapidsnark executable path and its executability status
+   */
+  async checkRapidsnark(): Promise<ICheckRapidsnark> {
+    const rapidsnarkExecutablePath = process.env.COORDINATOR_RAPIDSNARK_EXE || "";
+
+    const rapidsnarkIsAccessible = await promises
```
0xmad, line 33: Use `fs.promises` instead (applied to constants too)

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,176 @@
+import { EMode } from "@maci-protocol/sdk";
+import { Injectable } from "@nestjs/common";
+import { formatEther } from "ethers";
+import { zeroAddress } from "viem";
+
+import { constants, statSync, promises } from "fs";
```
0xmad, line 6: Do not use statSync

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,176 @@
...
+      // eslint-disable-next-line no-bitwise
+      .access(rapidsnarkExecutablePath, constants.F_OK | constants.X_OK)
+      .then(() => true)
+      .catch(() => false);
+
+    const rapidsnarkIsExecutable = await promises
+      // eslint-disable-next-line no-bitwise
+      .stat(rapidsnarkExecutablePath)
+      .then((stats) => stats.isFile())
+      .catch(() => false);
```
0xmad, line 43: Use `Promise.all`

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,176 @@
...
+        try {
+          acc[name!] = this.fileService.getZkeyFilePaths(name!);
+        } catch (error) {
+          acc[name!] = undefined;
+        }
+
+        return acc;
+      }
+
+      const modeEntries = modes.reduce((modeAcc: ICheckZkeysDirectory["availableZkeys"], mode) => {
```
0xmad, line 95: And you don't need to disable eslint inside reduce
```suggestion
      const modeEntries = modes.reduce<ICheckZkeysDirectory["availableZkeys"]>((modeAcc, mode) => {
```

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,176 @@
...
+    let zkeysDirectoryExists = false;
+
+    try {
+      const stats = statSync(zkeysDirectoryPath);
+      zkeysDirectoryExists = stats.isDirectory();
+    } catch (error) {
+      zkeysDirectoryExists = false;
+    }
+
+    const availableZkeys = requiredZkeyNames.reduce((acc: ICheckZkeysDirectory["availableZkeys"], { name, modes }) => {
```
0xmad, line 84:
```suggestion
    const availableZkeys = requiredZkeyNames.reduce<ICheckZkeysDirectory["availableZkeys"]>((acc, { name, modes }) => {
```

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,176 @@
...
+      const signer = getSigner(networks[0]);
+      address = await signer.getAddress();
+    } catch (error) {
+      return {
+        address: zeroAddress,
+        fundsInNetworks: [],
+      };
+    }
+
+    const fundsInNetworks: ICheckWalletFunds["fundsInNetworks"] = await Promise.all(
```
0xmad, line 136:
```suggestion
    const fundsInNetworks = await Promise.all(
```

apps/coordinator/ts/health/types.ts
```diff
@@ -0,0 +1,91 @@
+import { IGetZkeyFilePathsData } from "../file/types";
```
0xmad, line 1:
```suggestion
import type { IGetZkeyFilePathsData } from "../file/types";
```

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /**
+   * Check coordinator service health
+   * @returns an object containing the results of all the health checks
+   */
+  async checkCoordinatorHealth(): Promise<IHealthCheckResponse> {
+    const rapidsnark = await this.checkRapidsnark();
+    const zkeysDirectory = this.checkZkeysDirectory();
+    const coordinatorWalletFunds = await this.checkWalletFunds();
```
0xmad, line 168: Use `Promise.all`

apps/coordinator/ts/health/health.service.ts
NicoSerranoP: Anton in my computer the eslint error is not desappearing with this solution:

image


Please let's merge this PR asap because I need to focus on other issues

JohnGuilding (approved):

apps/coordinator/ts/health/health.service.ts
```diff
@@ -0,0 +1,164 @@
...
+   * Check if coordinator wallet has funds
+   * @returns the address and its funds in all networks
+   */
+  async checkWalletFunds(): Promise<ICheckWalletFunds> {
+    let address = "0x0";
+    const fundsInNetworks: ICheckWalletFunds["fundsInNetworks"] = {};
+
+    const networks = Object.values(ESupportedNetworks);
+
+    try {
```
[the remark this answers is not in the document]
NicoSerranoP: Yes! In some cases the code inside the try block could fail:

1. `const signer = getSigner(networks[0]);`: if an invalid supported network is set (e.g. Fuel L2)
2. `address = await signer.getAddress();`: a supported network is set (e.g. ETH Sepolia) but the RPC does not have authorization to be used in that network

0xmad (approved):
@NicoSerranoP thanks!

apps/coordinator/ts/health/health.service.ts
NicoSerranoP: @0xmad I just squashed the commits and activated the auto merge! 🔥 I need your approval in the requested changes

PR #2643, opened 2025-07-14
fix(contracts): small typo

VolodymyrBg:
pointAdd function incorrectly used y1 instead of _y2 for the second point. 
I know you don't accept typo PRs, but this one looks reasonable and important


NicoSerranoP (approved):

NicoSerranoP:
@VolodymyrBg thank you for this! Could you rename the commit to follow the commit lint rules. It could be something like:

fix(contracts): small typo


Please consider using lowercase at the beginning of the commit and using fix, feat, chore, doc

0xmad:
And make sure there is only one commit

PR #2651, opened 2025-07-16
feat(coordinator): add redis module

NicoSerranoP:
# Description

Added the Redis module for a persistent database, some utility functions and types. Also created some tests. It is a sub PR of #2642 

I decided to divide that PR in:

1. ✅  Redis module
2. Scheduler module
3. Finalize function in proof module (merge, generate and submit)
4. Get deployment poll

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/.github/workflows/coordinator.yml
+++ b/.github/workflows/coordinator.yml
@@ -21,6 +21,8 @@ env:
   COORDINATOR_RPC_URL: ${{ secrets.COORDINATOR_RPC_URL }}
   COORDINATOR_ADDRESSES: ${{ vars.COORDINATOR_ADDRESSES }}
   COORDINATOR_ALLOWED_ORIGINS: ${{ vars.COORDINATOR_ALLOWED_ORIGINS }}
+  COORDINATOR_REDIS_HOST: ${{ vars.COORDINATOR_REDIS_HOST }}
+  COORDINATOR_REDIS_PORT: ${{ vars.COORDINATOR_REDIS_PORT }}
   COORDINATOR_PORT: ${{ vars.COORDINATOR_PORT }}
   SUBGRAPH_NAME: ${{ vars.COORDINATOR_SUBGRAPH_NAME }}
   SUBGRAPH_PROVIDER_URL: ${{ vars.COORDINATOR_SUBGRAPH_PROVIDER_URL }}
@@ -37,6 +39,12 @@ concurrency:
 jobs:
   build-and-test:
     runs-on: ubuntu-22.04
+
+    services:
+      redis:
+        image: redis:8
+        ports:
+          - 6379:6379
 
     steps:
       - uses: actions/checkout@v4
--- a/apps/coordinator/Dockerfile
+++ b/apps/coordinator/Dockerfile
@@ -7,17 +7,10 @@ WORKDIR /builder
 # Copy the entire monorepo to maintain workspace structure
 COPY . .
 
-WORKDIR /builder/apps/coordinator
-
-# Update @maci-protocol package versions in coordinator specifically
-RUN sed -i 's/"@maci-protocol\/contracts": "[^"]*"/"@maci-protocol\/contracts": "ci"/g' package.json && \
-    sed -i 's/"@maci-protocol\/crypto": "[^"]*"/"@maci-protocol\/crypto": "ci"/g' package.json && \
-    sed -i 's/"@maci-protocol\/domainobjs": "[^"]*"/"@maci-protocol\/domainobjs": "ci"/g' package.json && \
-    sed -i 's/"@maci-protocol\/sdk": "[^"]*"/"@maci-protocol\/sdk": "ci"/g' package.json && \
-    sed -i 's/"@maci-protocol\/subgraph": "[^"]*"/"@maci-protocol\/subgraph": "ci"/g' package.json
-
 RUN pnpm install
 RUN pnpm run build
+
+WORKDIR /builder/apps/coordinator
 
 # Create image by copying build artifacts
 FROM node:20-slim AS runner
--- a/apps/coordinator/docker-compose.yml
+++ b/apps/coordinator/docker-compose.yml
@@ -1,4 +1,16 @@
 services:
+  coordinator-database:
+    container_name: coordinator-database
+    image: redis:latest
+    restart: unless-stopped
+    env_file:
+      - ./.env
+    ports:
+      - "${COORDINATOR_REDIS_PORT:-6379}:${COORDINATOR_REDIS_PORT:-6379}"
+    command: redis-server --save 20 1 --loglevel warning
+    volumes:
+      - coordinator-redis:/data
+
   coordinator-service:
     container_name: coordinator-service
     build:
@@ -9,11 +21,16 @@ services:
     environment:
       # Check Dockerfile to see where bin is downloaded
       - COORDINATOR_RAPIDSNARK_EXE=/home/node/rapidsnark/build/prover
+      - COORDINATOR_REDIS_HOST=coordinator-database
+      - COORDINATOR_REDIS_PORT=${COORDINATOR_REDIS_PORT:-6379}
     volumes:
       - coordinator-data:/app
     ports:
       - "${COORDINATOR_PORT:-3001}:${COORDINATOR_PORT:-3001}"
 
 volumes:
+  coordinator-redis:
+    driver: local
+
   coordinator-data:
     driver: local
--- a/apps/coordinator/ts/redis/__tests__/redis.service.test.ts
+++ b/apps/coordinator/ts/redis/__tests__/redis.service.test.ts
@@ -1,0 +1,168 @@
+import { createClient, RedisArgument, RedisClientType } from "@redis/client";
+
+import { ESupportedNetworks } from "../../common";
+import { RedisService } from "../redis.service";
+import { IScheduledPoll } from "../types";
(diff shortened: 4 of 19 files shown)


apps/coordinator/ts/redis/redis.service.ts
```diff
@@ -0,0 +1,53 @@
...
+import { createClient, RedisClientType } from "@redis/client";
+// import { createClient, RedisClientType } from "redis";
+
+@Injectable()
+export class RedisService implements OnModuleInit {
+  client: RedisClientType;
+
+  constructor() {
+    this.client = createClient({
+      url: `redis://${process.env.COORDINATOR_REDIS_HOST || "redis"}
```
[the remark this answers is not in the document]
JohnGuilding: How are you finding using copilot for reviews out of interest? What percentage of it's comments are useful and actionable comments?

.github/workflows/coordinator.yml
```diff
@@ -38,6 +40,12 @@ jobs:
   build-and-test:
     runs-on: ubuntu-22.04
 
+    services:
+      redis:
+        image: redis:7
+        ports:
+          - 6379:6379
```
JohnGuilding, line 47: 1. What does `redis:7` mean? Why the 7?
2. Why choose `6379` out of interest?

apps/coordinator/Dockerfile
```diff
@@ -7,18 +7,11 @@ WORKDIR /builder
 # Copy the entire monorepo to maintain workspace structure
 COPY . .
 
-WORKDIR /builder/apps/coordinator
-
-# Update @maci-protocol package versions in coordinator specifically
```
JohnGuilding, line 12: Why did we need to do this? And why did we remove it?

apps/coordinator/docker-compose.yml
```diff
@@ -1,4 +1,16 @@
 services:
+  coordinator-database:
+    container_name: coordinator-database
+    image: redis:latest
+    restart: unless-stopped
+    env_file:
+      - ./.env
+    ports:
+      - "${COORDINATOR_REDIS_PORT:-6379}:${COORDINATOR_REDIS_PORT:-6379}"
+    command: redis-server --save 20 1 --loglevel warning
```
JohnGuilding, line 10: What do 20 and 1 mean here?

apps/coordinator/ts/redis/redis.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  async get(key: string): Promise<string | null> {
+    return this.client.get(key);
+  }
+
+  async getAll(match?: string): Promise<string[]> {
+    let result: string[] = [];
+    let currentCursor = "0";
+
+    do {
+      // using a cursor is recommended to be use in production environment
```
JohnGuilding, line 35: Nice thanks for the comment

apps/coordinator/ts/redis/__tests__/redis.service.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (keys.length) {
+      await service.client.del(keys);
+    }
+  });
+
+  afterAll(async () => {
+    // Clean up after tests
+    const keys = await service.client.keys(REDIS__GET_ALL_PREFIX);
+    if (keys.length) {
+      await service.client.del(keys);
```
JohnGuilding, line 33: Do we need the same logic to delete values in `beforeAll` and `afterAll`?

apps/coordinator/ts/redis/__tests__/redis.service.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  it("should return null when non-existent key is requested", async () => {
+    const nonExistentKey = "non-existent-key";
+    const result = await service.get(nonExistentKey);
+
+    expect(result).toBeNull();
+  });
+
+  it("should set and get multiple values", async () => {
+    const pollOne = JSON.parse(JSON.stringify(scheduledPoll)) as IScheduledPoll;
```
JohnGuilding, line 65: Nit: Do we need to stringify then parse every time? Could we just set pollId and then `stringify`?

apps/coordinator/ts/redis/__tests__/redis.service.test.ts
JohnGuilding: Based on another comment, if we wanted isolated unit tests, we could consider `afterEach`

apps/coordinator/ts/redis/__tests__/redis.service.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    const storedPollThree = JSON.parse((await service.get(pollKeyThree)) || "{}") as IScheduledPoll;
+
+    expect(pollOne).toEqual(storedPollOne);
+    expect(pollTwo).toEqual(storedPollTwo);
+    expect(pollThree).toEqual(storedPollThree);
+  });
+
+  it("should retrieve all values using getAll", async () => {
+    // 1 entry added in the set and get value test
+    // 3 entries added in the set and get multiple values test
```
JohnGuilding, line 91: Not too opinionated but what do you think about writing isolated unit tests which do not rely on state from previous tests?

It could cause unexpected failures if other tests would change for example. For this sort of logic where we want to test sequential state change logic, we could write an integration test/dedicated unit test depending on how many components it involved

apps/coordinator/ts/redis/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { ESupportedNetworks } from "../common";
+
+import { IScheduledPoll } from "./types";
+
+export const getPollKeyForRedis = (
+  chain: ESupportedNetworks,
+  maciAddress: string,
+  pollId: string,
+  test = false,
```
JohnGuilding, line 9: What's the advantage of having test keys and non-test keys?

JohnGuilding (approved):
Thanks Nico, this was super nice to review!

Left quite a few questions for my understanding, then also a few non-blocking comments

.github/workflows/coordinator.yml
NicoSerranoP: 1. it is the version. I increased it to redis 8 which is the current one. Thanks for the catch 🚀 
2. 6379 is the default redis port

apps/coordinator/docker-compose.yml
NicoSerranoP: It means that redis will persist the nada if at least 1 write operation has ocurrend within 20 seconds

apps/coordinator/ts/redis/types.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Whether the MACI contract's state root has been merged
+   */
+  merged?: boolean;
+
+  /**
+   * Whether the proofs has been generated
+   */
+  proofsGenerated?: boolean;
```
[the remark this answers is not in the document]
NicoSerranoP: 1. My initial idea did not had `merged` and proofsGenerated` but then I noticed that we could lose connection between steps. So a poll that has been merged should go to generate and not try to merge again (this would actually throw an error). I am making them not optional 🚀 

2. The state matters to the coordinator because we cannot merge twice so if a poll has been merged, the connection broke and then we retry again we will get an "already merged" error

apps/coordinator/ts/redis/redis.service.ts
```diff
@@ -0,0 +1,84 @@
+import { Injectable, OnModuleInit } from "@nestjs/common";
+import { createClient, RedisClientType } from "@redis/client";
+
+/**
+ * Redis service to interact with the Redis database.
+ */
+@Injectable()
+export class RedisService implements OnModuleInit {
+  client: RedisClientType;
```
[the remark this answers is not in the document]
NicoSerranoP: For testing we need client to be public. Should I instantiate a client directly in the testing file instead of doing `redisService.client`?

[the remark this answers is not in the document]
0xmad: Why do you need client for unit test? Unit tests shouldn't contain any external dependencies, they should be mocked and we can run unit tests without using real databases, dependencies and etc (we can use them only for integration/e2e tests).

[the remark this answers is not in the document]
NicoSerranoP: @0xmad I just pushed some changes:

1. made `client` private
2. add the `isOpen` call into the health module (to check if the database is correctly connected)
3. created a e2e.redis.test.ts to test connection with a Redis database (it is the previous test file but moved to the e2e tests directory)
4. created a unit tests file for Redis (I am mocking all returned values from client to verify that the redis service is working correctly)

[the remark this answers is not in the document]
0xmad: Nice, thanks @NicoSerranoP !

apps/coordinator/ts/redis/__tests__/redis.service.test.ts
```diff
@@ -0,0 +1,167 @@
...
+    const pollThree = { ...scheduledPoll };
+    pollThree.pollId = "3";
+    const pollKeyThree = getPollKeyFromObject(pollThree, true);
+    await service.set(pollKeyThree, JSON.stringify(pollThree));
+
+    mockRedisClient.get.mockImplementation(async (key: RedisArgument) => {
+      if (["1", "2", "3"].includes(String(key))) {
+        return Promise.resolve("{}");
+      }
+      return Promise.resolve(null);
```
0xmad, line 130: Please, add new line after if

Issue #2649, opened 2025-07-15
feat(contracts): get deployment block from MACI and Poll


NicoSerranoP:
Add a `getDeploymentBlock()` function in MACI.sol/Poll.sol to retrieve the block when the MACI/Poll contract was deployed. Otherwise it is hard to get the deployment block programatically and the generate proofs function could take forever

PR #2655, opened 2025-07-17
feat: store maci and poll deployment block

ctrlc03:
# Description

Store maci and poll deployment blocks 
## Related issue(s)

re #2649 
## Confirmation

> [!IMPORTANT]


0xmad:
Agree with @JohnGuilding here that it can be handled offchain and there is no need to add this logic onchain.
Also, we have subgraph where we can add deploy block and fetch this data from there.

Issue #2523, opened 2025-06-05
feat(coordinator): keep track of poll finalization status to avoid race conditions


NicoSerranoP:
Imagine poll 15 has ended and two people called the coordinator service to finalize it. The last one to call should receive a "this poll is finalizing at the moment" message

NicoSerranoP:
We might not need this because end users will only schedule poll finalization and this race condition won't happen / will be handled by the coordinator https://github.com/privacy-scaling-explorations/maci/pull/2675

Issue #2638, opened 2025-07-11
feat(coordinator): implement AA


NicoSerranoP:
# Prerequisites

Please answer the following question for yourself before submitting a feature request.

- [X] I checked to make sure that this request has not already been filed

# Motivation

We were trying to integrate ZeroDev and AA (EIP-7702 specifically) into the MACI coordinator. I think we should continue doing this because it helps out in:

1. Could prevent possible attack in https://github.com/privacy-scaling-explorations/maci/issues/2369#issuecomment-3020929200 by allowing coordinators to have prepaid wallets that would submit the tally results
2. Improve UX by allowing users to use their wallet to finalize a poll

# Details

Work in progress.

Make sure that you have reviewed [MACI's Contributor Guidelines](https://maci.pse.dev/docs/contributing)

PR #2671, opened 2025-07-25
feat(coordinator): scheduler module

NicoSerranoP:
# Prerequisities
We need to merge this PR first: https://github.com/privacy-scaling-explorations/maci/pull/2670

# Description

1. Implement scheduler module to store and schedule polls for finalization
2. Tests for service and controller

## Related issue(s)

Closes https://github.com/privacy-scaling-explorations/maci/issues/2369 and https://github.com/privacy-scaling-explorations/maci/issues/2523

## Confirmation

> [!IMPORTANT]


apps/coordinator/ts/scheduler/__tests__/scheduler.controller.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      try {
+        await controller.register(scheduledPoll);
+
+        fail("Expected error was not thrown");
+      } catch (error: unknown) {
+        const returnedError = error as Error;
+
+        expect(returnedError.message).toBe(errorMessage);
+        expect(returnedError).toHaveProperty("status", HttpStatus.BAD_REQUEST);
+      }
```
0xmad, line 58: Why do you use try/catch here? It's possible to check rejected values with jest.

apps/coordinator/ts/scheduler/__tests__/scheduler.controller.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      try {
+        await controller.status(scheduledPoll);
+
+        fail("Expected error was not thrown");
+      } catch (error: unknown) {
+        const returnedError = error as Error;
+
+        expect(returnedError.message).toBe(errorMessage);
+        expect(returnedError).toHaveProperty("status", HttpStatus.BAD_REQUEST);
+      }
```
0xmad, line 82: Same here

apps/coordinator/ts/scheduler/__tests__/scheduler.controller.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      try {
+        await controller.delete(scheduledPoll);
+
+        fail("Expected error was not thrown");
+      } catch (error: unknown) {
+        const returnedError = error as Error;
+
+        expect(returnedError.message).toBe(errorMessage);
+        expect(returnedError).toHaveProperty("status", HttpStatus.BAD_REQUEST);
+      }
```
0xmad, line 108: Same here

apps/coordinator/ts/scheduler/__tests__/scheduler.service.test.ts
```diff
@@ -0,0 +1,362 @@
...
+      });
+      expect(timeouts.length).toBe(polls.length);
+    });
+
+    it("should not restore timeouts for tallied polls", async () => {
+      (getPoll as jest.Mock).mockImplementation(({ pollId }) => {
+        if (pollId === 2n) {
+          return Promise.resolve({ isMerged: true });
+        }
+        return Promise.resolve({ isMerged: false });
```
0xmad, line 327:
```suggestion
        return Promise.resolve({ isMerged: pollId === 2n });
```

apps/coordinator/ts/scheduler/__tests__/scheduler.service.test.ts
```diff
@@ -0,0 +1,362 @@
...
+      });
+      (isTallied as jest.Mock).mockImplementation(({ pollId }) => {
+        if (pollId === "2") {
+          return Promise.resolve(true);
+        }
+        return Promise.resolve(false);
```
0xmad, line 333:
```suggestion
        return Promise.resolve(pollId === "2");
```

apps/coordinator/ts/scheduler/dto.ts
```diff
@@ -0,0 +1,95 @@
...
+    enum: ESupportedNetworks,
+  })
+  @IsEnum(ESupportedNetworks)
+  chain!: ESupportedNetworks;
+}
+
+/**
+ * Data transfer object for registering a poll for finalization
+ */
+export class SchedulePollWithSignerDto extends IdentityScheduledPollDto {
```
0xmad, line 49: Why is it called `SchedulePollWithSignerDto` and `sessionKeyAddress` and `approval` are optional?

apps/coordinator/ts/scheduler/scheduler.controller.ts
```diff
@@ -0,0 +1,84 @@
+import { Body, Controller, HttpException, HttpStatus, Logger, Post } from "@nestjs/common";
+import { ApiResponse, ApiTags } from "@nestjs/swagger";
+
+import { IIdentityScheduledPoll } from "../redis/types";
```
0xmad, line 4:
```suggestion
import type { IIdentityScheduledPoll } from "../redis/types";
```

apps/coordinator/ts/scheduler/scheduler.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+  async register(@Body() args: SchedulePollWithSignerDto): Promise<IIsPollScheduledResponse> {
+    const schedulePollWithSignerArgs = {
+      ...args,
+      pollId: args.pollId.toString(),
+      endDate: 0,
+      merged: false,
+      proofsGenerated: false,
+    } as ISchedulePollWithSignerArgs;
+
+    return this.schedulerService.registerPoll(schedulePollWithSignerArgs).catch((error: Error) => {
```
0xmad, line 40:
```suggestion
    return this.schedulerService.registerPoll(schedulePollWithSignerArgs).catch((error) => {
```

apps/coordinator/ts/scheduler/scheduler.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+  @ApiResponse({ status: HttpStatus.OK, description: "Poll finalization status" })
+  @ApiResponse({ status: HttpStatus.NOT_FOUND, description: "BadRequest" })
+  @Post("status")
+  async status(@Body() args: IdentityScheduledPollDto): Promise<IIsPollScheduledResponse> {
+    const identityScheduledPoll = {
+      ...args,
+      pollId: args.pollId.toString(),
+    } as IIdentityScheduledPoll;
+
+    return this.schedulerService.isPollScheduled(identityScheduledPoll).catch((error: Error) => {
```
0xmad, line 60:
```suggestion
    return this.schedulerService.isPollScheduled(identityScheduledPoll).catch((error) => {
```

apps/coordinator/ts/scheduler/scheduler.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+  @ApiResponse({ status: HttpStatus.OK, description: "Poll finalization has been deleted" })
+  @ApiResponse({ status: HttpStatus.NOT_FOUND, description: "BadRequest" })
+  @Post("delete")
+  async delete(@Body() args: IdentityScheduledPollDto): Promise<IIsPollScheduledResponse> {
+    const identityScheduledPoll = {
+      ...args,
+      pollId: args.pollId.toString(),
+    } as IIdentityScheduledPoll;
+
+    return this.schedulerService.deleteScheduledPoll(identityScheduledPoll).catch((error: Error) => {
```
0xmad, line 79:
```suggestion
    return this.schedulerService.deleteScheduledPoll(identityScheduledPoll).catch((error) => {
```

apps/coordinator/ts/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,282 @@
...
+import { Injectable, Logger, OnModuleInit } from "@nestjs/common";
+import { SchedulerRegistry } from "@nestjs/schedule";
+
+import { ErrorCodes } from "../common";
+import { RedisService } from "../redis/redis.service";
+import { IIdentityScheduledPoll, IScheduledPoll } from "../redis/types";
+import { getPollKeyFromObject } from "../redis/utils";
+import { SessionKeysService } from "../sessionKeys/sessionKeys.service";
+
+import {
```
0xmad, line 11:
```suggestion
import type {
```

apps/coordinator/ts/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,282 @@
+import { getPoll, isTallied } from "@maci-protocol/sdk";
+import { Injectable, Logger, OnModuleInit } from "@nestjs/common";
+import { SchedulerRegistry } from "@nestjs/schedule";
+
+import { ErrorCodes } from "../common";
+import { RedisService } from "../redis/redis.service";
+import { IIdentityScheduledPoll, IScheduledPoll } from "../redis/types";
```
0xmad, line 7:
```suggestion
import type { IIdentityScheduledPoll, IScheduledPoll } from "../redis/types";
```

apps/coordinator/ts/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,282 @@
...
+   * @returns true if poll is scheduled for finalization, false otherwise
+   */
+  async isPollScheduled(scheduledPoll: IIdentityScheduledPoll): Promise<IIsPollScheduledResponse> {
+    const key = getPollKeyFromObject(scheduledPoll);
+    const { maciAddress, pollId, chain } = scheduledPoll;
+
+    const storedPoll = await this.redisService.get(key);
+    if (!storedPoll) {
+      return { isScheduled: false };
+    }
```
0xmad, line 61:
```suggestion
    const storedPoll = await this.redisService.get(key);

    if (!storedPoll) {
      return { isScheduled: false };
    }
```

apps/coordinator/ts/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    return { isScheduled: true };
+  }
+
+  /**
+   * Return the poll end date and whether it is tallied
+   * @param args - Arguments to retrieve poll data
+   */
+  async pollEndDateAndIsTallied({
```
0xmad, line 90: Let's rename it. It should be a verb like `getPollFinalizationData` or something like this and let's add return type to docs.

0xmad (changes requested):
@NicoSerranoP thanks, left some comments

apps/coordinator/ts/scheduler/__tests__/scheduler.controller.test.ts
NicoSerranoP: it was because when using `await expect(controller.register(scheduledPoll).rejects.Throw(...)` I could not chain together `.reject.Throw(errorMessage).toHaveProperty("status", HttpStatus.BAD_REQUEST)`

I decided to go this way: one call, one fail and 2 expects.

I could do: 2 calls, 2 fails and 2 expects if it helps readibility 😊

apps/coordinator/ts/scheduler/dto.ts
NicoSerranoP: `sessionKeyAddress` and `approval` are used for the AA signer (that uses session keys). The AA signer was giving us some problems in the deploy poll step so we decided to use a normal ethers js signer if `sessionKeyAddress` and `approval` were not provided.

I hope we fix the issues with the AA signers and users will be able to use the normal ethers js and the AA signer. I have an issue to keep track of this: https://github.com/privacy-scaling-explorations/maci/issues/2638

apps/coordinator/ts/scheduler/__tests__/scheduler.controller.test.ts
NicoSerranoP: I was able to restructure it different and more clean. It is using the same as https://github.com/privacy-scaling-explorations/maci/blob/3b89cb83f2a10f890622f60097055115dda96777/apps/coordinator/ts/proof/__tests__/proof.controller.test.ts#L114

apps/coordinator/ts/scheduler/scheduler.controller.ts
0xmad: Oh, didn't know that it doesn't infer the type automatically. If so, let's stick to the first option, it's better than explicit casting.

apps/coordinator/ts/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,282 @@
...
+      endDate: Number(endDate),
+      isPollTallied: isMerged ? isPollTallied : false,
+    };
+  }
+
+  /**
+   * Store poll information and schedule its finalization.
+   * @param args - Arguments for storing and scheduling poll finalization
+   */
+  async saveAndSchedulePoll({
```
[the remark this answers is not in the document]
0xmad: If you need to use it for mocks, it's possible to mock it without public access.
But if you want to test it, it should be tested through public methods, not private.
Just let me know how you use it and let's decide (not sure about protected function, seems like it's too much for testing).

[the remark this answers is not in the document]
NicoSerranoP: I decided to use `registerPoll` to test internally `setupPollFinalization`. It is working the same way as before but we are not explicitly executing `setupPollFinalization` tests.

1. The method is private
2. it is been tested through public networks
3. We are not using `protected` nor binding methods

NicoSerranoP:
@0xmad let me know when it is ready and I will squash the commits and merge it 😊

apps/coordinator/ts/scheduler/__tests__/scheduler.service.test.ts
```diff
@@ -0,0 +1,338 @@
...
+      const { isPollTallied } = await service.getPollFinalizationData({
+        maciAddress: scheduledPoll.maciAddress,
+        pollId: scheduledPoll.pollId,
+        chain: scheduledPoll.chain,
+      });
+
+      expect(isPollTallied).toBe(false);
+    });
+
+    it("should return isPollTallied as false if poll is not merged and is tallied (it is impossible)", async () => {
```
JohnGuilding, line 166: Why do we need this test if it is impossible?

JohnGuilding (approved):
Looks good!

I think we are [overusing](https://dev.to/alexanderop/the-problem-with-as-in-typescript-why-its-a-shortcut-we-should-avoid-2km4) `as`. We could consider enabling https://typescript-eslint.io/rules/no-unsafe-type-assertion/

apps/coordinator/ts/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,293 @@
...
+    await this.redisService.set(
+      key,
+      JSON.stringify({
+        maciAddress,
+        pollId,
+        mode,
+        deploymentBlockNumber,
+        chain,
+        endDate,
+      } as IScheduledPoll),
```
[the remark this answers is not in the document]
NicoSerranoP: I am enforcing the json value to be `IScheduledPoll` so in the future if we add or remove a property in the interface, the es-lint functionality will throw an error here until we add or remove the property here.

Without `as IScheduledPoll` we will not have the warning and we could be saving incomplete data in redis

apps/coordinator/ts/scheduler/scheduler.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+  @ApiResponse({ status: HttpStatus.BAD_REQUEST, description: "BadRequest" })
+  @Post("register")
+  async register(@Body() args: SchedulePollWithSignerDto): Promise<IIsPollScheduledResponse> {
+    const schedulePollWithSignerArgs = {
+      ...args,
+      pollId: args.pollId.toString(),
+      endDate: 0,
+      merged: false,
+      proofsGenerated: false,
+    } as ISchedulePollWithSignerArgs;
```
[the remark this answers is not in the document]
NicoSerranoP: In this case, the controller function `register` receives the args as `SchedulePollWithSignerDto` but the service function `registerPoll` receives the args as  `ISchedulePollWithSignerArgs`.


We initially agree that the DTOs interfaces/classes should not be used as arg types in the service functions. In our case the DTO does not have the enough properties to be arguments of the service function (`merged: false` and `proofsGenerated: false`)

apps/coordinator/ts/scheduler/scheduler.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    return this.schedulerService.registerPoll(schedulePollWithSignerArgs).catch((error: Error) => {
+      this.logger.error(`Error:`, error);
+      throw new HttpException(error.message, HttpStatus.BAD_REQUEST);
```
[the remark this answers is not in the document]
NicoSerranoP: Totally agree. Initially the coordinator service implemented only 400 bad requests but that is not scalable. I just created an issue for it: https://github.com/privacy-scaling-explorations/maci/issues/2677

apps/coordinator/ts/scheduler/scheduler.service.ts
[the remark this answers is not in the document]
JohnGuilding: I'm not sure I follow. The issue is that `as` does not enforce any type safety. We are ignoring the typescript compiler and saying trust me this type is safe, instead of enforcing the type at compile time - this is the fundamental problem with using `as`, we are basically saying no thank you we do not want to use you typescript. 

That said, we also have the same type (`IScheduledPoll`) coming in as an argument into the function right? So that is already providing some type checking.

If we wanted to enforce the type we could do something like the following:

```ts
// option #1 - set the object to the type
const scheduledPoll: IScheduledPoll = {
  maciAddress,
  pollId,
  mode,
  deploymentBlockNumber,
  chain,
  endDate,
};
await this.redisService.set(key, JSON.stringify(scheduledPoll));

// option #2 - use `satisfies`
const scheduledPoll = {
  maciAddress,
  pollId,
  mode,
  deploymentBlockNumber,
  chain,
  endDate,
} satisfies IScheduledPoll;
await this.redisService.set(key, JSON.stringify(scheduledPoll));
```

Let me know if I am misunderstanding something related to the eslint checking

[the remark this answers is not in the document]
NicoSerranoP: 1. I decided to use the whole poll object and therefore the ` as ` is deleted in the next PR: https://github.com/privacy-scaling-explorations/maci/blob/8e8dd14da3f55e626b91d31f203bcf001d42d07f/apps/coordinator/ts/scheduler/scheduler.service.ts#L192


2. I searched for others ` as ` in this file and I found:
https://github.com/privacy-scaling-explorations/maci/blob/25347a43205a391d1343313ae4c35562a565272d/apps/coordinator/ts/scheduler/scheduler.service.ts#L257

I tried using option 1 (set type to the variable) but I am getting this lint error
image


3. I searched for others ` as ` in the coordinator service directory (excluding tests) and changed them to use option 1. Most of them were in the scheduler controller. There are some offsets in file service and session keys service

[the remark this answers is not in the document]
JohnGuilding: > I decided to use the whole poll object and therefore the as is deleted in the next PR:

Ok sweet sounds good.

> I searched for others as in this file and I found:

`const poll = JSON.parse(value) as IScheduledPoll; ` using `as` when parsing json is fine here imo as you are already dealing with `any`, and it is coming from our own code, so we know we already serialised the correct object. 

If it was coming from an external service, we would trust it less and the case for validation would be stronger, but I think it's fine in this case.

If you did want to validate it in the future, you could use typeguards as mentioned in [this article](https://dev.to/alexanderop/the-problem-with-as-in-typescript-why-its-a-shortcut-we-should-avoid-2km4). 

> I tried using option 1 (set type to the variable) but I am getting this lint error

I think it is fine to use `as` here

apps/coordinator/ts/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,293 @@
...
+      isTallied({
+        maciAddress,
+        pollId: String(pollId),
+        signer,
+      }),
+    ]);
+
+    return {
+      endDate: Number(endDate),
+      isPollTallied: isMerged ? isPollTallied : false,
```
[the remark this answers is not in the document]
NicoSerranoP: I delete the `isMerged` in the next PR https://github.com/privacy-scaling-explorations/maci/pull/2675 🫡

apps/coordinator/ts/file/file.service.ts
```diff
@@ -68,7 +68,8 @@ export class FileService {
    * @returns session key
    */
   getSessionKey(address: string): Hex | undefined {
-    return this.db.get(address).value() as Hex | undefined;
+    const sessionkey: Hex | undefined = this.db.get(address).value();
+    return sessionkey;
```
0xmad, line 72: Why did you change it?

apps/coordinator/ts/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,292 @@
...
+    const toBeExecutedIn = shouldBeExecutedIn > 0 ? shouldBeExecutedIn : 0;
+    const delay = toBeExecutedIn + BUFFER_TIMEOUT;
+
+    const timeout = setTimeout(() => {
+      // TODO: execute this.finalizePoll (implementing this in the next PR)
+      this.redisService.delete(key);
+    }, delay);
+
+    const doesExist = this.schedulerRegistry.doesExist("timeout", key);
+    if (doesExist) {
```
0xmad, line 159: And it should have different name because variable should not be a verb

```suggestion
    const doesExist = this.schedulerRegistry.doesExist("timeout", key);

    if (doesExist) {
```

apps/coordinator/ts/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,292 @@
...
+            chain: poll.chain,
+            merged: poll.merged,
+            proofsGenerated: poll.proofsGenerated,
+            endDate,
+          });
+
+          this.logger.log(`poll ${value} to be executed at ${new Date(Date.now() + delay).toISOString()}`);
+        } catch (error) {
+          this.logger.error(`Error restoring timeout for poll key: ${key}`, error);
+        }
```
0xmad, line 288: Why is this error ignored?

apps/coordinator/ts/file/file.service.ts
NicoSerranoP: To remove the ` as ` and explicitly specify the type in the variable

apps/coordinator/ts/scheduler/scheduler.service.ts
NicoSerranoP: If  one poll cannot be restored we are logging the error but not stopping the process of restoring the other ones

PR #2672, opened 2025-07-25
fix: add padding key to signup tree on initialization

silent-cipher:
# Description

Ensure `generateSignupTreeFromKeys` correctly builds the signup tree by inserting the `PAD_KEY_HASH` before inserting hash of actual user public keys into the LeanIMT tree.

### Changes
Inserted the predefined `PAD_KEY_HASH` at the beginning of the LeanIMT tree in `generateSignupTreeFromKeys`.

### Why
There’s no on-chain event corresponding to the `padKey`, so it cannot be fetched via the subgraph.
Two approaches were considered:

1. Manually appending a `padKey` to the list of public keys returned by the subgraph through `getKeys` function.

2. Inserting a known `PAD_KEY_HASH` directly when constructing the tree from public keys.

This PR implements option 2, since the actual public key used to produce `PAD_KEY_HASH` isn’t known. And, only  `PAD_KEY_HASH` is used in sdk.

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/packages/sdk/ts/trees/stateTree.ts
+++ b/packages/sdk/ts/trees/stateTree.ts
@@ -139,6 +139,7 @@ export const generateSignUpTreeWithEndKey = async ({
  */
 export const generateSignUpTreeFromKeys = (publicKeys: PublicKey[]): LeanIMT => {
   const signUpTree = new LeanIMT(hashLeanIMT);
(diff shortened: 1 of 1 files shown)


silent-cipher:
Closing this in favor of #2673 which uses a better approach.

PR #2723, opened 2025-08-13
fix(sdk): start block before fetching events

NicoSerranoP:
# Description

In the Aragon MACI Plugin we were having some issues with Rabby and Rainbow wallet. The main problem is that wallets use their own RPC endpoint when executing write functions. When you request a transaction, the wallet is the JSON‑RPC client and will use its own node(s) to send/broadcast. That’s by design in EIP‑1193.

The issue was that these RPC endpoints had some rate limits lower than Metamask and we were getting errors related to "the amount of blocks to fetch is not available in the free tier" or similar.

I fixed some things in the UI repo (https://github.com/privacy-scaling-explorations/maci-aragon-osx-gov-app/pull/65) but I was able to map down the error to these two functions: `generateMaciStateTreeWithEndKey` and `generateMaciStateTree`. So I decided to move the `fromBlock` logic a little so that when a start block is sent, the default start block is not fetched.

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/packages/sdk/ts/user/utils.ts
+++ b/packages/sdk/ts/user/utils.ts
@@ -1,7 +1,11 @@
-import { MACI__factory as MACIFactory, Poll__factory as PollFactory } from "@maci-protocol/contracts/typechain-types";
-import { IJsonMaciState, MaciState } from "@maci-protocol/core";
+import {
+  MACI__factory as MACIFactory,
+  Poll__factory as PollFactory,
+  type MACI,
+} from "@maci-protocol/contracts/typechain-types";
+import { type IJsonMaciState, MaciState } from "@maci-protocol/core";
 import { poseidon, stringifyBigInts } from "@maci-protocol/crypto";
-import { Keypair, PrivateKey, PublicKey } from "@maci-protocol/domainobjs";
+import { Keypair, type PrivateKey, PublicKey } from "@maci-protocol/domainobjs";
 
 import fs from "fs";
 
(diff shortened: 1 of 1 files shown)


packages/sdk/ts/user/utils.ts
```diff
@@ -271,11 +271,17 @@ export const generateMaciStateTree = async ({
...
+
+  if (startBlock) {
+    fromBlock = Number(startBlock);
+  } else {
+    // compute default start block after confirming that startBlock is not provided
+    const defaultStartBlock = await maciContract
+      .queryFilter(maciContract.filters.SignUp(), 0)
+      .then((events) => events[0]?.blockNumber ?? 0);
+    fromBlock = defaultStartBlock;
+  }
```
0xmad, line 284: Don't use let and don't create temp defaultStartBlock const.
Also, the code is duplicated below so it can be moved into a function.

0xmad (changes requested):
@NicoSerranoP thanks, just one minor change

packages/sdk/ts/user/utils.ts
```diff
@@ -256,6 +260,14 @@ export const getPollJoiningCircuitInputsFromStateFile = async ({
   return circuitInputs;
 };
 
+/**
+ * Get the block number when the first SignUp event was emitted
+ * @param maciContract - The MACI contract to fetch events from
+ * @returns The block number when the first SignUp event was emitted
+ */
+export const getFirstSignUpBlockNumber = async (maciContract: MACI): Promise<number> =>
+  maciContract.queryFilter(maciContract.filters.SignUp(), 0).then((events) => events[0]?.blockNumber ?? 0);
```
0xmad, line 269:
```suggestion
export const getFirstSignUpBlockNumber = async (maciContract: MACI, startBlock?: number): Promise<number> => {
  if (startBlock) {
    return startBlock;
  }

  return maciContract.queryFilter(maciContract.filters.SignUp(), 0).then((events) => events[0]?.blockNumber ?? 0);
```

PR #2819, opened 2026-01-12
feat: add tally result added event

JohnGuilding:
# Description

Adds an event to the `addTallyResult` function to make it easier to retrieve information on completed tallies. For example, pulling the data for a subgraph.

## Confirmation

> [!IMPORTANT]


packages/contracts/contracts/Tally.sol
```diff
@@ -420,5 +423,7 @@ contract Tally is Clone, SnarkCommon, Hasher, DomainObjs, ITally {
     }
 
     previous.value = _tallyResult;
+
+    emit ResultAdded(_voteOptionIndex, _tallyResult);
```
NicoSerranoP: I like the idea but should the event be emitted on `addTallyResults` rather than `addTallyResult` ? My understanding is that the coordinator calls once `addTallyResults` and then `addTallyResult` is called multiple times internally.

I feel it might be more helpful to get an event after all results have been tallied rather than on every execution of partial results tallying 😊

0xmad: Better to use it in `addTallyResult` to use it simplier with subgraph.

PR #2827, opened 2026-01-29
feat(sdk,cli): add isOpen status to poll data and improve CLI output

johnson86tw:
## Summary
- Added `isOpen` field to poll data in SDK, calculated by comparing current block timestamp with poll start/end dates
- Added transaction hash output for `publish` command
- Added poll status display (Open/Closed) and `--quiet` option for `getPoll` command
- Added success message for `verifyTallyCommitment` command


0xmad (approved):
Thanks!

